PR #21, opened 2024-04-17
Upgrade Hipcheck dependencies

### LINKED ISSUE #7: Get Hipcheck dependencies up to date
Since Hipcheck hasn't been actively worked on in more than a year, it's dependencies are generally out of date, and need to be updated. It's likely that the process of updating dependencies will involve some easy cases (no breaking changes), and more difficult cases (no breaking changes). The easy cases should be prioritized first, and then the remaining cases assessed to identify what's necessary to bring Hipcheck's usage up to the latest version.

alilleybrinker:
Resolves #7

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -429,31 +429,11 @@ dependencies = [
 
 [[package]]
 name = "dirs"
-version = "4.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ca3aa72a6f96ea37bbc5aa912f6788242832f75369bdfdadcb0e38423f100059"
-dependencies = [
- "dirs-sys 0.3.7",
-]
-
-[[package]]
-name = "dirs"
 version = "5.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "44c45a9d03d6676652bcb5e724c7e988de1acad23a711b5217ab9cbecbec2225"
 dependencies = [
- "dirs-sys 0.4.1",
-]
-
-[[package]]
-name = "dirs-sys"
-version = "0.3.7"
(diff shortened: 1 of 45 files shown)


alilleybrinker:
Report on the remaining deps to upgrade:

<details>

<summary>Output of <code>cargo outdated -R</code></summary>

```text
hipcheck
================
Name  Project  Compat  Latest  Kind    Platform
----  -------  ------  ------  ----    --------
clap  2.34.0   ---     4.5.4   Normal  ---

hc_git
================
Name  Project  Compat  Latest  Kind    Platform
----  -------  ------  ------  ----    --------
nom   6.1.2    ---     7.1.3   Normal  ---

hc_github
================
Name            Project  Compat  Latest  Kind    Platform
----            -------  ------  ------  ----    --------
graphql_client  0.10.0   ---     0.14.0  Normal  ---
ureq            1.5.5    ---     2.9.6   Normal  ---
```

</details>


So we have `clap`, `nom`, `graphql_client`, and `ureq` left to upgrade.

alilleybrinker:
Down to one final dependency upgrade, `ureq` 1.x to 2.x! This is hilariously the most tedious one, because the got rid of the convenience method to ensure that all calls through a specific `Agent` used HTTP Basic Authentication. Instead, you have to add the specific header on each request, and you have to construct the header yourself. Not _terrible_, but tedious given that they had a thing that worked before and removed it because they didn't like the API.

alilleybrinker:
Alright, I've put it through its paces, fixed some bugs, and now feel confident that the new code works as intended! In the end, the actual changes to Hipcheck itself were pretty minimal, even for major version breaks, which is a nice surprise!

PR #26, opened 2024-04-23
Define devcontainer config

alilleybrinker:
This commit defines a devcontainer configuration for Hipcheck, to enable
contributors to Hipcheck to spin up a devcontainer that has what they
need to work on the project.

Right now that means it installs the latest Rust toolchain, the latest
Long-Term Support (LTS) Node.js toolchain, and then sets up OpenSSL dev
headers and the NPM module-deps package.


alilleybrinker:
My local host Docker setup is currently broken, so I'm going to test this post-merge with GitHub Codespaces.

PR #37, opened 2024-05-01
feat: Move common-use crates into hc_common

mchernicoff:
As part of a general bookkeeping effort, this moves the following common-use crates into the `hc_common` crate:

- `hc_command_util`
- `hc_error`
- `hc_filesystem`
- `hc_pathbuf`
- `hc_test_util`


mchernicoff:
Open to suggestions to restructure hc_common with some nested folders inside src, if that makes things more legible code-wise.

alilleybrinker:
Thanks! I think we can start here and then consider a module restructure after that. LGTM!

alilleybrinker (dismissed):

alilleybrinker:
Ah, looks like there are conflicts preventing a rebase. @mchernicoff think you could resolve them? Happy to do it myself if that's easier.

alilleybrinker:
Hm, looks like it didn't do the Conventional Commits checking in CI, which it should be doing now. So I'll also debug that

mchernicoff:
@alilleybrinker I can do another rebase.

alilleybrinker:
Taking a look at the updates now!

alilleybrinker (approved):

alilleybrinker:
Okay @mchernicoff looks like it just needs a final review from you since I pushed some minor fixes to the PR branch. Then we're good to merge!

PR #41, opened 2024-05-01
feat: Add 'cargo-dist' for prebuilt binaries.

alilleybrinker:
This commit is just the result of getting 'cargo dist init' running, and
sets up the basics for distributing Hipcheck to end users with pre-built
binaries. Note that currently the installation after the binary is set
up will still be missing additional files Hipcheck needs, as well as its
broader dependencies, so this isn't a sufficient installation setup on
its own.


alilleybrinker:
Rebased on the latest from `main`.

PR #42, opened 2024-05-01
fix: Remove 'atty' dep for GHSA-g98v-hv3f-hcfr

alilleybrinker:
This commit removes `atty` as a dependency. The Rust standard library
provides the same functionality built in now, and `atty` has a known
low-severity vulnerability and is unmaintained.


alilleybrinker:
See: https://github.com/mitre/hipcheck/security/dependabot/1

PR #45, opened 2024-05-03
feat: Creates a single analysis crate that handles most of the Hipcheck analysis pipeline

mchernicoff:
Moves the code from the below crates into a single `hc_analysis` crate:

- `hc_analysis`
- `hc_metric`
- `hc_score`
- `hc_session`
- `hc_report_builder`

The idea is that most of the code thar starts a Hipcheck session, calculates metrics, analyzes metrics, scores analysis, and constructs a final report will all be inside a single crate.


mchernicoff:
Marking as a draft for now because more testing may be needed to make sure these changes do not break Hipcheck.

alilleybrinker:
I know it's still a draft, but I'm reviewing now!

alilleybrinker:
LGTM! Tested locally and everything ran without issue.

PR #48, opened 2024-05-07
Added "Release Engineering" RFD

alilleybrinker:
This commit adds a "Release Engineering" RFD which proposes both how the workflow should run for publication of new Hipcheck versions, as well as changes to Hipcheck itself to support the new workflow.

__[Rendered](https://github.com/mitre/hipcheck/blob/4cdc6d3358fe4640a80a96a4f7b59ce108580bda/docs/rfds/0001-release-engineering.md)__

### DIFF
--- a/docs/rfds/0001-release-engineering.md
+++ b/docs/rfds/0001-release-engineering.md
@@ -1,0 +1,60 @@
+# Hipcheck Release Engineering
+
+One of the key questions for any piece of software is how new versions will be
+created and distributed out to end-users.
+
+Currently, Hipcheck basically handles distribution through the Git repository
+itself, and an `install.sh` script. The script clones the Hipcheck repository,
+builds Hipcheck from source with a Rust toolchain, and then places the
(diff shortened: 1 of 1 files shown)


mchernicoff (approved):
Sounds reasonable to me. My only concern is that we have enough warning messages and safeguards in place to cover what happens if the setup command is not run correctly. Or, if it is run, but something has changed about the installation between when the binary has been installed and setup is run (you never know what could happen to the files in the meantime).

alilleybrinker:
> Sounds reasonable to me. My only concern is that we have enough warning messages and safeguards in place to cover what happens if the setup command is not run correctly. Or, if it is run, but something has changed about the installation between when the binary has been installed and setup is run (you never know what could happen to the files in the meantime).

This is a reasonable concern. My intent here would be that we basically keep in place the error handling we have today for things like configuration files not being found, or missing required contents, and ideally add a line to the output recommending that they run `hc setup check` for more detailed information on what's wrong with their setup.

PR #68, opened 2024-05-13
chore: Do not run tests if changes outside code folders

mchernicoff:
Filter GitHub workflow to not run tests if changes to a push or pull-request are outside of code folders `config`, `hipcheck`, `scripts`, or `xtask`.


alilleybrinker (approved):
LGTM!

PR #70, opened 2024-05-14
Add "Hipcheck's Values" RFD

### LINKED ISSUE #17: Write an explainer for Hipcheck's values as a project
Hipcheck has some values which are central to _why_ it exists and help inform its features, and which are also important for differentiating it from other tools in the software supply chain space. Once we've gotten Hipcheck's documentation site up and running (see #5), we should document the values of the Hipcheck project. These include:

* __Security-Focused__: Hipcheck's analyses are solely intended to be focused on security-relevant practices and identifiable risks associated with a project, and _not_ analyses which identify non-security-relevant best practices.
* __Information, Not Scores__: We do not ever want Hipcheck's scores to become a metric which users of open source software use to bully maintainers into compliance. Open source maintainers are doing often thankless work with limited resources in both time and funding, and we should always focus on how Hipcheck can empower them, not make their job harder by creating a metric to drive up.
* __Configurable__: Security _is_ context-sensitive. Different risks may be weighted differently by different users, or even be show-stoppers for some. The nature of what is concerning may change depending on the context. We are making a security tool, and while we _can_ and _should_ have recommended defaults, we should also empower users to tailor the tool to their needs, so it can be adaptable to their systems and their security requirements, rather than them adapting to our limitations.

alilleybrinker:
This commit adds a new RFD describing Hipcheck's product values and project values.

[__Rendered__](https://github.com/mitre/hipcheck/blob/25ff3b28ffa6ffb40ab5ebe935a8360f9a116c49/docs/rfds/0002-hipcheck's-values.md)

Closes #17

### DIFF
--- a/docs/rfds/0002-hipcheck's-values.md
+++ b/docs/rfds/0002-hipcheck's-values.md
@@ -1,0 +1,77 @@
+
+# Hipcheck's Values
+
+Often, the _values_ associated with an open source software project are left
+implicit rather than explicit. They form based on the beliefs and expectations
+of the original contributors, and are reflected in the practices and
+priorities of the project, but are not written down. The problem with this is
+that values are central to successful collaboration, which requires shared
+goals. Any open source software project relies on having a variety of
+interested contributors who are rowing in the same direction. When values
+are implicit, novel situations may arise which reveal deep divergence in
+values between those contributors, leading to crises of governance and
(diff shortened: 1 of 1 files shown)


mchernicoff (approved):

mchernicoff:
LGTM

PR #71, opened 2024-05-14
Added RFD #3, "Plugin Architecture Vision"

### LINKED ISSUE #10: Write the "Plugin Architecture End State" RFD
One of the major goals for Hipcheck this year is to rearchitect it to support third-party plugins for both data collection and data analysis. Before we get into designing the specifics of this new system, it'll be important to lay out exactly what "this new system" means.

We'll need an RFD which specifies what we want third-parties to be able to do abstractly (without specifically describing the API), and what the user experience of the new architecture should be.

It also ought to reflect the architecture of how we expect data to flow within the Hipcheck system, and how Hipcheck will be planned to orchestrate the execution of components.

This will likely manifest as a sort of "lifecycle" of a Hipcheck execution, which begins with the user describing the workload they want to run (some set of analyzes, relying on some set of measurements and some set of data), validating that workload against the registered plugins, planning the execution of the workload to maximize performance, and then executing the workload. The Hipcheck engine would call out to plugins as determined in the work plan to gather data, store it, and then analyze it, eventually producing a final analysis based on the user's configuration.

Later RFD's will further clarify details about the API, and the actual implementation to produce this end state.

alilleybrinker:
This commit adds an RFD defining the vision for Hipcheck's planned addition of support for third party plugins.

[__Rendered__](https://github.com/mitre/hipcheck/blob/4c90164714636e8aa338d8290a8073f78cde2c7d/docs/rfds/0003-plugin-architecture-vision.md)

Closes #10


mchernicoff (approved):

mchernicoff:
LGTM

PR #73, opened 2024-05-17
resolve #65: Update README.md to reflect correct installation instructions

### LINKED ISSUE #65: Update `README.md` to reflect correct installation instructions.
The `README.md` currently says you can install Hipcheck with `cargo xtask install` but that command is actually no longer present and will not work. The proper command is now `cargo install --path hipcheck`.

j-lanson:
Changes are basically all prosaic, not much technical to discuss

Design decisions
- Added Node as a required dependency of Hipcheck. After the plugin architecture is implemented, Node won't be a blanket requirement, but I opted to have the README reflect the current state
- Describe `install.sh` as supported installation method, different than manual "clone and build" approach. `install.sh` tells users to set `HC_DATA` and `HC_CONFIG` with provided values, whereas manual approach requires the user to decide on and define their own `data/config` values that may be passed as CLI args or set as env vars
- Chose to copy the description of `--home, --config, --data` wholesale from the book. Maybe there's a better way to have them both point to the same text without replication?


README.md
```diff
@@ -92,16 +113,43 @@ $ docker run --env "HC_GITHUB_TOKEN=<GITHUB_TOKEN>" hipheck:3.1.0 [<HIPCHECK_ARG
 
 ### Direct Usage
 
-You can run Hipcheck with the `hc` binary.
+To run Hipcheck, make sure you export `HC_GITHUB_TOKEN` with a valid token for
```
mchernicoff, line 116: Strictly speaking you don't _have_ to set `HC_GITHUB_TOKEN` if you are willing to use the appropriate flag every time you run Hipcheck. Maybe change the language to say that we recommend doing this to avoid having to specify the token when running Hipcheck.

j-lanson: What portion of the code implements that cmdline arg? When I look at `analysis/session/mod.rs::resolve_token()` it looks like it only checks `HC_GITHUB_TOKEN`

README.md
```diff
@@ -66,18 +66,39 @@ $ # Run the following from the root of the Hipcheck repository.
...
 
 First, install the Rust compiler. We recommend following the official
 [installation instructions][install_rust]. Make sure to add
 `${CARGO_HOME}/bin` to your `PATH`.
 
-Next, get the Hipcheck repository and run `cargo xtask install`.
+You will also need Node installed. We recommend following the official
+[installation_instructions][install_node].
+
+You may either install using an [automated script](#script-based-installation)
```
j-lanson, line 78: Should amend to "install Hipcheck" here since just was talking about Node last sentence

docs/book/src/using/cli.md
```diff
@@ -29,7 +29,7 @@ Analyzes a repository or single pull request with a set of analysis tools. Hipch
 
 A github token must be set in the HC_GITHUB_TOKEN system environment variable for `hc check` to run. Hipcheck only needs permissions for accessing public repository data, so those are the only permissions to assign to your generated token.
 
-There are currently two working secondary subcommands under `check`. Running `hc check` without a subtask runs `hc help check`. The available subcommands are:
+There are currently six working secondary subcommands under `check`. Running `hc check` without a subtask runs `hc help check`. The available subcommands are:
```
mchernicoff, line 32: I don't think the SPDX check is working yet either. Which puts us at 5 working subcommands.

README.md
mchernicoff: Oh never mind.  We had that as a feature at one point but we didn't implement it.

docs/book/src/using/cli.md
j-lanson: Should `SPDX` be removed, or distinguished as a yet-unfinished command but left in the book?

README.md
mchernicoff: Agree

docs/book/src/using/cli.md
mchernicoff: I would remove it from the book for now. We even took the preliminary code to handle it out of the repo because it wasn't finished yet. When I update the CLI, I will indicate that its a future feature there.

mchernicoff (approved):

mchernicoff:
LGTM now. Can always make more changes to the README in the future, but at least we don't have wrong information in there now.

PR #79, opened 2024-05-20
chore (cli): Changes hc CLI to use compile-time "derive" syntax

mchernicoff:
This changes the CLI for the `hc` command to use `clap`'s "derive" syntax over the previous "build" syntax. This makes the CLI run at compile-time, which will save time when Hipcheck is run on different data sources without recompilation in between (as we expect) with typical use.

As a bonus feature, the CLI no longer panics if no command argument is provided. Instead it displays the appropriate help text, as expected.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,17 +4,17 @@ on:
   push:
     branches: [main]
     paths:
-      - 'config/**'
-      - 'hipcheck/**'
-      - 'scripts/**'
-      - 'xtask/**'
+      - "config/**"
+      - "hipcheck/**"
+      - "scripts/**"
+      - "xtask/**"
   pull_request:
     branches: [main]
     paths:
-      - 'config/**'
-      - 'hipcheck/**'
-      - 'scripts/**'
-      - 'xtask/**'
+      - "config/**"
+      - "hipcheck/**"
+      - "scripts/**"
+      - "xtask/**"
 
 permissions:
   contents: read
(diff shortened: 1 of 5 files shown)


mchernicoff:
Turns out I had already fixed the panic w/o arguments issue, but I was testing with an older build.

mchernicoff:
Annoyingly, the test passes when I run it locally.

Test is likely failing because we are trying to have our cake and eat it too with respect to our help command.

The current version of the CLI implements a custom help text that works on either the `-h`/`--help` flag or the `help` subcommand (as well as calling `hc` without any arguments). Despite renaming the flag and turning off default help, we are still failing this test.

Options are:

1. Figure out a way to pass the test
2. Delete the test, because it is more harm than help at this point
3. Get rid of the `--help` flag, because we already have two other ways to get the help text.

alilleybrinker:
@mchernicoff taking a look now. I rebased on the latest `main`, including fixing a conflict, and also added a new CI step to see the full `cargo tree` output to help debug this and future CI issues. This does mean the remote branch has been force pushed and can't just be `pull`-ed. If you have un-pushed local work let me know, happy to help with Git issues.

alilleybrinker:
The specific assert that's causing the test to fail is here: https://github.com/clap-rs/clap/blob/master/clap_builder/src/builder/debug_asserts.rs#L334-L339

Basically, it's running through all subcommands from the `.get_subcommands()` method and panicking if it finds any duplicates. In this case it's saying that the `help` subcommand is duplicated.

Unfortunately I haven't been able to duplicate this in CI myself. Will keep working on it.

PR #80, opened 2024-05-20
Chore: Remove OpenSSL as a Hipcheck dependency

mchernicoff:
Removes OpenSSL as a necessary Hipcheck dependency. Accomplishes this by switching the [ureq](https://crates.io/crates/ureq) https client's TLS implementation from [native-tls](https://crates.io/crates/native-tls) (which uses OpenSSL on Linux platforms) to [rustls](https://crates.io/crates/rustls).

Updates the documentation and Docker Containerfile updated to reflect Hipcheck not needing a separate OpenSSL installation.


j-lanson (approved):
Builds and runs fine on my local machine too. Looks good.

PR #81, opened 2024-05-24
feat: Adds hc ready command

mchernicoff:
Adds the `hc ready` command. This checks that:

- git and NPM are installed and sufficiently up to date, and if installed prints their versions
- The home directory, config file, and data directory can all be found, and if set prints their respective locations
- The HC_GITHUB_TOKEN environment variable is set (but it does not check that the token is valid)

If git and NPM are up to date and home, config, and data are all found, the command reports that Hipcheck is ready to run. Otherwise it reports that Hipcheck is not ready.

### DIFF
--- a/hipcheck/src/cli.rs
+++ b/hipcheck/src/cli.rs
@@ -71,6 +71,9 @@ pub enum Commands {
 	/// Analyzes a repository or pull/merge request
 	#[command(disable_help_subcommand = true)]
 	Check(CheckArgs),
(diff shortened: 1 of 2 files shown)


mchernicoff:
Draft PR because the PR with the CLI changes should be merged first, as this branch was forked from the branch for that PR. Once that PR is ready, rebase this branch and then merge.

mchernicoff:
Tests are failing for the same reason as PR 79. See that PR for possible resolutions.

PR #85, opened 2024-05-28
chore(http): Change `ureq` Agent to use native system certs

mchernicoff:
Use the `rustls-native-certs` crate to have `rustls` use the user's native system certs instead of the Mozilla root certs when constructing a `ureq` authentication agent. This is the only place Hipcheck needs these certs, so it is now configured to use system certs with TLS.

### DIFF
--- a/hipcheck/Cargo.toml
+++ b/hipcheck/Cargo.toml
@@ -36,6 +36,7 @@ paste = "1.0.7"
 pathbuf = "1.0.0"
 petgraph = { version = "0.6.0", features = ["serde-1"] }
 regex = "1.5.5"
+# Exactly matching the version of rustls used by ureq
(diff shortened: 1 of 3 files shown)


hipcheck/Cargo.toml
```diff
@@ -38,6 +38,8 @@ paste = "1.0.7"
 pathbuf = "1.0.0"
 petgraph = { version = "0.6.0", features = ["serde-1"] }
 regex = "1.5.5"
+rustls = "0.22.4"
```
alilleybrinker, line 41: Added comment noting we're exactly matching the versions used by `ureq`.

alilleybrinker (changes requested):

alilleybrinker (approved):

Issue #86, opened 2024-05-29
Change "validate" to "check in `xtask ci`


mchernicoff:
`xtask ci` is still calling `xtask validate`, which has been renamed to `xtasck check`. Update this to the correct name to keep `xtask ci` from failing.

PR #89, opened 2024-05-29
fix: Change xtask validate to xtask check when xtask ci is called.

mchernicoff:
Resolves Issue #86  Fixes the `xtask ci` task to correctly call `xtask check` instead of the deprecated `xtask validate`.


j-lanson (approved):
Looks good! Straightforward change.

PR #93, opened 2024-05-30
Refactor CLI

alilleybrinker:


alilleybrinker:
This should hopefully fix the broken CLI test. It's also all been rebased on `main` to incorporate the `ready` command.

alilleybrinker:
Realized that the existing tests actually no longer made sense with the CLI refactor. I've moved the key tests into the `cli` module and rewritten them to properly test the new API. The key thing they're testing is just the path args, as those are the only ones with the "cascade" logic that we care about.

Issue #106, opened 2024-06-01
Make `hc schema` equivalent to `hc schema -h`


alilleybrinker:
See #105, do the same for `hc schema`.

PR #107, opened 2024-06-02
feat(cli): hc check and schema will print help msg when no subcommand

### LINKED ISSUE #105: Make `hc check` equivalent to `hc check -h`
Right now if you run `hc check`, you get an error message about an unknown check type. It would be a better user experience to treat this as equivalent to running the `-h`/`--help` command.

j-lanson:
This MR resolves #105  and #106 issues pertaining to the replacement of unhelpful error msgs that Hipcheck emits when either `check` or `schema` subcommands are invoked without any arguments. 

This is addressed simply by updating the type of the `command` field in the `CheckArgs` and `SchemaArgs` structs from an `Option` to just the underlying type. This conveys to `clap` that **some** subcommand must be parsed from the remaining arguments for things to move forward - if this does not happen, it will print the `-h` help msg. 

The `match` statements associated with these `command` fields in `main.rs` are updated in accordance with this change.

### DIFF
--- a/hipcheck/src/cli.rs
+++ b/hipcheck/src/cli.rs
@@ -390,7 +390,7 @@ pub enum Commands {
 #[derive(Debug, Clone, clap::Args)]
 pub struct CheckArgs {
 	#[clap(subcommand)]
-	pub command: Option<CheckCommand>,
+	pub command: CheckCommand,
 }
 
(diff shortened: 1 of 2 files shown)


alilleybrinker:
LGTM!

alilleybrinker (approved):

PR #112, opened 2024-06-03
Add `hc setup` subcommand to finish Hipcheck installation

### LINKED ISSUE #103: Implement `hc setup` subcommand to complete installation of Hipcheck
Per RFD 002, the goal for the new setup process is that you install just the Hipcheck binary (ideally, pre-built) and then run `hc setup` to complete the setup process. This command should basically do what the current `install.sh` script does today: copy config files and script files to the proper location, and then direct the user to set up the relevant environment variables.

j-lanson:
Resolves #103.

- Uses `CliConfig` to get definitions for `HC_DATA` and `HC_CONFIG` from user in same precedence order as main `hc` commands. Difference is we may create these dirs if they do not exist
- `HC_BIN` - if `hc` is not already in path, this is the target location for `hc` to be copied to. First tries `HC_BIN` env var, then uses `dirs::executable_dir()`. 

Steps:
1. Resolve and create (if necessary) `HC_DATA` and `HC_CONFIG` dirs
2. Check if `hc` on path. If not, determine `HC_BIN` val and copy `hc` exe to that location. If in new location `hc` still not on sys path, warn user.
3. Copy config from `./config` to `HC_CONFIG` and data from `./scripts` to `HC_DATA`. **this expects `hc setup` to be run from repo base as we don't have `cargo-dist` built packages yet**. This copies the file contents, not the directories wholesale to minimize chance of user accidentally trying to overwrite important system dirs with bad `HC_CONFIG` value, etc.
4. Emit suggestions for `HC_CONFIG`, `HC_DATA` values as `install.sh` did

Things `install.sh` does that this does not:
- Check for `cargo` and `curl`
- Download and build Hipcheck

Design Discussions:
1. Handling of src dirs for `HC_CONFIG` and `HC_DATA` in Step 3 ok for now?
2. Anything added that should be moved out of `main.rs`?
3. Maybe make `cmd_setup` more sheepish about overwriting existing files (e.g. someone calls `hc setup` which would overwrite a customized `Hipcheck.toml` file in target dir) and add a `--force` flag?

### DIFF
--- a/hipcheck/src/cli.rs
+++ b/hipcheck/src/cli.rs
@@ -361,6 +361,7 @@ fn hc_env_var_value_enum<E: ValueEnum>(name: &'static str) -> Option<E> {
 pub enum FullCommands {
 	Check(CheckArgs),
 	Schema(SchemaArgs),
+	Setup,
 	Ready,
 	PrintConfig,
 	PrintData,
@@ -372,6 +373,7 @@ impl From<&Commands> for FullCommands {
 		match command {
 			Commands::Check(args) => FullCommands::Check(args.clone()),
 			Commands::Schema(args) => FullCommands::Schema(args.clone()),
+			Commands::Setup => FullCommands::Setup,
 			Commands::Ready => FullCommands::Ready,
 		}
 	}
@@ -383,6 +385,8 @@ pub enum Commands {
 	Check(CheckArgs),
 	/// Print the JSON schema for output of a specific `check` command.
 	Schema(SchemaArgs),
+	/// Initialize Hipcheck config file and script file locations.
+	Setup,
 	/// Check if Hipcheck is ready to run.
 	Ready,
 }
@@ -498,7 +502,7 @@ pub struct SchemaArgs {
 
 #[derive(Debug, Clone, clap::Subcommand)]
 pub enum SchemaCommand {
-	/// Print the JSONN schema for running Hipcheck against a Maven package
+	/// Print the JSON schema for running Hipcheck against a Maven package
 	Maven,
 	/// Print the JSON schema for running Hipcheck against a NPM package
 	Npm,
--- a/hipcheck/src/main.rs
+++ b/hipcheck/src/main.rs
@@ -148,31 +148,33 @@ fn cmd_schema(args: &SchemaArgs) {
 /// Copy individual files in dir instead of entire dir, to avoid users accidentally
 /// overwriting important dirs such as /usr/bin/
 fn copy_dir_contents<P: AsRef<Path>, Q: AsRef<Path>>(from: P, to: Q) -> Result<()> {
-	let src: PathBuf = from.as_ref().to_path_buf();
-	if !src.is_dir() {
-		return Err(hc_error!("source path must be a directory"));
-	}
-	let dst: PathBuf = to.as_ref().to_path_buf();
-	if !dst.is_dir() {
-		return Err(hc_error!("target path must be a directory"));
-	}
-
-	for entry in walkdir::WalkDir::new(&src) {
-		let src_f_path = entry?.path().to_path_buf();
-		if src_f_path == src {
-			continue;
+	fn inner(from: &Path, to: &Path) -> Result<()> {
+		let src = from.to_path_buf();
+		if !src.is_dir() {
+			return Err(hc_error!("source path must be a directory"));
 		}
-		let mut dst_f_path = dst.clone();
-		dst_f_path.push(
-			src_f_path
-				.file_name()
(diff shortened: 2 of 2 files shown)


hipcheck/src/cli.rs
```diff
@@ -498,7 +502,7 @@ pub struct SchemaArgs {
 
 #[derive(Debug, Clone, clap::Subcommand)]
 pub enum SchemaCommand {
-	/// Print the JSONN schema for running Hipcheck against a Maven package
+	/// Print the JSON schema for running Hipcheck against a Maven package
```
alilleybrinker, line 505: Haha good catch!

hipcheck/src/main.rs
```diff
@@ -144,6 +145,130 @@ fn cmd_schema(args: &SchemaArgs) {
...
+	if !tgt_data_path.exists() && create_dir_all(tgt_data_path).is_err() {
+		print_error(&hc_error!("failed to create missing target data dir"));
+	}
+	let Ok(abs_data_path) = tgt_data_path.canonicalize() else {
+		print_error(&hc_error!("failed to canonicalize HC_DATA path"));
+		return ExitCode::FAILURE;
+	};
+
+	// Copy current `hc` binary to OS-specific bin path if not there already
+	if which::which("hc").is_err() {
```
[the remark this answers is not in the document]
alilleybrinker: I think this may be holder behavior from `install.sh`. The distinction in this case is that `hc setup` would be run _after_ the `hc` binary itself is in place on the local system.

hipcheck/src/main.rs
```diff
@@ -144,6 +145,130 @@ fn cmd_schema(args: &SchemaArgs) {
...
+		};
+		if which::which("hc").is_err() {
+			println!(
+				"warning: `hc` copied to '{}', which is not in your system path. Consider adding it",
+				abs_bin_path.parent().unwrap().display(),
+			);
+		}
+	}
+
+	// Copy local config/data dirs to target locations
```
alilleybrinker, line 239: I know it's not testable until `cargo-dist` has published a new version, but the right flow here really will be to get the additional files from the `cargo-dist` bundle for the release, download them, and place them into the correct location.

alilleybrinker (changes requested):

hipcheck/src/main.rs
```diff
@@ -144,6 +145,130 @@ fn cmd_schema(args: &SchemaArgs) {
 	}
 }
 
+/// Copy individual files in dir instead of entire dir, to avoid users accidentally
+/// overwriting important dirs such as /usr/bin/
+fn copy_dir_contents<P: AsRef<Path>, Q: AsRef<Path>>(from: P, to: Q) -> Result<()> {
```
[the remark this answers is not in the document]
j-lanson: Wow this is a neat trick for getting around the size drawbacks of generics! I didn't know you could nest functions in Rust.

hipcheck/src/main.rs
j-lanson: Would you rather leave this functionality out for now, or keep/alter it and make a ticket to update it once we have a compatible `cargo-dist` bundle?

alilleybrinker: I think this MR will have to wait until I can publish a new release which includes those files with `cargo-dist`.

alilleybrinker:
When we're able to finish and merge this we should update the `cargo-dist` config setting `custom-success-msg` to tell the user to run `hc setup`.

j-lanson:
Notes from 6/20 discussion with Andrew:

- dist file something like `hc-setup-materials.tar.xz`, contains binary + `config`, `setup` folders
- `hc setup`
  - Optionally take a flag to specify where to find the extra setup files (can also be set by environment variable)
  - Otherwise, checks system download directory for compressed bundle that matches current hc version
  - Otherwise, hits network, checks GitHub release for bundle (probably even have a flag to never do that) (--offline)
- Paths to installation:
  - (recommended) cargo-dist-generated installer
  - cargo binstall
  - cargo install
  - (closed environment) downloading proper *.tar.xz bundle for their architecture
- Searching for `config`, `scripts` dirs:
  - the current directory.
  - the platform downloads directory, if known - either as a `.tar.xz` file or a top-level `cargo-dist-<architecture_name>-<version_number>`
- Other todos:
  - remove install.sh
  - merge update readme branch
  - update install instructions in readme to reflect new flow

alilleybrinker:
@j-lanson with the 3.3.2 release, we _do_ now include the `config/` and `scripts/` directories in the bundled material, so I think we're good to with the stuff we discussed yesterday.

alilleybrinker:
Closing in favor of #152 which is a rebased version of this that was merged today.

PR #113, opened 2024-06-03
Publish tagged HC releases on Dockerhub in MITRE namespace

### LINKED ISSUE #22: Push images to Docker Hub
Can coordinate with @Amndeep7. We have a `Containerfile` but don't currently push any prebuilt images to Dockerhub. Ideally we'd do this with each new version in an automated fashion. The SAF folks control the MITRE namespace on Docker Hub so we'll coordinate with them.

- [ ] Talk with Amndeep about how to authenticate for publication to the MITRE Docker Hub namespace.
- [ ] Investigate GitHub Actions for publishing new images when new release version tags are pushed to Github.

j-lanson:
Resolves #22.

Repo has been updated with secrets for `DOCKER_USERNAME` and `DOCKER_TOKEN`.

Being a github action, this is difficult to test without publishing a new release or manually triggering by modifying the `on` field of the action's `.yml` file. 

In theory we should be able to manually trigger from GitHub once this is merged into `main`

### DIFF
--- a/.github/workflows/push-to-docker.yml
+++ b/.github/workflows/push-to-docker.yml
@@ -1,0 +1,43 @@
+name: Push Hipcheck image to Docker Hub on every tagged release
+
+on:
+  push:
+    tags:
+      - '**[0-9]+.[0-9]+.[0-9]+*'
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version'
+        required: true
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Run string replace
+        uses: frabert/replace-string-action@v2
+        id: format-tag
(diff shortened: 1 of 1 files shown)


alilleybrinker (changes requested):
Thanks @j-lanson! Some quick questions:

- Why do we need qemu and buildx?
- For the pushed tag, instead of `hipcheck:release-latest`, can it just be `hipcheck:latest`? 
- As I understand it, the `workflow-dispatch` setting for Actions triggers means we can manually trigger publication of a new version. I'm inclined to keep it to _only_ publish on new tags.
- The regular expression for triggering new image publication should be scoped to just tags with the pattern `hipcheck-\d+\.\d+\.\d+`, as I _think_ the current one would match publication of `hipcheck-macros` or `xtask` as well (not that I expect to tag them, but it's good to be conservative with this)/

j-lanson:
If we're not planning on doing multi-platform Docker builds, then we can get rid of buildx (and qemu). Just something used in the example for `build-and-push` action that SAF likely copied, and I copied from them.

alilleybrinker (approved):

PR #114, opened 2024-06-03
fix mishandling of `HC_CONFIG` with new CLI

### LINKED ISSUE #108: Tries to open "HC_CONFIG" dir as file without appending "Hipcheck.toml"
On `main` branch (commit 287346535), I have installed Hipcheck with `cargo install --path hipcheck`, and have `HC_CONFIG` env var set to `/home/jlanson/.config/hipcheck`. When I try to run a `hc check repo ...` command, I get an error during analysis:

```
In Progress loading configuration and data files  
Error Failed to load configuration. Please make sure the config files are in the config directory.
                  can't parse config file
                  failed to read as UTF-8 string '/home/jlanson/.config/hipcheck'
                  Is a directory (os error 21)
```

This implies that the path passed to `Config::load_from()` in `analysis/session/mod.rs::load_config_and_data()` is the config directory, and fails to append `Hipcheck.toml` to the end of the path before trying to "load" it.

j-lanson:
Resolves #108 .

Function `resolve_config()` is no longer called in `load_config_and_data()`, and used to return a path with `Hipcheck.toml` at the end, necessitating a call to `.parent()` to get the conf dir to pass to `Config::load_from()`. With the new CLI, the path doesn't have `Hipcheck.toml` on the end, but the `.parent()` call was left accidentally.

This fix removes the unneeded call, and adds a check to `Config::load_from()` that the passed path is not a regular file to provide users with a clearer error message if they accidentally set `HC_CONFIG` to point to `Hipcheck.toml`.

### DIFF
--- a/hipcheck/src/config.rs
+++ b/hipcheck/src/config.rs
@@ -28,8 +28,7 @@ impl Config {
 				"Hipcheck config path must be a directory, not a file."
 			));
 		}
-		let mut config_file = PathBuf::from(config_path);
-		config_file.push("Hipcheck.toml");
+		let config_file = pathbuf![config_path, "Hipcheck.toml"];
 		file::exists(&config_file)?;
 		let config = file::read_toml(config_file).context("can't parse config file")?;
 
--- a/hipcheck/src/config.rs
+++ b/hipcheck/src/config.rs
@@ -4,6 +4,7 @@
 
 use crate::context::Context;
 use crate::error::Result;
+use crate::hc_error;
 use crate::util::fs as file;
 use crate::BINARY_CONFIG_FILE;
 use crate::F64;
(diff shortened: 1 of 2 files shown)


hipcheck/src/config.rs
```diff
@@ -22,8 +23,15 @@ use std::rc::Rc;
...
 	pub fn load_from(config_path: &Path) -> Result<Config> {
-		file::exists(config_path)?;
-		let config = file::read_toml(config_path).context("can't parse config file")?;
+		if config_path.is_file() {
+			return Err(hc_error!(
+				"Hipcheck config path must be a directory, not a file."
+			));
+		}
+		let mut config_file = PathBuf::from(config_path);
+		config_file.push("Hipcheck.toml");
```
alilleybrinker, line 32: You could use the `pathbuf::pathbuf` macro here instead:

```rust
let config_file = pathbuf![config_path, "Hipcheck.toml"];
```

alilleybrinker (changes requested):
One small change, otherwise LGTM!

Issue #98, opened 2024-06-01
Split out HTTP requests into a helper module


alilleybrinker:
Right now we have some logic for making HTTP requests, used for accessing the GitHub REST API. It would be good to pull that out into something reusable in other places, also to make sure that any settings which impact HTTP requests in the future can be set in one place and then be consistent across the application.

PR #118, opened 2024-06-04
chore: Creates general `http` module for making requests

mchernicoff:
Resolves issue #98.

Moves code for creating authenticated agents for http requests into its own `http` module, in case we want to use that code for something other than making requests to the GitHub API.

This module includes a new, separate function for constructing `ureq` `Agents` with TLS using native system certificates. Even parts of our code that do not need authenticated `Agents` now use this function to make sure we are using the same TLS settings everywhere in Hipcheck.


mchernicoff:
This is one possible way to handle Issue #98. Happy to take suggestions for other ways to do this.

j-lanson (approved):
Looks good!

alilleybrinker:
@mchernicoff, LGTM! Anything keeping this marked as a draft?

mchernicoff:
@alilleybrinker Just me making sure this was what we were looking to do. Fine to merge as is.

PR #120, opened 2024-06-06
Refactor `hc check` CLI

### LINKED ISSUE #94: Refactor CLI for `check` args
Right now, we have a bunch of subcommands for `hc check` to specify the specific type of object being checked. Instead, it would be nice to have something like the following:

```sh
$ hc check [-t/--target = "<TYPE>"] <TARGET>
```

So it might look like the following:

```sh
$ hc check -t npm express                         # specified as an NPM package
$ hc check pkg:npm/express                        # inferred as an NPM package
$ hc check my-package.spdx                        # inferred as a text-format SPDX file
$ hc check https://github.com/expressjs/express   # explicit GitHub repo
$ hc check pkg:github/expressjs/express           # Package URL GitHub repo
```

Basically, it can be an explicit URI, a Package URL, or a package name. Package names require a type, so we know what package host it's on.

---

## Compatibility

The prior format will need to be preserved to avoid breaking compatibility, but we should hide it in the help text and encourage people to transition to the new format.

j-lanson:
Resolves #94 .

This MR introduces a new CLI feature to allow omission of the subcommand in `hc check <subcommand> <target>`. The target value can instead be analyzed after `clap` parses arguments to derive the proper subcommand (repo, maven, npm, etc.). If the target cannot be automatically derived, the `-t / --target` flag is provided to make it explicit. As part of this merge, the old `hc check <subcommand>` is soft-deprecated, and hidden in the CLI `--help` text.

`hipcheck/src/cli.rs`
- the `CheckArgs` struct is augmented with new direct arguments. The `<TARGET>` arg is required, except when a subcommand has been parsed first. We use a trailing var arg to collect any additional args the user might provide when the non-subcommand usage style is invoked.
- The `command` field of `CheckArgs` is made private, and an accessor method `command()` is provided for use in its place. If the `command` field has a value, it will be used, otherwise we will try to derive from the `target_type` and `target` fields.
- the various subcommands under `CheckCommand` are hidden. I could not find a better way to do this with a single, higher-level `hide = true`. 
- test cases are introduced to ensure 1. `--help` is still printed when `hc check` is invoked without any arguments, 2) the deprecated mode still works 3) an Err obj is returned when a target can't be derived, 4) target derivation works and 5) `-t` command will override auto-derivation.

`hipcheck/src/main.rs` 
- Now uses `.command()` accessor method, and appropriately prints error / returns failure code when an `Ok()` response is not received.

`hipcheck/src/target.rs`
- Houses the logic for deriving a `TargetType` enum value from the target string provided by the user.


j-lanson:
@alilleybrinker If you could, please pull and try this branch out locally, I imagine you'll have feedback on the usage printout.

alilleybrinker (approved):

PR #127, opened 2024-06-17
refactor: added new types to form the basis of scoring refactor

j-lanson:
**Note**: Not sure why the Ubuntu `clippy` is complaining about unused fields now that it used to not have a problem with. Perhaps a change was made over the past few days that removed `Debug` from the field usage calculation?

An attempt at defining new types for overhauling the `AnalysisResult` system and Hipcheck scoring in general. `HC` prefix was added to structs/variants to disambiguate with existing structs and can/should be renamed once the old system has been removed in a subsequent PR.

The goal of this PR is to produce an `AnalysisReport` replacement with the following changes:

- Analysis implementations no longer check `db.<X>_active()` to decide themselves whether or not to run. The old `AnalysisOutcome` enum contains the `Skip` variant which is absent from the new one. Instead, we wrap the `HCAnalysisResult` in an option to represent the skipped/executed status of an analysis.
- Analyses no longer determine pass/fail themselves, they will simply return a "value". Thus, `Pass`, `Fail` language is also removed from this replacement infrastructure, and analyses won't know or return their own thresholds. 
- An infrastructure that does not require a separate enum variant of `AnalysisReport` for each analysis type as the existing system does. In this new system, analyses will return a basic/composite value type, and the config/analysis will select from a set of "predicates" that specify how that result is to be interpreted for pass/fail determination. The only one that we have right now is `ThresholdPredicate`, which allows specifying that a basic value must be above/equal/below a certain value of the same type.

I added the `indexmap` depedency for the `IndexMap` object, which is an order-preserving HashMap drop-in replacement. In the past I have found this important for JSON dict comparison. I imagine we will want to update the predicates for comparing composite types.

### DIFF
--- a/hipcheck/src/analysis/result.rs
+++ b/hipcheck/src/analysis/result.rs
@@ -1,0 +1,125 @@
+use crate::hc_error;
+use crate::report::Concern;
+use crate::Result;
+use crate::F64;
+
+/// Represents the enhanced result of a hipcheck analysis. Contains the actual outcome
+/// of the analysis, plus additional meta-information the analysis wants to provide to
+/// HipCheck core, such as raised concerns.
+struct HCAnalysisResult {
+	outcome: AnalysisOutcome,
+	concerns: Vec<Concern>,
+}
+
+/// Represents the result of a hipcheck analysis. Either the analysis encountered
+/// an error, or it completed and returned a value.
+enum AnalysisOutcome {
+	Error(HCAnalysisError),
+	Completed(HCAnalysisValue),
+}
+
+/// Enumeration of potential errors that a HipCheck analysis might return. The Generic
+/// variant enables representing errors that aren't covered by other variants.
+enum HCAnalysisError {
+	Generic(crate::error::Error),
+}
+
+/// A HipCheck analysis may return a basic or composite value. By splitting the types
+/// into two sub-enums under this one, we can eschew a recursive enum definition and
+/// ensure composite types only have a depth of one.
+enum HCAnalysisValue {
+	Basic(HCBasicValue),
+	Composite(HCCompositeValue),
+}
+
+/// Basic HipCheck analysis return types
+#[derive(Debug)]
+enum HCBasicValue {
+	Integer(i64),
+	Unsigned(u64),
+	Float(F64),
+	Bool(bool),
+	String(String),
+}
+impl From<i64> for HCBasicValue {
+	fn from(value: i64) -> Self {
+		HCBasicValue::Integer(value)
+	}
+}
+impl From<u64> for HCBasicValue {
+	fn from(value: u64) -> Self {
+		HCBasicValue::Unsigned(value)
+	}
+}
+impl From<F64> for HCBasicValue {
+	fn from(value: F64) -> Self {
+		HCBasicValue::Float(value)
+	}
+}
+impl TryFrom<f64> for HCBasicValue {
+	type Error = crate::Error;
+	fn try_from(value: f64) -> Result<HCBasicValue> {
+		let inner = F64::new(value)?;
+		Ok(HCBasicValue::Float(inner))
+	}
+}
+impl From<bool> for HCBasicValue {
+	fn from(value: bool) -> Self {
+		HCBasicValue::Bool(value)
+	}
+}
+impl From<String> for HCBasicValue {
+	fn from(value: String) -> Self {
+		HCBasicValue::String(value)
+	}
+}
+impl From<&str> for HCBasicValue {
+	fn from(value: &str) -> Self {
+		HCBasicValue::String(value.to_owned())
+	}
+}
+
+/// Composite HipCheck analysis return types
+enum HCCompositeValue {
+	List(Vec<HCBasicValue>),
+	Dict(indexmap::IndexMap<String, HCBasicValue>),
+}
+
+/// The set of possible predicates for deciding if a source passed an analysis.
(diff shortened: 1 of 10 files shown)


alilleybrinker:
There's a new Rust version out as of last Thursday, and I suspect that's the issue with the CI.

hipcheck/src/analysis/result.rs
```diff
@@ -0,0 +1,136 @@
...
+	fn from(value: bool) -> Self {
+		HCBasicValue::Bool(value)
+	}
+}
+impl From<String> for HCBasicValue {
+	fn from(value: String) -> Self {
+		HCBasicValue::String(value)
+	}
+}
+impl From<&str> for HCBasicValue {
```
alilleybrinker, line 83: Could probably consolidate the `From<&str>` and `From<String>` impls with `impl<T: AsRef<str>> From<T> for HCBasicValue`. Right now `From<String>` will take ownership of `String` rather than copying or taking by reference the way it does for everything else, which means callers are likely to need to do their own `clone` anyway, so an `as_ref().to_owned()` chain isn't that big a deal (this is basically a clone).

hipcheck/src/analysis/result.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/// Enumeration of potential errors that a HipCheck analysis might return. The Generic
+/// variant enables representing errors that aren't covered by other variants.
+#[allow(dead_code)]
+pub enum HCAnalysisError {
+	Generic(crate::error::Error),
+}
+
+/// A HipCheck analysis may return a basic or composite value. By splitting the types
```
alilleybrinker, line 33: Small thing: Hipcheck only capitalizes the "H"

hipcheck/src/analysis/result.rs
```diff
@@ -0,0 +1,136 @@
+use crate::hc_error;
+use crate::report::Concern;
+use crate::Result;
+use crate::F64;
+
+/// Represents the enhanced result of a hipcheck analysis. Contains the actual outcome
+/// of the analysis, plus additional meta-information the analysis wants to provide to
+/// HipCheck core, such as raised concerns.
+#[allow(dead_code)]
```
alilleybrinker, line 9: I'd just make a module-level `#[allow(unused)]`, and then remove it when you're done, rather than item-specific `allow` annotations.

hipcheck/src/analysis/result.rs
```diff
@@ -0,0 +1,136 @@
...
+/// This predicate determines analysis pass/fail by whether a returned value was
+/// greater than, less than, or equal to a target value.
+#[allow(dead_code)]
+pub struct ThresholdPredicate {
+	pub value: HCBasicValue,
+	pub threshold: HCBasicValue,
+	pub ordering: std::cmp::Ordering,
+}
+
+fn pass_threshold<T: PartialOrd>(a: &T, b: &T, ord: &std::cmp::Ordering) -> Result<bool> {
```
alilleybrinker, line 111: Why `PartialOrd` bound instead of `Ord`? I think if you went with `PartialOrd` you could avoid the "math failed" error case in this function.

hipcheck/src/analysis/result.rs
```diff
@@ -0,0 +1,136 @@
...
+	a.partial_cmp(b)
+		.ok_or(hc_error!("threshold comparison failed for unknown reason"))
+		.map(|x| x == *ord)
+}
+
+#[allow(dead_code)]
+impl ThresholdPredicate {
+	// @FollowUp - would be nice for this match logic to error at compile time if a new
```
alilleybrinker, line 119: Usual way to do this would be a procedural macro. You could try writing one in `hipcheck_macros` if you like.

alilleybrinker (changes requested):

alilleybrinker:
@j-lanson just some minor changes!

hipcheck/src/analysis/result.rs
j-lanson: When I tried to do this, I got the attached "conflicting implementations" error message. So instaed I just opted to delete the `From<String>` impl, and converters from `String` can just call `as_str()` first.
conflicting_impls_error

alilleybrinker: Ah, overlapping impls. Classic. Sounds like a good solution!

hipcheck/src/analysis/result.rs
```diff
@@ -142,24 +131,21 @@ impl ThresholdPredicate {
 	}
 }
 
-fn pass_threshold<T: PartialOrd>(a: &T, b: &T, ord: &Ordering) -> Result<bool> {
-	a.partial_cmp(b)
-		.ok_or(hc_error!("threshold comparison failed for unknown reason"))
-		.map(|x| x == *ord)
+fn pass_threshold<T: Ord>(a: &T, b: &T, ord: &Ordering) -> bool {
```
alilleybrinker, line 134: I just checked, and `Ordering` impls `Copy`, and because it's only a 3-element enum is likely smaller than a pointer anyway. I recommend changing this to take `Ordering` instead of `&Ordering`, which also means you could remove the dereference from the function body.

alilleybrinker (approved):

alilleybrinker:
My one remaining comment about turning the `&Ordering` param to `Ordering` in `pass_threshold`, can be done in a follow-up PR.

alilleybrinker:
Thanks @j-lanson!

PR #128, opened 2024-06-18
Update the README

alilleybrinker:
This change is dependent on:

- [ ] ~The launch of the Hipcheck website~ (no longer blocking)
- [x] Beginning to publish container images to Docker Hub
- [x] Adding the `hc setup` command to complete setup of Hipcheck

Once those are done, this can be merged to reflect those changes and better introduce people to the project.


alilleybrinker:
This is a more barebones README than the current one. The goal is that more complete information will be on the website, and that we'll augment and improve the README with this as a starting point.

Issue #28, opened 2024-04-25
Add a command to view scoring breakdown


alilleybrinker:
Right now you can configure Hipcheck to produce scores based on weights for different analyses, but you can't really see the calculated percentages for each leaf of the scoring tree, which makes it difficult to understand how scores are actually calculated. It would be really great if there were a command to see this broken down, and ideally also a way to see it in the output (both human and machine readable) when you run analyses.

- [x]  Define new `AnalysisReport` struct to abstract value/threshold predicate and move into Pass/Fail enumeration. 
- [x] Update analyses to use new `AnalysisReport`. 
- [x] Move decision to run/skip analysis out of `X_analysis()` functions and into the HipCheck core
- [x] Change `AnalysisResults` to be a `HashMap` that maps analysis labels to updated `AnalysisReport` objects
- [x] Create `WeightTree` object and provider for getting a normalized version based on config
- [x] Update scoring to combine updated `AnalysisResults` map + normalized `WeightTree` to produce a `ScoreTree`
- [x] Update errored analyses to act as `Fail` for the purpose of score calculation, and notify user of errors
- [ ] Add ability to print what percentage of the total score each analysis accounted for by walking the `ScoreTree`

alilleybrinker:
Hey @vcfxb, with Julian out for a bit, think you could bring this over the finish line? With the introduction of the weight tree abstraction we should have all the underlying mechanisms in place to add the new subcommand. I'm happy to sync to discuss what's left for this.

vcfxb:
Looking into this now -- I think it would probably be good to sync up and discuss what still needs to be done sometime @alilleybrinker

PR #130, opened 2024-06-19
chore: Demonstrate use of hashmap with new result objects for analysis storage

j-lanson:
This PR addresses the 4th task in #28 , showing how the new `results.rs` structs/enums can be used to replace `AnalysisResults` with a `Hashmap<String, HCStoredResult>`, enabling a future dynamic set of analyses to support the plugin architecture.

Particular focus on impl of `struct AltAnalysisResults` and the code savings of the following line:
```let (act_score, outcome) = alt_results.table.get(ACTIVITY_PHASE).unwrap().score();```

I also updated `AnalysisResults::activity` to be of type `Option<HCStoredResults>`. Since this makes it so the compiler cannot know the type is really `ThresholdPredicate` inside `dyn HCPredicate`, I had to do some `Any` wizardry and add `as_any()` as a trait func for `HCPredicate` so `report_builder.rs` could downcast it as such: `let Some(pred) = analysis.as_any().downcast_ref::<ThresholdPredicate>()`. This may be able to be removed in the future if new reporting infrastructure removes then need to access particular predicate fields.

### CODE REVIEW

File: hipcheck/src/analysis/result.rs [hunk 1]
```diff
@@ -102,8 +112,9 @@ pub enum HCCompositeValue {
 }
 
 /// The set of possible predicates for deciding if a source passed an analysis.
-pub trait HCPredicate: Display + Clone + Eq + PartialEq {
+pub trait HCPredicate: Display + std::fmt::Debug + std::any::Any + 'static {
 	fn pass(&self) -> Result<bool>;
+	fn as_any(&self) -> &dyn std::any::Any;
```
alilleybrinker, line 117: Why the `as_any` method and the new `Any` bound?

hipcheck/src/analysis/report_builder.rs
```diff
@@ -28,20 +28,26 @@ pub fn build_report(session: &Session, scoring: &ScoringResults) -> Result<Repor
...
-				return Err(hc_error!("activity analysis has a non-u64 value"));
-			};
-			builder.add_analysis(Analysis::activity(value, thresh), concerns.clone())?;
-		}
-		Some((Err(error), _concerns)) => {
-			builder.add_errored_analysis(AnalysisIdent::Activity, error);
+	if let Some(stored) = &scoring.results.activity {
+		match &stored.result {
+			Ok(analysis) => {
+				let Some(pred) = analysis.as_any().downcast_ref::<ThresholdPredicate>() else {
```
alilleybrinker, line 34: Maybe I've missed something, it looks like you have a `ThresholdPredicate`, use `as_any` to get an `Any` trait object, then immediately downcast it back to the original type. Why the maneuver?

j-lanson: I have a `Rc<dyn HCPredicate>`, not a `Rc<ThresholdPredicate>`, so the type is not known at compile-time.

I changed `AnalysisResults` to store `HCStoredResults` instead of `ThresholdPredicate` for consistency with the new `AltAnalysisResult` functionality. It cleans up the code in `score.rs` further.

hipcheck/src/analysis/result.rs
j-lanson: The goal is so that I can downcast `Rc<dyn HCPredicate>` to an implementing type of my choosing in `report_builder.rs`. Even though I added `Any + 'static` to `HCPredicate`, I couldn't get it working without the `as_any()` function requirement.

hipcheck/src/analysis/report_builder.rs
j-lanson: If I try something like this:
```
let p: Rc<dyn HCPredicate> = analysis.clone();
let x = p.downcast::<ThresholdPredicate>();
```

I get `method not found in Rc<dyn HCPredicate>` even though it extends `Any`

hipcheck/src/analysis/result.rs
alilleybrinker: Right, I see. I think I'd probably prefer a refactor to use an enum instead of dyn trait approach here. Since we statically know all the predicate types (it's a bounded set), that transformation is possible. See: https://github.com/alilleybrinker/notes/blob/main/implementation-selection.md

j-lanson: Sure. just pushed changes to switch to using an enum. Only thing is now there isn't as much reason for the `HCPredicate` trait to exist. Should it get removed?

alilleybrinker (approved):

hipcheck/src/analysis/result.rs
alilleybrinker: It can be removed. Feel free to do that in a follow-up PR.

PR #131, opened 2024-06-19
Initial work on performance

vcfxb:
This PR includes several commits, but the most major change is that `Rc` has been replaced with `Arc` throughout much of the repository, in order to prepare for improvements from multithreading.


alilleybrinker:
Looks like there are just some unused-related warnings, plus Conventional Commits issues, and then it's ready to merge. Probably just squash down to one commit and reword the message to match Conventional Commits.

vcfxb:
@alilleybrinker assuming this last check is successful (it should be) I think this is ready to merge!

alilleybrinker (approved):

hipcheck/src/cli.rs
```diff
@@ -388,6 +390,19 @@ pub enum Commands {
...
 	Ready,
+
+	/// Unstable command to aid in benchmarking
+	#[command(hide = true)]
+	Unstable(UnstableArgs),
+}
+
+// If no subcommand matched, default to use of '-t <TYPE> <TARGET' syntax. In
+// this case, `target` is a required field, but the existence of a subcommand
+// removes that requirement
```
j-lanson, line 401: If this comment was intentionally duplicated, is there a good reason for doing so?

hipcheck/src/unstable.rs
```diff
@@ -0,0 +1,64 @@
...
+	cli::{CliConfig, UnstableArgs},
+	session::session::Check,
+	shell::{ColorChoice, Output},
+};
+use std::io::Write;
+use std::{io, time::Instant};
+use tempdir::TempDir;
+
+/// Handle unstable commands.
+pub fn main(args: UnstableArgs, config: &CliConfig) -> io::Result<()> {
```
j-lanson, line 11: I would stylistically opt for a different name, such as `run` or `run_unstable`, but not a big deal

j-lanson (commented):
I for one welcome our `Arc` overlords

hipcheck/src/cli.rs
vcfxb: This was unintentional! Probably happened while merging. Thanks for catching this!

PR #133, opened 2024-06-20
Improve performance of grapheme frequency calculation

vcfxb:
This PR improves the performance of grapheme frequency calculation by replacing several single threaded operations with parallel operations using `rayon`. Additionally we switch from using `unicode_segmentation` to `finl_unicode` to improve performance of grapheme cluster identification and iteration. On my machine, this all results in about a 50% total runtime reduction when checking `npm/express`.


alilleybrinker (approved):

alilleybrinker:
Thanks @vcfxb!

PR #134, opened 2024-06-20
feat: parallelize grapheme collection across files

j-lanson:
Replace `for file_diff in &commit_diff.diff.file_diffs {` in the original impl with a `rayon` parallelization.

First have to run through the file list in serial since `salsa db` doesn't impl `Send + Sync`, to filter out non-source files and files with empty patches. But then can parallelize the rest of the computation.

Unclear if this actually provides savings in practice, but theoretically it should.


hipcheck/src/metric/entropy.rs
```diff
@@ -150,8 +152,70 @@ fn is_likely_source_file(
...
+		.diff
+		.file_diffs
+		.iter()
+		.filter(|file_diff| {
+			// Filter out any that are probably not source files, or are empty patches
+			match db.is_likely_source_file(Arc::clone(&file_diff.file_name)) {
+				Ok(true) => file_diff.patch.is_empty().not(),
+				_ => false,
+			}
+		})
```
vcfxb, line 170: This section is the reason I avoided parallelizing over files -- any `Err` returned here just gets filtered out of the calculation rather than being reported.

j-lanson: Just pushed a fix to this (swapping `filter` for `filter_map`) to propagate errors from `db.is_likely_source_file`. and return an error before we get to grapheme calculation

alilleybrinker:
@j-lanson looks like this needs a rebase.

I'll also note that our Salsa database _can_ and _will_ be `Send` and `Sync` in the future once all the types it contains are `Send` and `Sync`. This is part of the design of `salsa`, and once we've achieved it I would recommend we add some code to compile-time ensure we don't break `Send`/`Sync`-ness.

hipcheck/src/metric/entropy.rs
```diff
@@ -150,8 +152,78 @@ fn is_likely_source_file(
...
+		.collect();
+	tgt_diffs?
+		.par_iter()
+		.map(|file_diff| {
+			// Count the number of graphemes in this patch, add it to the total,
+			// and track the number of each grapheme.
+			file_diff
+				.patch
+				// Iterate in parallel over the lines of the patch.
+				.par_lines()
```
vcfxb, line 187: It's probably preferable to `flat_map()` through to `par_lines()` here, if we're doing everything in parallel.

alilleybrinker:
Closing in favor of #146, which has the same contents.

PR #136, opened 2024-06-20
Crate features for benchmarking and new `Drop` based benchmarking struct.

vcfxb:
This PR finishes stripping out the old benchmarking `unstable` subcommand, and replaces it by adding less intrusive `Drop` based benchmarking throughout functions of interest. This will be invisible to the end user, as it is all hidden by a compiler flag that is off by default.


j-lanson (commented):
very slick, I like it!

alilleybrinker:
In the short term I think this works. In the longer term I think we'll likely want to end up with something similar to the timing infrastructure in Cargo.

Cargo works by having a job queue of build jobs, each of which gets timed. If timing isn't enabled, then the timing calls become no-ops. This is all turned on or off at runtime rather than being a compile time flag. So something like `hc check --timing` (or perhaps `hc check --unstable:timing` or some similar form while it's still unstable).

alilleybrinker (approved):

PR #146, opened 2024-06-24
feat: Parallelize grapheme collection across files.

alilleybrinker:
This is #134, but reworked to be onto `main` instead of `venus/performance`, and to not have the many merge conflicts that branch had on both `venus/performance` and `main`.

### DIFF
--- a/hipcheck/src/metric/entropy.rs
+++ b/hipcheck/src/metric/entropy.rs
@@ -158,7 +158,6 @@ fn grapheme_freqs(commit_diff: &CommitDiff, db: &dyn MetricProvider) -> Result<C
 	let grapheme_table: DashMap<String, u64> = DashMap::new();
 
 	// Use this variable to track the total number of graphemes accross all patches in this commit diff.
-	let mut res: Vec<usize> = vec![];
 	let tgt_diffs: Result<Vec<&FileDiff>> = commit_diff
 		.diff
 		.file_diffs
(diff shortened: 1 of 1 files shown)


hipcheck/src/metric/entropy.rs
```diff
@@ -158,52 +158,62 @@ fn grapheme_freqs(commit_diff: &CommitDiff, db: &dyn MetricProvider) -> Result<C
...
+		.collect();
+	tgt_diffs?
+		.par_iter()
+		.map(|file_diff| {
+			// Count the number of graphemes in this patch, add it to the total,
+			// and track the number of each grapheme.
+			file_diff
+				.patch
+				// Iterate in parallel over the lines of the patch.
+				.par_lines()
```
vcfxb, line 189: We should probably `flat_map` to `par_lines` here, if we're trying to do everything in parallel.

hipcheck/src/metric/entropy.rs
```diff
@@ -158,52 +158,62 @@ fn grapheme_freqs(commit_diff: &CommitDiff, db: &dyn MetricProvider) -> Result<C
...
-							grapheme_table.insert(grapheme.to_owned(), 1);
-						}
-					})
-					// get the grapheme count for this normalized line.
-					.count()
-			})
-			.sum::<usize>();
-	}
+	// Use this variable to track the total number of graphemes accross all patches in this commit diff.
+	let total_graphemes: usize = res.into_iter().sum();
```
vcfxb, line 216: Instead of `collect_into_vec` and then turning the `res` vec into an iterator and summing, we should just sum directly here (and avoid allocating a `res` vec at all).

vcfxb (changes requested):

alilleybrinker:
Looks like this is ready for re-review @vcfxb

vcfxb (approved):
LGTM!

PR #149, opened 2024-06-24
RFD #4: Plugin API

### LINKED ISSUE #11: Write the "Plugin Architecture API" RFD
Once we've characterized the expected end state for Hipcheck's new plugin architecture, we ought to then specify what the API would look like. This involves multiple "parts" of the API, which I will decompose here, taking [inspiration from the design of `dtrace`](https://opensource.apple.com/source/xnu/xnu-3789.41.3/bsd/sys/dtrace.h.auto.html) (the dynamic tracing tool originally designed for Solaris at Sun, now ported to a number of other kernels and operating systems). The `dtrace` implementation breaks down the following parts of the API:

* `dtrace`-to-Provider ("Provider" is what `dtrace` calls their plugins, which enable instrumentation of the system)
    * This is the API that `dtrace` exposes to Providers
* Provider-to-`dtrace`
    * This is how Providers register themselves with `dtrace` and tell `dtrace` how to use them

To analogize to Hipcheck, this would basically become:

* Hipcheck-to-Plugin
    * When a plugin is registering itself, the plugin would provide a structure of functions to Hipcheck which describe the operations necessary for performing the operations provided by the plugin.
* Plugin-to-Hipcheck
    * The API Hipcheck would expose to plugins to enable registration of new plugins with Hipcheck

`dtrace` also has a mechanism for "meta providers" which is how `dtrace` supports Userland Statically-Defined Tracing (USDT), for which there wouldn't be an analogue in our context.

To put the above breakdown another way, there needs to be:

* A way for plugins to be registered, so Hipcheck knows about them and can call them.
* A way for plugins to enable Hipcheck to call the functions the plugins provide.

Note that this RFD should still be agnostic about the specific way that plugins are implemented. We'll want to explore multiple options for implementing plugins, including:

* WebAssembly
* Separate binaries in their own subprocess with inter-process communication between them and Hipcheck
* C FFI with dynamic linking

Exploration of these options will be addressed by other RFD's.

alilleybrinker:
Closes #11.

Edit: **[__Rendered__](https://github.com/mitre/hipcheck/blob/416530d6e496d7c9a0ef756c7ec32292063bd529/site/content/rfds/0004-plugin-api.md)**

### DIFF
--- a/site/content/rfds/0004-plugin-api.md
+++ b/site/content/rfds/0004-plugin-api.md
@@ -1,0 +1,922 @@
+
+# Plugin API
+
+While [RFD #3][rfd_3] addressed the overall vision for how the planned plugin
+architecture for Hipcheck will work, the intent of this RFD is to describe in
+greater detail the specifics of the interface between the `hc` binary and
+Hipcheck plugins.
+
+The goals in writing this RFD are:
+
+- To flesh out details of the design of Hipcheck's plugin system amongst
+  the Hipcheck dev team itself, by providing a point of coordination and
+  discussion.
+- To provide guidance to prospective plugin authors about what will be
+  required to write a Hipcheck plugin, and simultaneously to empower those
+  authors with enough knowledge to give feedback to the Hipcheck team about
+  what they think won't work or will be challenging in the planned
+  architecture.
+- To inform the Hipcheck development team in some detail about what they need
+  to build into Hipcheck to realize the goal of supporting third-party plugins.
+
+Thus, moving forward this RFD has two audiences: __Hipcheck's Developers__,
+and __Potential Plugin Authors__. The remainder of the RFD will be written
+with _both_ of these audiences in mind.
+
+The remainder of this RFD is written with the assumption that you have already
+read RFD #3.
+
+## Defining a Plugin
+
+### Plugin Materials
+
+One of the goals of Hipcheck's plugin system is that anyone can create a
+Hipcheck plugin in _any_ programming language, not just Rust. This creates
+challenges around distributing and executing plugins which can run on any
+of Hipcheck's supported architectures. While some programming languages
+are intended to be platform-agnostic (where the code as-distributed can
+be run on any platform that has a usable runtime for the language), others
+must be specifically compiled for the target architecture. We intend for
+Hipcheck plugins to be usable in either case.
+
+Defining a Hipcheck plugin means defining three components:
+
+- The plugin's executable artifact
+- The plugin manifest
+- The plugin discovery manifest
+
+Let's cover each of these in turn.
+
+The plugin's __executable artifact__ is the binary, set of executable
+program files, Docker container, or other artifact which can be run as a
+command line interface program through a singular "start command" defined
+in the plugin's manifest.
+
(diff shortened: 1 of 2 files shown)


site/content/rfds/0004-plugin-api.md
```diff
@@ -0,0 +1,932 @@
...
+```kdl
+publisher "mitre"
+name "affiliation"
+version "0.1.0"
+license "Apache-2.0"
+entrypoint {
+  on arch="aarch64-apple-darwin" "./hc-mitre-affiliation"
+  on arch="x86_64-apple-darwin" "./hc-mitre-affiliation"
+  on arch="x86_64-unknown-linux-gnu" "./hc-mitre-affiliation"
+  on arch="x86_64-pc-windows-msvc" "./hc-mitre-affiliation"
```
mchernicoff, line 119: Just confirming that that would be the correct entrypoint even on Windows.

site/content/rfds/0004-plugin-api.md
```diff
@@ -0,0 +1,932 @@
...
+
+The RFD so far has described how plugins are defined by plugin creators, and
+the API expected of plugins in order for Hipcheck to interact with them. What
+has _not_ been described is how end-users of Hipcheck should use plugins
+themselves. This section answers that question.
+
+### Policy Files
+
+Today, Hipcheck determines what analyses to run and how to configure those
+analyses a "configuration file," and when you first install Hipcheck you
```
tpepper-gov, line 393: sentence fragment missing word(s?)

site/content/rfds/0004-plugin-api.md
```diff
@@ -0,0 +1,932 @@
...
+
+The __plugin discovery manifest__ is an additional manifest which tells
+Hipcheck where to find the correct plugin artifact for a specific target
+architecture. In the future, Hipcheck's plugin system may support one or
+more plugin registries which ease discovery of Hipcheck plugins. In the
+short term, the standard mechanism for discovery of Hipcheck plugins
+will be for plugin producers to distribute a per-plugin discovery manifest
+which contains version and target architecture information necessary for
+Hipcheck to download the _correct_ plugin executable artifact to run
+on the end-user's host.
```
j-lanson, line 135: Do we want to consider at this time the possibility of some users wanting to run in a fully-offline mode with plugins they already have on-disk?

site/content/rfds/0004-plugin-api.md
alilleybrinker: Thanks! I do think that we'll likely want to enhance the target triple stuff to permit some "grouping expressions" when granularly listing every target triple is too tedious. Rust has a version of this in the `cfg` system where you can just say "windows" for example and cover _all_ windows targets.

site/content/rfds/0004-plugin-api.md
alilleybrinker: Ah good catch!

site/content/rfds/0004-plugin-api.md
alilleybrinker: Good point! I think this would properly go in the "Running Plugins" section.

mchernicoff (approved):

PR #152, opened 2024-06-25
Add `hc setup` installation command

### LINKED ISSUE #103: Implement `hc setup` subcommand to complete installation of Hipcheck

j-lanson:
Resolves #103.

This PR adds the `hc setup` CLI command for finishing the installation of an `hc` binary. This primarily includes:
-  acquiring target locations for the `HC_CONFIG, HC_DATA` directories. 
    - creating those directories if they don't exist
- finding a suitable local copy of a `cargo-dist` release bundle for Hipcheck
  - this could be a dir, zip, or tar.xz file
  - if none is found, one will be automatically downloaded from our Hipcheck release page on github
- decompressing the source bundle if applicable, and copying the contents of the source `config/, scripts/` dirs to `HC_CONFIG, HC_DATA`, respectively.
- cleaning up any downloaded files or uncompressed directories created in service of the above operations

The cmdline is `hc setup [--offline] [SOURCE]`. If `--offline` provided, we won't try to pull the release bundle from github as a last resort when failing to find a local copy. `SOURCE` is an optional positional arg for the user to explicitly specify a directory/file that is the release bundle. 

**New Dependencies**
In order to support decompressing `.zip` and `.tar.xz` files, I added dependencies on `tar`, `zip`, and `xz2` crates. An alternative design decision could be to use `process::Command` and expect these utilities exist on the host machine, but I opted for adding libraries so we avoid OS-specific considerations.

**setup.rs::try_resolve_source_path()**
See code comments for overview. I chose to make it so this source resolution system does not take into account target triple of the release. Since we are just copying files (no executables) from dirs, I figured it shouldn't matter. When pulling from github, it does use the correct version, but otherwise it doesn't check the version of any local releases found. 

**setup.rs::SetupSourcePath**
Designed to be a representation of a file/dir that can be used for copying `config`, `scripts` dirs. `try_unpack()` is for decompressing archives into dirs so we can do the actual subdir copying. `delete: bool` allows us to track what should and should not be deleted. For instance, if we find a local `hipcheck` bundle dir already decompressed, we should use it but not touch it. However, if we need to download the release bundle from Github, we should delete both the downloaded archive and the folder that it gets decompressed into. We could consider replacing `delete()` with `impl Drop` in the future, but I thought that was maybe more fancy than useful.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,23 @@
 version = 3
 
 [[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
+name = "aes"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
+dependencies = [
+ "cfg-if",
+ "cipher",
+ "cpufeatures",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


alilleybrinker:
Excited to test this tomorrow! We should try to test on different platforms. I have macOS on hand, and can easily spin up a Linux container. If anyone has Windows that would be good to check too.

alilleybrinker (changes requested):
This is looking great! Some small changes for the user experience:


- `--offline` should not have a short flag. `-o` is well-understood to be a common short flag for an output flag, and we don't want to confuse this norm. If people want to specify it, they can do so with the long flag.
- Source should be a flag, not a positional argument. It should be `-s`/`--source`.
- `hc setup`'s help text should make clear that the `--config` and `--data` directories will be used as destinations for the files being set up.
- The error message for when `hc setup` can't find a local copy and can't pull from remote because of the `--offline` flag should be clearer.
	- Right now it says "could not find suitable source file"

mchernicoff:
No obvious errors running it on Windows.

alilleybrinker (approved):
Going to merge and then open a PR with my minor UI fixes.

PR #155, opened 2024-06-26
feat: Slightly modified arguments for `hc setup`.

alilleybrinker:
This is intended to make it a little bit clearer to use, specifically by converting the positional argument into a flag, and by adding more explanatory text.

### DIFF
--- a/hipcheck/src/cli.rs
+++ b/hipcheck/src/cli.rs
@@ -389,6 +389,9 @@ pub enum Commands {
 	/// Print the JSON schema for output of a specific `check` command.
 	Schema(SchemaArgs),
 	/// Initialize Hipcheck config file and script file locations.
+	///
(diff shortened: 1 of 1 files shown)


hipcheck/src/cli.rs
```diff
@@ -389,6 +389,9 @@ pub enum Commands {
 	/// Print the JSON schema for output of a specific `check` command.
 	Schema(SchemaArgs),
 	/// Initialize Hipcheck config file and script file locations.
+	///
+	/// NOTE: The `--data` and `--config` flags specify where to install the
+	/// extra files Hipcheck needs.
```
j-lanson, line 394: Should this comment explain that the `data` and `config` paths can also be resolved by env var or platform default?

j-lanson (approved):
Looks good!

PR #156, opened 2024-06-26
Initial exploration of KDL for configuration

alilleybrinker:
This is not yet ready for review.


alilleybrinker:
Closing this, as it's diverged pretty heavily and was generally experimental. With RFD <span>#</span>4 approved now, we'll be working on implementing this in a different way than that explored here.

PR #157, opened 2024-06-27
chore: refactor all remaining repo analyses

j-lanson:
Continuation of effort on #28 .

This PR applies the refactor demonstrated in #132 to all repo analyses (PR analyses will be removed in a subsequent commit following @alilleybrinker 's guidance).

The number of lines changed seems daunting, but its really just the same pattern:
- Update analysis function and storage to return and store `HCAnalysisReport` and `HCStoredResult`, respectively
- Update analysis function content to not check for `db.X_active()`, and just return the calculated value instead of checking own threshold and scoring by itself
- Remove analysis type from `score.rs::phase_outcome()`
- Change analysis execution in `score.rs::score_results()` to use new macro
- Change processing of results for analysis in `report_builder.rs`

Of the above, ensuring that the analysis functions themselves is correct is probably the most worthwhile thing to examine. However, I've verified that the generated score tree graph is the same, minus the fact that errored analyses are now treated as failures. 

Other Misc Changes:
- Changed behavior of `pass_threshold()` in `result.rs` to treat `Ordering::Less` threshold spec as `<=`. This keeps its behavior the same as the existing analysis infrastructure. Without this, `0, 0, Ordering:Less` would return a failure.

### DIFF
--- a/hipcheck/src/analysis/analysis.rs
+++ b/hipcheck/src/analysis/analysis.rs
@@ -39,25 +39,25 @@ pub trait AnalysisProvider:
 	fn affiliation_analysis(&self) -> Arc<HCAnalysisReport>;
 
 	/// Returns result of binary analysis
-	fn binary_analysis(&self) -> Result<Arc<AnalysisReport>>;
+	fn binary_analysis(&self) -> Arc<HCAnalysisReport>;
 
 	/// Returns result of churn analysis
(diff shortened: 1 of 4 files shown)


alilleybrinker:
Love to see a diff that removes so many lines! Tested out locally and the analyses all seem to work as intended. Thanks Julian!

PR #158, opened 2024-06-27
chore: remove pr analysis and scoring infrastructure

j-lanson:
While working on the scoring infrastructure refactor, I encountered the fact that PR analyses have their own `PrConcern` type separate from `Concern` used by repo analyses. In discussing with @alilleybrinker he advised to remove the PR analysis infrastructure entirely. This PR demonstrates what that would look like.


alilleybrinker (dismissed):
The nice thing is of course that Git retains all of this so if for some reason we wanted to add this back in the future, we could.

alilleybrinker (approved):

PR #170, opened 2024-07-01
Weight Tree Printing

### LINKED ISSUE #28: Add a command to view scoring breakdown

vcfxb:
Provide CLI functionality to print the resolved weights used in analysis. 

Closes #28

### DIFF
--- a/hipcheck/src/analysis/score.rs
+++ b/hipcheck/src/analysis/score.rs
@@ -147,11 +147,12 @@ fn normalize_st_internal(node: NodeId, tree: &mut Arena<ScoreTreeNode>) -> f64 {
 }
 
 #[derive(Debug, Clone)]
-pub struct AltScoreTree {
+pub struct ScoreTree {
 	pub tree: Arena<ScoreTreeNode>,
 	pub root: NodeId,
 }
-impl AltScoreTree {
+
+impl ScoreTree {
 	pub fn new(root_label: &str) -> Self {
 		let mut tree = Arena::<ScoreTreeNode>::new();
 		let root = tree.new_node(ScoreTreeNode {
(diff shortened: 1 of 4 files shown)


alilleybrinker:
Testing locally now!

alilleybrinker (changes requested):
Rename to `hc scoring`, otherwise LGTM!

alilleybrinker (approved):

vcfxb:
It took 2 attempts at squashing and force-pushing but I think this is finally properly squashed and ready to merge.

alilleybrinker (approved):

Issue #56, opened 2024-05-08
Incorporate self-updating with new `hc update` command.


alilleybrinker:
`cargo-dist` automatically generates a self-updater, and it would be nice if we could leverage it, if present, to allow an easy path for users to self-update Hipcheck.

The basic idea would be to add an `hc update` command which uses the auto-generated and installed `hc-update` CLI produced and distributed automatically with the installer `cargo-dist` produces.

See here for more: https://opensource.axo.dev/cargo-dist/book/installers/updater.html

alilleybrinker:
Updating here to note that @mchernicoff has been working on this extensively. Patches are prepped and we're engaged with the Axodotdev folks upstream. They've indicated they want to resolve this issue in their `axoasset` crate which all of their crates use, and we're now waiting for their resolution.

PR #176, opened 2024-07-03
feat: Adds hc update command

mchernicoff:
Resolves issue #56  Adds `hc update` command to run the self-updater (if installed). This updater comes with the current Hipcheck installer as created by `cargo-dist` using the `axoupdater` crate. The `hc-update` command simply calls the `axoupdater` provided updater, but integrated into the regular Hipcheck CLI. 

This command automatically brings Hipcheck up to date with the latest released version if run without any flags. The user can includes optional flags to update to versions other than the latest.

N.B. Because of a bug in `cargo-dist`, we cannot be sure if the `axoupdater` updater will be installed as `hc-update` or `hipcheck-update`. Therefore the new `hc update` command tries both commands. If this bug is resolved in the future, we can remove the step of checking both possible commands.

Also updates the version of `cargo-dist` to 0.17.0.

### DIFF
--- a/hipcheck/src/main.rs
+++ b/hipcheck/src/main.rs
@@ -47,6 +47,7 @@ use cli::FullCommands;
 use cli::SchemaArgs;
 use cli::SchemaCommand;
 use cli::SetupArgs;
+use cli::UpdateArgs;
 use command_util::DependentProgram;
 use config::WeightTreeNode;
 use config::WeightTreeProvider;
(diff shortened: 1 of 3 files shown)


hipcheck/src/main.rs
```diff
@@ -608,6 +612,46 @@ fn cmd_ready(config: &CliConfig) {
 	}
 }
 
+/// Run the Hipcheck self-updater to update to the latest release version.
+/// If the updater is not found, returns an error
+fn cmd_update(args: &UpdateArgs) {
+	// Try "hc-update" as the updater command, and see if it runs
+	let mut hc_command = Command::new("hc-update");
```
alilleybrinker, line 619: It might be easier to use `which::which` here rather than trying the command and then trying the other name if it fails.

alilleybrinker:
@mchernicoff it looks like a newline was mistakenly added to the bottom of the `cargo-dist` GitHub Action file, which is causing a CI error.

mchernicoff:
> @mchernicoff it looks like a newline was mistakenly added to the bottom of the `cargo-dist` GitHub Action file, which is causing a CI error.
Saw that. Fixing it in next commit.

hipcheck/src/main.rs
mchernicoff: Changed the code to do it that way.

alilleybrinker (approved):

PR #178, opened 2024-07-03
chore: Update cargo-dist to version 0.17.0

mchernicoff:
Changes the version of `cargo-dist` used by Hipcheck to version 0.17.0. Makes changes in `release.yml` to make it compatible with that version.


mchernicoff:
@alilleybrinker Take a look at the changes to `release.yml` and let me know where we might be throwing the baby out with the bathwater.

alilleybrinker (approved):

PR #198, opened 2024-07-08
feat: Shell rework

vcfxb:
This PR contains an early but hopefully mostly functional draft of the new hipcheck shell. 

This is marked as a draft because there are almost certainly bugs that I haven't gotten around to fixing yet, but it should be in a state where people can start finding them and contributing commits to fix them or letting me know about them. Once enough of the bugs are cleared, we can merge this, and then have other pull requests to start converting various phases and analyses to use the new `SpinnerPhase` and `ProgressPhase` structs. 

This includes all the changes from #195, which I'm assuming will get merged before this.

### DIFF
--- a/hipcheck/Cargo.toml
+++ b/hipcheck/Cargo.toml
@@ -31,7 +31,6 @@ duct = "0.13.5"
 env_logger = { version = "0.11.3" }
 graphql_client = "0.14.0"
 hipcheck-macros = { path = "../hipcheck-macros", version = "0.3.0" }
-libc = "0.2.155"
 log = "0.4.22"
 maplit = "1.0.2"
 nom = "7.1.3"
@@ -55,7 +54,6 @@ serde_derive = "1.0.137"
 serde_json = "1.0.119"
 smart-default = "0.7.1"
 spdx-rs = "0.5.0"
-termcolor = "1.1.3"
 toml = "0.8.14"
 unicode-normalization = "0.1.19"
 ureq = { version = "2.10.0", default-features = false, features = [
@@ -81,6 +79,8 @@ zip = "2.1.3"
 xz2 = "0.1.7"
 indextree = "4.6.1"
 num-traits = "0.2.19"
+console = { version = "0.15.8", features = ["windows-console-colors"] }
+term_size = "0.3.2"
 
 # Exactly matching the version of rustls used by ureq
 # Get rid of default features since we don't use the AWS backed crypto provider (we use ring).
@@ -94,16 +94,6 @@ features = [
     "tls12",
     "ring"
 ]
-
-# Windows-specific dependencies
-[target.'cfg(windows)'.dependencies]
-winapi = { version = "0.3.9", features = [
-    "handleapi",
-    "processenv",
-    "winbase",
-    "wincon",
-    "winnt",
-] }
 
 [build-dependencies]
 anyhow = "1.0.86"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -759,6 +759,7 @@ dependencies = [
  "anyhow",
  "chrono",
  "clap",
+ "console",
  "content_inspector",
  "dashmap",
  "dirs",
@@ -773,7 +774,6 @@ dependencies = [
  "indextree",
  "indicatif",
  "indicatif-log-bridge",
- "libc",
  "log",
  "maplit",
  "nom",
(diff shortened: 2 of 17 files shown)


alilleybrinker:
Looking forward to testing this later today!

alilleybrinker:
Just tested it out and this is very cool! Will discuss more at today's sync :smile:

vcfxb:
Let me know if/when this should move out of "draft" status -- I think the remaining things to do are:

- [ ] Ensure progress bars are hidden when verbosity is set to silent.
- [ ] (optionally) create a silence guard type that sets the verbosity to silent when instantiated, then returns it to it's previous state when dropped. 
- let me know if there is anything else.

alilleybrinker:
> Let me know if/when this should move out of "draft" status -- I think the remaining things to do are:
> 
> * [ ]  Ensure progress bars are hidden when verbosity is set to silent.
> * [ ]  (optionally) create a silence guard type that sets the verbosity to silent when instantiated, then returns it to it's previous state when dropped.
> * let me know if there is anything else.

I think those are the blockers, yeah. Once that's done I think we can merge and then make further refinements from there.

vcfxb:
@mitre/hipcheck-maintainers This PR now has the last two items need to come out of draft and be ready for full review!

j-lanson:
when trying to run `cargo build` I get the following:
```
error[E0658]: associated type bounds are unstable
   --> hipcheck/src/shell/par_iter.rs:306:3
    |
306 |         Tracker: ExactSizeIterator + DoubleEndedIterator + Iterator<Item = T>,
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #52662 <https://github.com/rust-lang/rust/issues/52662> for more information

For more information about this error, try `rustc --explain E0658`.
```
Has this been encountered yet?

vcfxb:
> when trying to run `cargo build` I get the following:
> 
> ```
> error[E0658]: associated type bounds are unstable
>    --> hipcheck/src/shell/par_iter.rs:306:3
>     |
> 306 |         Tracker: ExactSizeIterator + DoubleEndedIterator + Iterator<Item = T>,
>     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>     |
>     = note: see issue #52662 <https://github.com/rust-lang/rust/issues/52662> for more information
> 
> For more information about this error, try `rustc --explain E0658`.
> ```
> 
> Has this been encountered yet?

Haven't seen that on my end -- what version of `rustc` are you using?

j-lanson:
> Haven't seen that on my end -- what version of `rustc` are you using?

rustc 1.77.2

alilleybrinker:
Ah, Rust recently stabilized associated type bounds like Venus has done here. In general our policy today is that we track stable, so we technically no longer support 1.77.0. Maybe it's worth revisiting that policy, but since we generally try to focus on our pre-built binaries rather than build from source for users, I'm not especially worried.

In the future this could become an issue if we wanted to pursue packaging with system package ecosystems like Debian `apt`, where they run older Rust toolchains.

hipcheck/Cargo.toml
```diff
@@ -54,7 +53,7 @@ serde_derive = "1.0.137"
 serde_json = "1.0.119"
 smart-default = "0.7.1"
 spdx-rs = "0.5.0"
-termcolor = "1.1.3"
+# termcolor = "1.1.3"
```
alilleybrinker, line 56: If we're not using it, I think it's better to remove fully rather than comment out.

alilleybrinker (changes requested):
One small night. Otherwise LGTM. Ping me after the fix and I'll give it an approval.

hipcheck/Cargo.toml
vcfxb: Fixed!

j-lanson (commented):
Looks good! I'll leave explicit approval to Andrew

hipcheck/Cargo.toml
alilleybrinker: Thanks!

alilleybrinker (dismissed):
I think all that’s left is squashing to one commit!

mchernicoff (approved):

Issue #185, opened 2024-07-04
Enable users to provide "VCS URLs" as targets


alilleybrinker:
The `pip` package manager from the Python ecosystem specifies "VCS URLs," which actually appear in many other contexts and tools as a sort of informal specification. These URLs are a way of encoding some source repository information in a URL in a way that actually closely matches the information we care about for Hipcheck source repository targets anyway. It would be good to explicitly add support for these as targets in Hipcheck.

https://pip.pypa.io/en/stable/topics/vcs-support/

PR #209, opened 2024-07-16
feat(CLI): CLI can infer target from VCS URL

mchernicoff:
Resolves issue #185 

`hc check` will now attempt to resolve the target repo from a given target VCS URL (see [here](https://pip.pypa.io/en/stable/topics/vcs-support/). Currently we support remote Git repo URLs and local Git repo filepaths. If a target type can be resolved, Hipcheck will extract the information it needs to run from the VCS URL.

We currently ignore any additional Git ref information that may be provided with the VCS URL. We assume that if such a ref is given, it follows the VCS URL convention of `git+https://github.com/owner/repo.git@ref`, with the "`.git`" suffix provided before the "@".

### DIFF
--- a/hipcheck/src/cli.rs
+++ b/hipcheck/src/cli.rs
@@ -985,6 +985,39 @@ mod tests {
 	}
 
 	#[test]
+	fn test_deductive_check_repo_vcs_https() {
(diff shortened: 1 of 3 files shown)


j-lanson (approved):
Looks good!

PR #210, opened 2024-07-16
feat: added types for new target repo spec and resolution

j-lanson:
Adds new types, including `Target` which is designed to replace `Source` going forward. Future work will integrate this type into the Hipcheck analysis, and derive `Target` from `CheckArgs`

Additionally adds the optional `ref_` arg to `CheckRepoArgs` and fixes `trailing_args` to also consume flag arguments. Without this, `--ref` can't be passed when not using deprecated `hc check repo/maven/etc <TARGET>` format.


mchernicoff (approved):
LGTM

PR #221, opened 2024-07-22
fix: Add missing step to install instructions

### LINKED ISSUE #215: Update install guide and quickstart guide to mention `hc setup`
See #214, right now we don't tell the user in our docs to run `hc setup` after installing the `hc` binary, which means they're likely to try running Hipcheck _without_ the necessary additional files set up.

alilleybrinker:
Fixes #215


mchernicoff (approved):
LGTM!

PR #212, opened 2024-07-17
Add `git2` based cloning with a custom `rustls` transport. 

vcfxb:
This also includes a few improvements/additions to the global `Shell` architecture.

PR #222, opened 2024-07-22
feat: Add progress bars when cloning repos

vcfxb:
This supersedes #212 by implementing the functionality downstream -- letting us avoid needing to wait on upstream changes.

This PR includes a shim function to pass trace messages from `libgit2` through our logging infra, a `rustls` + `ureq` based transport for `libgit2` that's faster than the native mac implementation and more verbose in handling errors, and updates to the process of cloning repos to take advantage of `libgit2`'s progress reporting.

Currently, I am still sometimes getting a TLS error on my machine -- If this is not reproducible on other machines then I think we can proceed with merging this PR. This error occurs even when using `git` proper, so I expect it's an issue with my machine more than with this code.


j-lanson (approved):
Looks good! In the future we'll likely want to resolve the `agent()` constructor duplication, but I don't think it's important enough to block the merge

Issue #182, opened 2024-07-04
Add `hc cache` subcommand for managing Hipcheck's cache


alilleybrinker:
Hipcheck creates local clones of every repository it analyzes, because analyzing a local clone is _much_ faster than trying to work over the network. But some repositories can be quite large, and this cache will currently just continue to grow until the user notices it (ideally before something concerning like running out of disk space) and deletes some of the clones.

We ought to add an `hc cache` subcommand to manage the cache, probably by:

- [ ] Listing repositories in the cache, their head commit, when they were last updated, and their size
- [x] Deleting repositories in the cache, either by specifying specific repositories to delete, or doing operations like "delete `n` oldest" or "delete `n` largest"

We may also want to enable configuring Hipcheck to alert when the cache directory passes some threshold. For this we probably want a default threshold, and let the user both change the threshold and to turn alerting off entirely. The alerting could take the form of a warning when Hipcheck is run.

alilleybrinker:
Resolved by #224 and #236.

PR #224, opened 2024-07-23
feat: add `hc cache` subcommand for manually listing/deleting elements of the cache

j-lanson:
This PR contributes to #182 . 

It introduces the `hc cache` command which allows controlled listing/deleting of cache entries (rooted in the `HC_CACHE` path resolved by the standard CLI).

**This PR does not currently take measures to short-cut dir-size calculation, so each operation can take multiple seconds to complete.** It should be discussed whether this is important enough to require before this PR is merged or can be addressed subsequently. 

`cli.rs` - changes to introduce the subcommand.
`main.rs` - call `cmd_cache()` function when CLI dictates
`cache.rs` - supporting types and the `HcCache` object itself, with public-facing functions for listing/deleting entries.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -153,6 +153,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "79296716171880943b8470b5f8d03aa55eb2e645a4874bdbb28adb49162e012c"
 
 [[package]]
+name = "bytecount"
+version = "0.6.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5ce89b21cab1437276d2650d57e971f9d548a2d9037cc231abdc0562b97498ce"
+
+[[package]]
 name = "byteorder"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -644,6 +650,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "fs_extra"
+version = "1.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
+
+[[package]]
 name = "generic-array"
 version = "0.14.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -793,6 +805,7 @@ dependencies = [
  "duct",
  "env_logger",
  "finl_unicode",
+ "fs_extra",
  "git2",
  "graphql_client",
  "hipcheck-macros",
@@ -822,6 +835,7 @@ dependencies = [
  "serde_json",
  "smart-default",
  "spdx-rs",
+ "tabled",
  "tar",
  "tempfile",
  "term_size",
@@ -1249,6 +1263,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "papergrid"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9ad43c07024ef767f9160710b3a6773976194758c7919b17e63b863db0bdf7fb"
+dependencies = [
+ "bytecount",
+ "fnv",
+ "unicode-width",
+]
+
+[[package]]
 name = "parking_lot"
 version = "0.11.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1355,6 +1380,30 @@ name = "ppv-lite86"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5b40af805b3121feab8a3c29f04d8ad262fa8e0561883e7653e024ae4479e6de"
+
+[[package]]
+name = "proc-macro-error"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "da25490ff9892aab3fcf7c36f08cfb902dd3e71ca0f9f9517bea02a73a5ce38c"
+dependencies = [
+ "proc-macro-error-attr",
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+ "version_check",
+]
+
+[[package]]
+name = "proc-macro-error-attr"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a1be40180e52ecc98ad80b184934baf3d0d29f979574e439af5a55274b35f869"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "version_check",
+]
 
 [[package]]
 name = "proc-macro2"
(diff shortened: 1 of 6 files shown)


alilleybrinker:
Reviewing now!

alilleybrinker:
Okay, overall I really like this! I think the operations make sense and the listing UI is really clear.

For deletion, I think the following flag structure would make sense, improve the docs, and improve the compositionality of the options:

- `-f`/`--filter <PATTERN>`: Filter packages to delete based on pattern. This reduces the packages selected by any `--strategy` or `--max` provided.
- `-x`/`--execute`: Actually run the deletion instead of saying which ones _would_ be deleted. (dry run by default)
- `-s`/`--strategy <STRATEGY>`: `largest`, `smallest`, `oldest`, `newest`, `alpha` (alphabetical), `ralpha` (reverse alphabetical), `all`, `none`. Defaults to `none`.
- `-m`/`--max <N>`: Max number of entries to delete. Default varies based on strategy if left unspecified. `largest`/`smallest`/`oldest`/`newest` all default 1. `alpha` and `none` defaults to 0, `all` defaults to no limit.

This would be instead of the more subcommand based options so far. Now this _does_ mean that the user could create some nonsensical combinations which would need to be rejected. Some examples of good and bad:

```sh
$ hc cache delete # Deletes nothing, warns the user that no packages were selected and nothing was done.
$ hc cache delete -x # Not a dry run, but doesn't delete anything. Warns the user.
$ hc cache delete -s oldest --max 2 # Dry-run delete the 2 oldest in the cache.
$ hc cache delete -s oldest -x # Actually delete the 1 oldest.
$ hc cache delete -s alpha -m 1 -x # Actually delete the first alphabetical package
```

@j-lanson what do you think?

alilleybrinker (changes requested):
Forgot to submit a review. Logging a "Request Changes" review just to show not to merge yet.

j-lanson:
> * `-f`/`--filter <PATTERN>`: Filter packages to delete based on pattern. This reduces the packages selected by any `--strategy` or `--max` provided.
> * `-x`/`--execute`: Actually run the deletion instead of saying which ones _would_ be deleted. (dry run by default)
> * `-s`/`--strategy <STRATEGY>`: `largest`, `smallest`, `oldest`, `newest`, `alpha` (alphabetical), `ralpha` (reverse alphabetical), `all`, `none`. Defaults to `none`.
> * `-m`/`--max <N>`: Max number of entries to delete. Default varies based on strategy if left unspecified. `largest`/`smallest`/`oldest`/`newest` all default 1. `alpha` and `none` defaults to 0, `all` defaults to no limit.

I personally find that this is way more complicated and likely to cause confusion for users.
1. The syntax of `cache list` and `cache delete` is different, unless you intend for `cache list` to copy the same syntax, but it wasn't clear in your comment. If not, this makes the usage immediately less intuitive.
2. The fact that the syntax of `cache list/delete` is the same currently, there isn't a need for the concept of a "dry run" because one can run `cache list` with one set of qualifiers, see what would be printed, and if they want to delete those, they just change to using the `delete` subcommand. The difference between a "dry run" and a "listing" is very limited in my mind.
3. It seems unintuitive why `alpha` would default to deleting none but the others would default to 1
4. The current CLI was designed to have syntax as close as possible to English, so users wouldn't have to constantly reference the help menu for flag usage. `hc cache delete oldest 2`, `hc cache delete all` `hc cache delete hipcheck` are all very obvious to replicate once you have done it once. On this point, removing `--invert` and adding `smallest`, `newest`, `ralpha` would be a good takeaway from your proposal to pull into the existing CLI.

That being said, I defer to your judgement on the matter and if you still prefer your proposal, I'm totally okay refactoring the PR.

alilleybrinker:
Okay, I've been thinking about this some more. First, I'll note that two other tools I looked at for prior art around handling deletion operations don't offer any flexibility beyond picking individual ones to delete, or using `--all` to delete all of them.

1. `docker image rm`: https://docs.docker.com/reference/cli/docker/image/rm/
2. `gh cache delete`: https://cli.github.com/manual/gh_cache_delete

The [CLI Guidelines](https://clig.dev/#arguments-and-flags) recommend "confirm[ing] before anything dangerous" and give some examples, including the dry-run/execute pattern I suggested above.

I'll also refer to the Oxide & Friends podcast episode, ["Rebooting a datacenter: A decade later"](https://oxide.computer/podcasts/oxide-and-friends/1935437) which details a story from Joyent about an engineer accidentally rebooting all hosts in an entire datacenter, in part because of a misleading CLI for a destructive operation (rebooting).

---

This is all just a collection of prior art, which still leaves the question of what we should do. As I see it, I don't think the currently-implemented approach of specifying a strategy and a number is quite right. `clap`'s handling of subcommands means that the full documentation for this API is hidden under additional `-h/`--help` invocations on the individual subcommands themselves, which means the help text isn't as informative as it could be.

As an MVP, we could implement something like the Docker and GitHub CLI examples above, permitting specifying individual cache entries to delete, or use an `--all` flag (or perhaps `--strategy all`, for forward compatibility with more strategies) to delete everything. This gives us more time to figure out how to shape this interface.

j-lanson:
Consider:

`list --strategy/-s <STRAT> [--max/-m <N>[ [--filter/-f <PATTERN>]`
where strat: `oldest, newest, largest, smallest, alpha, ralpha`

`delete --strategy/-s <STRAT> --filter/-f <PATTERN> --force`
where strat: `all OR {oldest, alpha, smallest, ...} <N>`

the distinction here is that a sorting policy can be applied to `list` without a quantity `<N>`, but for `delete` it doesn't make sense to provide a sorting policy without a quantity.

`delete -s all`, `delete -s oldest 3`, `delete -s oldest` would all be valid, where the last would default to `oldest 1`. this would require us to do custom `clap` post-processing on the delete strategy but it does away with most of the nonsensical combinations.

we would put cache entry deletion at `MILD` from the CLI guidelines, so default behavior is to prompt `are you sure? y/n` when doing a `delete` operation, which can be skipped when `--force` is added.

alilleybrinker:
> Consider:
> 
> `list --strategy/-s <STRAT> [--max/-m <N>[ [--filter/-f <PATTERN>]` where strat: `oldest, newest, largest, smallest, alpha, ralpha`
> 
> `delete --strategy/-s <STRAT> --filter/-f <PATTERN> --force` where strat: `all OR {oldest, alpha, smallest, ...} <N>`
> 
> the distinction here is that a sorting policy can be applied to `list` without a quantity `<N>`, but for `delete` it doesn't make sense to provide a sorting policy without a quantity.
> 
> `delete -s all`, `delete -s oldest 3`, `delete -s oldest` would all be valid, where the last would default to `oldest 1`. this would require us to do custom `clap` post-processing on the delete strategy but it does away with most of the nonsensical combinations.
> 
> we would put cache entry deletion at `MILD` from the CLI guidelines, so default behavior is to prompt `are you sure? y/n` when doing a `delete` operation, which can be skipped when `--force` is added.

I like this! Too bad that `clap` doesn't support it natively, but I think this seems like a good way to go. Thanks Julian!

alilleybrinker (changes requested):
The error message if I omit the `--strategy` flag on `delete` doesn't make clear that I need to include the flag. Can you update the error message?

alilleybrinker:
Otherwise really like this!

alilleybrinker (approved):

PR #225, opened 2024-07-23
chore: Replace log bridge dependency with bespoke implementation

vcfxb:
`indicatif-log-bridge` was improperly filtering log messages in some instances, and was pretty easy to replace.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -801,7 +801,6 @@ dependencies = [
  "indexmap 2.2.6",
  "indextree",
  "indicatif",
- "indicatif-log-bridge",
  "log",
  "maplit",
  "nom",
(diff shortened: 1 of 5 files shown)


cstepanian:
Taking a look now!

cstepanian:
I'm not so familiar with the details of how log filtering works, so I'm trying to catch up to understand the change.

It looks like [`indicatif-log-bridge`](https://github.com/djugei/indicatif-log-bridge/blob/2e175d9ca2a8ad7108e4911a9a88bc9f47e863f3/src/lib.rs) goes through all the possible [`LevelFilter`](https://docs.rs/log/0.4.22/log/enum.LevelFilter.html) values, starting at the most permissive, and chooses the first one that passes `self.enabled(&meta)`. `enabled()` in turn forwards the call to the underlying wrapped `Log`'s `.enabled()` method.

The new code in `log_bridge.rs` is shorter, instead getting the filter from the wrapped [`env_logger::Logger`](https://docs.rs/env_logger/0.11.3/env_logger/struct.Logger.html).
Hipcheck uses `env_logger` by creating an `Env` and then calling `Builder::from_env()` on it, and finally calling `.build()` on the `Builder`.
I'm assuming that after `.build()` is called, the `HC_LOG` environment variable has been read and the filter(s) from it have been applied to the `Logger`.

I assume there's a good reason for this change, but I'm a bit lost as to what the effect of the change is.
In particular, I can't quite understand what's wrong with how `indicatif-log-bridge` is getting the max filter level.
I'm guessing that it may have to do with the fact that `env_logger` supports more precise filtering by directives that set a log level per target module.

Which log messages weren't being filtered properly, and what would the correct behavior be?

alilleybrinker (approved):

PR #227, opened 2024-07-24
feat: add more robust 'target resolution' system

j-lanson:
The need for the refactor represented by this PR was determined from discussions around tackling #183 .

This PR was a collaborative effort between myself and @mchernicoff .

This PR uses the types defined in `target/types.rs` for representing the "target" of a HipCheck analysis. Although information can be specified on the cmdline by users in a variety of ways (variants of `TargetSeed`), we eventually need to arrive at a local clone of a git repository that may or may not have an associated remote repo / package information. This is all encapsulated by the `Target` struct.

This PR provides glue code for converting `cli.rs` structures to `TargetSeed` instances using the `ToTargetSeed` trait. The `resolve_target()` function does the work of `TargetSeed` to `Result<Target>` translation. 

In order to limit the LOC changed by this PR, the generated `Target` is then translated to the existing `Source` object in `source/query.rs`. In a future PR we will do the work of removing the `Source` ecosystem and fully replacing it with `Target`.

### DIFF
--- a/hipcheck/src/session/pm.rs
+++ b/hipcheck/src/session/pm.rs
@@ -6,6 +6,7 @@ use crate::error::Error;
 use crate::error::Result;
 use crate::hc_error;
 use crate::http::agent;
+use crate::target::{Package, PackageHost};
 use crate::CheckKind;
 use crate::EXIT_FAILURE;
 use serde_json::Value;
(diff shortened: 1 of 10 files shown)


hipcheck/src/session/pm.rs
```diff
@@ -317,13 +316,31 @@ fn extract_package_version_from_url(url: Url) -> Result<(String, String)> {
...
 		//npm gets the first two segments
-		let (package, version) = url
+		// let (package, version) = url
+		// 	.path_segments()
+		// 	.map(|mut i| (i.next(), i.next()))
+		// 	.ok_or_else(|| hc_error!("can't detect package name"))?;
+		// Ok((
+		// 	package.unwrap().to_string(), //this will graceful error if empty because of the mapping above I believe
+		// 	version.unwrap_or("no version").to_string(),
+		// ))
```
alilleybrinker, line 326: Is there a reason to want to keep this old code around? If not we ought to remove it.

alilleybrinker (changes requested):

alilleybrinker:
Just the one small nitpick, otherwise this LGTM! Ping me when it's ready for re-review.

hipcheck/src/session/pm.rs
mchernicoff: No, I think I just forgot to remove it. :) I can take it out and make another squashed commit tonight.

mchernicoff:
@alilleybrinker I deleted the commented out code. Ready for re-review.

alilleybrinker (approved):

PR #236, opened 2024-07-29
feat: shortcut cache entry size calculation with cache index file

j-lanson:
This PR introduces a filesystem backing for the `HcCache` object, stored at `$HC_CACHE/clones/index.json`. When instantiating a `HcCache` object, the constructor now looks for this file, and if it exists, will compare `last_modified` information to shortcut cache entry size calculation if the entry has not been modified since the last time a cache object was instantiated. 

Landing this will help with the potential feature described in #182 , where we may specify a cache size-reduction policy to run every time `hc check` or other subcommands are run. Without this shortcutting, re-calculating the disk size of each cache entry can add significant time overhead, impacting user experience with the tool.

Notably, the way this PR determines index entry invalidation is currently based on if the top-level dir for the repo's modified time has changed. On Linux, modifying the content of a subfolder does not propagate the `modified` information upwards, so if a cache entry `hipcheck` exists, and someone adds a file to `hipcheck/src/X`, the check will not detect that `hipcheck`'s size needs to be recomputed. Storing and checking git `ref` information would help detect one class of changes, but not others.

### DIFF
--- a/hipcheck/src/cache.rs
+++ b/hipcheck/src/cache.rs
@@ -5,12 +5,17 @@ use crate::hc_error;
 use dialoguer::Confirm;
 use pathbuf::pathbuf;
 use regex::Regex;
+use serde::{Deserialize, Serialize};
 use std::borrow::Borrow;
+use std::collections::HashMap;
+use std::fs;
 use std::path::{Path, PathBuf};
 use std::result::Result as StdResult;
 use std::time::SystemTime;
 use tabled::{Table, Tabled};
 use walkdir::{DirEntry, WalkDir};
+
+static CACHE_FILE_NAME: &str = "index.json";
 
 #[derive(Debug, Clone)]
 pub enum CacheSort {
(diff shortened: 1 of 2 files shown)


hipcheck/src/cache.rs
```diff
@@ -288,3 +333,28 @@ impl HcCache {
 		println!("{}", Table::new(to_show));
 	}
 }
+// This ensures the current state of the cache gets written to file. It is
+// important that a "HcCache" instance is dropped before execution terminates
+// so that this function gets a chance to run
```
alilleybrinker, line 338: Hm, Rust doesn't guarantee destructors, so if the program is ended early (for example, a `SIGINT`) this `Drop` impl won't be called. If you _really_ want to make sure it's updated you'll need a signal handler. I don't think that's blocking for this though; could be done later.

alilleybrinker: LogRocket has a decent guide to signal handling in Rust: https://blog.logrocket.com/guide-signal-handling-rust/

alilleybrinker (approved):
One comment, but otherwise LGTM!

Issue #186, opened 2024-07-04
Enable users providing CycloneDX SBOMs as targets


alilleybrinker:
Hipcheck already supports SPDX 2.3 SBOMs as targets, and tries to extract source repository information out of them. At the time this support was added, the existing Rust libraries for ingesting CycloneDX documents were poor quality and in particular had severe limitations around many fields being mistakenly private in the parsed documents with no accessor methods provided either. This meant while we could _parse_ CycloneDX documents, we couldn't _use_ the data in the parsed documents for anything.

This was a couple of years ago, and the state of CycloneDX tooling in Rust has improved substantially. We should revisit this and add support for CycloneDX SBOMs as targets.

lfrancke:
Hi,
I'm one of the maintainers of the [cyclonedx-rust-cargo project](https://github.com/CycloneDX/cyclonedx-rust-cargo) providing a library to read and write CycloneDX SBOMs as well as a CLI tool to generate them for Rust projects.

If you have any questions or anything is missing please feel free to ping us.
We also have a Slack channel in the CycloneDX Slack

alilleybrinker:
Thanks! We’ll definitely reach out if we have issues 😄

PR #237, opened 2024-07-30
feat: Hipcheck can process SBOMs using CycloneDX JSONs

mchernicoff:
Partially resolves Issue #186  Hipcheck can now process SBOMs using both SPDX and CycloneDX (currently JSON only) standard. Currently we only support CycloneDX JSONs in versions 1.3 - 1.5 of the CycloneDX specification.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,6 +20,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "ahash"
+version = "0.8.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
+dependencies = [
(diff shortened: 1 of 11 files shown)


alilleybrinker:
Hm, the test SBOM is pretty big (almost 30,000 lines). Could we find a smaller one?

Also, it looks like tests are currently broken do a test case hard-coding your local path info.

mchernicoff:
Oh whoops, left my home directory hard-coded into the test. Easy fix.

mchernicoff:
> Hm, the test SBOM is pretty big (almost 30,000 lines). Could we find a smaller one?
> 
> Also, it looks like tests are currently broken do a test case hard-coding your local path info.

I can also just chop off most of the SBOM file. Hipcheck only looks at the metadata right now.

alilleybrinker (approved):

Issue #235, opened 2024-07-29
hipcheck does not work with scopes


msbannert:
HipCheck 3.4.0 running on windows 10 /  powershell:

packages which are scoped cannot be analyzed.
Example:
https://registry.npmjs.org/@types/ua-parser-js

PS D:\hipcheck> hc check "@types/ua-parser-js" -t npm
        Analyzing @types/ua-parser
             Done loading configuration and data files
error: no repository given for npm package

alilleybrinker:
Thanks for reporting this! We'll let you know when it's fixed 😄

alilleybrinker:
The fix has been merged, and we're planning to release Hipcheck 3.5.0 this week on Thursday, which will include the fix.

PR #238, opened 2024-07-30
fix: Hipcheck now works with scoped npm packages.

mchernicoff:
Resolves Issue #235. Hipcheck will now include the scope for scoped NPM packages in the format `@scope/package@version`.

This works with NPM targets in `package@version`, NPM registry URL, or pURL formats.


mchernicoff:
Tests that were changed in another PR to main reverted back to the older form when I rebased. Fixing now.

alilleybrinker (approved):

PR #240, opened 2024-07-31
feat: update remote-repo updating to support git refs

j-lanson:
This PR splits `source/git.rs::update()` into `fetch()`, and `checkout()`. It also updates the `TargetSeed::RemoteRepo` variant to pass along `--ref` info from the cmdline to the new `checkout()` function.

The desire for `--ref` exposes some limitations of the current `TargetSeed` enum. Data passed on the CLI through `--ref` can easily be slotted into `LocalGitRepo`, but not into the other variants. One option would be to break away from using `Target` subfields for each variant so the wrapped structs can carry more information. I opted instead to just add another field to the variant, but the decision is very much up for discussion.

Note that this PR does not address passing `--ref` info for other `TargetSeed` variants.

### DIFF
--- a/hipcheck/src/cli.rs
+++ b/hipcheck/src/cli.rs
@@ -620,17 +620,20 @@ pub struct CheckRepoArgs {
 	pub source: String,
 	/// The ref of the repo to analyze
 	#[clap(long = "ref")]
-	pub ref_: Option<String>,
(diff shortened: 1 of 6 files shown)


alilleybrinker (approved):
Looks good to me! Glad to have this working for remote repos too.

PR #264, opened 2024-08-08
CI improvements

alilleybrinker:
This commit features several CI improvements:

- Clearer naming of CI files, pipelines, and jobs.
- Adds "website-test.yml" to test website changes on PRs.
- Removes duplicative tests on `main` after equivalent tests on PRs, given that PRs can only merge with passing tests.

### DIFF
--- a/.github/workflows/commits.yml
+++ b/.github/workflows/commits.yml
@@ -1,21 +1,15 @@
-name: Conventional Commits
+name: Commits
 
 on:
-  push:
-    branches: [main]
   pull_request:
     branches: [main]
 
 permissions:
   contents: read
 
-env:
-  RUSTFLAGS: -Dwarnings
-  CARGO_TERM_COLOR: always
-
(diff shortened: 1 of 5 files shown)


mchernicoff (approved):

cstepanian:
I think there is a caching issue caused by no longer running the tests on pushes to `main`.
It may seem duplicative to also build on `main`, but the end result will be that every Pull Request build will start without a cache. (subsequent pushes to the PR will be able to use the cache created for that PR)

GitHub Actions isolates caches created for Pull Requests, so they cannot be used when building other Pull Requests (except when the new PR is based on another PR that had a cache) or the `main` branch.
Builds on `main` are the only way for Actions to create caches that can be used by Pull Requests on their first build.

alilleybrinker:
Oh, how annoying. Okay, want to open a PR to run on main again?

Issue #171, opened 2024-07-01
Generate SBOMs for Hipcheck Distribution Artifacts


alilleybrinker:
Hipcheck today effectively produces three artifacts with each release, each of which should have an SBOM:

- [ ] `hc` binary
- [ ] `hc-update` binary (produced by `cargo-dist`)
- [ ] Hipcheck Docker image published to Docker Hub

Of these, the binaries are probably _easiest_ to produce an SBOM for, but it's the Docker container SBOM we probably care about the most.

This will also involve deciding if we want to produce CycloneDX and/or SPDX SBOMs.

mchernicoff:
We will use SPDX as the intended SBOM standard (for now), in the interest of choosing a standard.

PR #269, opened 2024-08-09
chore: Create docker image SBOM on release

mchernicoff:
Resolves first part of Issue #171 (other parts are waiting on changes to `cargo-dist`). Adds a GitHub action to create an SPDX SBOM for the Hipcheck docker image after we create and push that image during the normal release workflow. The SBOM _should_ be automatically added as an artifact to the same release.

Draft because this has not been tested yet. We need to come up with a good way to test this that does not involve creating many unnecessary docker images.

### DIFF
--- a/.github/workflows/dockerhub.yml
+++ b/.github/workflows/dockerhub.yml
@@ -63,3 +63,14 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
           readme-filepath: ./dist/dockerhub/README.md
+  
+  build-sbom:
+    needs: push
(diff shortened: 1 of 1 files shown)


alilleybrinker (approved):

alilleybrinker:
@mchernicoff looks like this has a merge conflict with the changes @patrickjcasey made for checking Docker README size. Can you rebase?

alilleybrinker:
Going to close this. I'm in the process of some more substantial reworking of how releases work, and while I definitely want SBOM building to be a part of that, I don't think it'll necessarily take this form. I'll be sure to refer back to this though, and @mchernicoff I appreciate the work you did on it!

PR #277, opened 2024-08-12
feat: add proto-file based code gen

j-lanson:
Initial commit to inject `hipcheck.proto` --> `hipcheck.rs` code gen into our build process. Updated site install instructions, GitHub CI, and Containerfile to include `protoc` install. 

Windows `protoc` install for GitHub CI is currently reliant on curling specific version, which will need to be updated over time. Assuming `chocolatey` is installed, could use [this package](https://community.chocolatey.org/packages/protoc) instead. Site shows a few warnings about binaries included, so I was hesitant.

### DIFF
--- a/hipcheck/Cargo.toml
+++ b/hipcheck/Cargo.toml
@@ -86,6 +86,8 @@ http = "1.1.0"
 dialoguer = "0.11.0"
 tabled = "0.15.0"
 fs_extra = "1.3.0"
+tonic = "0.12.1"
+prost = "0.13.1"
 
 # Exactly matching the version of rustls used by ureq
 # Get rid of default features since we don't use the AWS backed crypto provider (we use ring).
@@ -102,6 +104,7 @@ features = [
 
 [build-dependencies]
 anyhow = "1.0.86"
+tonic-build = "0.12.1"
 which = { version = "6.0.1", default-features = false }
 
 [dev-dependencies]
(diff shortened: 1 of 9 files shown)


hipcheck/Cargo.toml
```diff
@@ -86,6 +86,9 @@ http = "1.1.0"
 dialoguer = "0.11.0"
 tabled = "0.15.0"
 fs_extra = "1.3.0"
+tonic = "0.12.1"
+prost = "0.13.1"
+tonic-build = { version = "0.12.1", features = ["prost"] }
```
alilleybrinker, line 91: Does `tonic` not re-export `prost` or provide a feature for building? Seems surprising to need to import all 3.

alilleybrinker: Oh also is `tonic-build` a requirement at runtime, or just build time? I see it's also a build dependency.

alilleybrinker (changes requested):
Small question around imports, otherwise LGTM!

hipcheck/Cargo.toml
j-lanson: removing `prost` from the Cargo.toml causes compilation to fail. I don't see anything in the list of disabled `tonic` features that indicates such a feature, they're all compression- or TLS-related.

j-lanson: good catch on the `tonic-build` in `dependencies`, will remove

alilleybrinker: Interesting! Looks like you're right, I'm not seeing anything that re-exports `prost`, so I think just removing `tonic-build` is all we can do.

alilleybrinker (approved):

PR #278, opened 2024-08-12
feat: initial plugin startup and context management

j-lanson:
**Note:** this PR relies on #277 , and so that must be merged first.

This PR introduces the `plugin` module and some initial type definitions, as well as some code for turning a `Plugin` into a `PluginContext` object that will act as a wrapper interface for the gRPC channel (converting between `tonic`-generated types and "nicer" ones, maintaining the child process handle, etc.)`

The logic for how startup occurs is a first try, and should be subject to discussion. Rather than a `MAX_CONN_ATTEMPTS` to handle syncronizing the plugin process and `Hipcheck`, I would support adding a requirement that the plugin outputs to `stdout` some notifying message like "Ready." that we can wait for.

The code of this PR compiles, but is untested. It would require defining an actual plugin, and injecting a way to test this code into `main.rs`. We may want to consider switching to a library-style crate so that we can have multiple different `bin/` files to enable easier testing of code such as this.


hipcheck/src/plugin/manager.rs
```diff
@@ -0,0 +1,79 @@
...
+const MAX_CONN_ATTEMPTS: usize = 4;
+
+pub struct PluginExecutor {}
+impl PluginExecutor {
+	pub fn new() -> Self {
+		PluginExecutor {}
+	}
+	fn get_available_port(&mut self) -> Result<u16> {
+		for i in 40000..u16::MAX {
+			if let Ok(bind) = std::net::TcpListener::bind(format!("127.0.0.1:{i}")) {
```
alilleybrinker, line 16: Just occurred to me that there's a TOCTOU issue with our port-assignment protocol. The RFD specifies that `hc` controls port assignment and tells the plugins what port to bind to, but that means that a port may be claimed in the time between Hipcheck checking the port's availability and a plugin binding to it.

The alternative would be to have plugins chose their own port and then tell Hipcheck about the port through an init message (perhaps this could also tie into your "ready message" idea).

hipcheck/src/plugin/mod.rs
```diff
@@ -0,0 +1,13 @@
...
+
+use crate::plugin::manager::*;
+pub use crate::plugin::types::*;
+
+pub fn dummy() {
+	let plugin = Plugin {
+		name: "dummy".to_owned(),
+		entrypoint: "./dummy".to_owned(),
+	};
+	let manager = PluginExecutor::new(3, 5, 40000..u16::MAX, 1000, 10);
```
cstepanian, line 12: This feels like a lot of magic numbers in a row. It would be nice to have their meanings alongside the values.
I would normally suggest using the syntax `PluginExecutor { max_spawn_attempts: 3, ...}`, but you can't directly instantiate the struct.

For a more user-facing struct, the Builder pattern would probably be a good choice. I'm not sure if it's worth it here for `PluginExecutor`, especially since it probably won't be constructed many places in the codebase.

Perhaps just some comments would be the best choice here.

hipcheck/src/plugin/manager.rs
```diff
@@ -0,0 +1,121 @@
...
+	pub async fn start_plugin(&mut self, plugin: &Plugin) -> Result<PluginContext> {
+		let mut rng = rand::thread_rng();
+		// Plugin startup design has inherent TOCTOU flaws since we tell the plugin
+		// which port we expect it to bind to. We can try to ensure the port we pass
+		// on the cmdline is not already in use, but it is still possible for that
+		// port to become unavailable between our check and the plugin's bind attempt.
+		// Hence the need for subsequent attempts if we get unlucky
+		let mut spawn_attempts: usize = 0;
+		while spawn_attempts < self.max_spawn_attempts {
+			// Find free port for process. Don't use spawn_attempts, since
```
cstepanian, line 58: This comment seems incomplete or it was cut off in editing -- was there more to say about why not to use `spawn_attempts` here?

hipcheck/src/plugin/types.rs
```diff
@@ -0,0 +1,149 @@
...
+	fn try_from(value: ConfigurationStatus) -> Result<Self> {
+		use ConfigErrorType::*;
+		use ConfigurationStatus::*;
+		Ok(match value as i32 {
+			x if x == ErrorUnknown as i32 => Unknown,
+			x if x == ErrorMissingRequiredConfiguration as i32 => MissingRequiredConfig,
+			x if x == ErrorUnrecognizedConfiguration as i32 => UnrecognizedConfig,
+			x if x == ErrorInvalidConfigurationValue as i32 => InvalidConfigValue,
+			_ => {
+				return Err(hc_error!("status value is not an error"));
```
cstepanian, line 83: I think it would be good to include the invalid value in the error output, for more context if we need to debug this.

hipcheck/src/plugin/manager.rs
j-lanson: Fixed.

hipcheck/src/plugin/types.rs
j-lanson: Good idea, updated with this.

hipcheck/src/plugin/mod.rs
alilleybrinker: That's right, my intent is that these items become configurable in the future.

hipcheck/src/plugin/manager.rs
```diff
@@ -0,0 +1,122 @@
...
+use std::process::Command;
+use tokio::time::{sleep_until, Duration, Instant};
+
+pub struct PluginExecutor {
+	max_spawn_attempts: usize,
+	max_conn_attempts: usize,
+	port_range: Range<u16>,
+	backoff_interval: Duration,
+	jitter_pct: u8,
+	est_ports: HashSet<u16>,
```
alilleybrinker, line 16: Small preference thing: I'd prefer not to abbreviate to "est" or at least to spell out in a comment what it's a shorthand for.

j-lanson: Got it. I can update it if you want, but I'll also say that I have already removed this field in the next PR i intend to merge. For the purposes of using `join_all()` to initialize all plugins concurrently in async fashion, the Executor can't have mutable state

alilleybrinker: Okay cool, if it's getting removed that's fine. As far as mutexes, Tokio does have an async mutex in it.

alilleybrinker: Also, we may want to look into: https://marabos.nl/atomics/

cstepanian: On the topic of abbreviations, I wasn't sure about `pct` for "percent". I only became sure that it stood for "percent" after finding how the value was used, and in particular that it was divided by 100 at one point.

But maybe `pct` is a well established abbreviation that I just haven't seen much before.

j-lanson: Fair, I'll update that one.

alilleybrinker (approved):

PR #282, opened 2024-08-16
feat: added parsing logic for download manifests

patrickjcasey:
Implemented the logic for parsing download manifest files in rfd4:
https://github.com/mitre/hipcheck/blob/main/site/content/rfds/0004-plugin-api.md


j-lanson (approved):
Looks great to me, I like the trait design.

PR #293, opened 2024-08-20
qfeat: added parsing logic for plugin manifests

patrickjcasey:


hipcheck/src/plugin/parser/plugin_manifest.rs
```diff
@@ -0,0 +1,418 @@
...
+			version,
+			manifest,
+		})
+	}
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PluginDependencyList(Vec<PluginDependency>);
+
+impl PluginDependencyList {
```
j-lanson, line 132: Is there a reason not to implement this as `type PluginDependencyList = Vec<PluginDependency>;`? AFAIK you can `impl` a trait we define on a type we don't define.

hipcheck/src/plugin/parser/download_manifest.rs
```diff
@@ -1,55 +1,16 @@
...
-	/// Return the name of the attribute used to identify the node pertaining to this struct
-	fn kdl_key() -> &'static str;
-
-	/// Attempt to convert a `kdl::KdlNode` into Self
-	fn parse_node(node: &KdlNode) -> Option<Self>;
-}
-
-#[allow(unused)]
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Url(pub String);
```
j-lanson, line 11: Maybe we should take it a step further than a `String` and wrap a `url::Url` type?

patrickjcasey: I was not sure if it was always going to be a `Url`, if it is then we can definitely swap that from `String` to `url::Url`

hipcheck/src/plugin/parser/plugin_manifest.rs
patrickjcasey: I was thinking that it would be easier to add custom methods for dealing with all of the dependencies

j-lanson: I'm personally not a fan of the shallow-wrapper design, especially because it makes it so we have to re-define any `Vec` function we want to use on a `PluginDependencyList`. I'd advocate for just changing the internal `Vec` to `pub` so it can be accessed directly if need be.

hipcheck/src/plugin/parser/download_manifest.rs
j-lanson: Based on the RDF I think its safe to assume it will always be `Url`

hipcheck/src/plugin/parser/plugin_manifest.rs
patrickjcasey: I exposed the wrapped Vec with a `pub`. I like not needing to redefine all of the functionality of `Vec` by just using `pub`

j-lanson (approved):

PR #294, opened 2024-08-22
feat (cli): Adds new --policy flag to Hipcheck

mchernicoff:
Resolves Issue #292 .Adds new `-p`/`--policy` flag to pass a policy file to Hipcheck. Defaults to a file named `Hipcheck.KDL` in the current working directory if the flag is omitted.

`hc ready` now checks for a policy file. If one is missing, a custom error message reminds the user to provide one manually.

Currently, instead of erroring if *no* policy file is provided, Hipcheck errors *if* a file is provided, as policy files are not currently supported by Hipcheck. This will change in a future PR once downstream parsing of policy files has been implemented.

### DIFF
--- a/hipcheck/src/cli.rs
+++ b/hipcheck/src/cli.rs
@@ -17,7 +17,7 @@ use crate::target::{
 use clap::{Parser as _, ValueEnum};
 use hipcheck_macros as hc;
 use pathbuf::pathbuf;
-use std::path::{Path, PathBuf};
+use std::{env, path::{Path, PathBuf}};
 use url::Url;
 
 /// Automatated supply chain risk assessment of software packages.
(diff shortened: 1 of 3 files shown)


hipcheck/src/cli.rs
```diff
@@ -1181,6 +1201,29 @@ mod tests {
 		});
 	}
 
+	#[test]
+	fn resolve_policy_with_flag() {
+		let tempdir = TempDir::with_prefix(TEMPDIR_PREFIX).unwrap();
+
+		let expected = pathbuf![tempdir.path(), "HipcheckPolicy.KDL"];
```
alilleybrinker, line 1208: Nit: `KDL` ought to be lowercase

mchernicoff: I knew I would get that backwards somewhere. Fixed.

alilleybrinker (approved):

PR #290, opened 2024-08-20
Add dummy plugin and CLI to execute for testing

j-lanson:
This PR introduces a dummy plugin `rand_data` that takes an int N and returns a vec of random bytes with length N. It also adds a CLI subcommand `hc plugin` that will do the work of starting, intializing, and querying the plugin for an array of bytes, then closing. This is a temporary and hidden subcommand to help us with testing the plugin query-protocol code we write.

PR #297, opened 2024-08-23
feat: plugin interface handles multiple active query sessions

j-lanson:
*Note* This PR relies on #290 . Merge this only after it has been merged.

This PR further wraps `PluginTransport` in an `ActivePlugin` struct, which manages query "sessions". It is responsible for handing out query IDs to new queries (`next_id` is guarded by a Mutex), and converts `Query --> PluginResponse` object which either returns the result or packages the info needed to "resume" the current query once the required information has been gathered.

This also implements multiplexing on the `PuginTransport::rx` receiver, along with a backlog of messages. This way a query session can ignore but save messages not destined for them.

Overall, this PR also accomplishes removing `&mut` references from the plugin communication interface so it can be `Send` and thus eligible for salsa async database functionality.

### DIFF
--- a/hipcheck/src/plugin/mod.rs
+++ b/hipcheck/src/plugin/mod.rs
@@ -8,7 +8,7 @@ use crate::Result;
 use futures::future::join_all;
 use serde_json::Value;
 use std::collections::HashMap;
-use tokio::sync::mpsc;
+use tokio::sync::{mpsc, Mutex};
 
 pub fn dummy() {
 	let plugin = Plugin {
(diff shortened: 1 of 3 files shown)


hipcheck/src/plugin/mod.rs
```diff
@@ -41,9 +41,58 @@ pub async fn initialize_plugins(
...
+	pub fn new(channel: PluginTransport) -> Self {
+		ActivePlugin {
+			next_id: Mutex::new(0),
+			channel,
+		}
+	}
+	async fn get_unique_id(&self) -> usize {
+		let mut id_lock = self.next_id.lock().await;
+		let res: usize = *id_lock;
+		*id_lock += 2;
```
alilleybrinker, line 58: Probably just drop a comment here about IDs from `hc` always being even.

hipcheck/src/plugin/types.rs
```diff
@@ -185,8 +186,9 @@ impl PluginContext {
 		);
 		self.set_configuration(&config).await?.as_result()?;
 		let default_policy_expr = self.get_default_policy_expression().await?;
-		let (tx, mut out_rx) = tokio::sync::mpsc::channel::<PluginQuery>(10);
+		let (tx, mut out_rx) = mpsc::channel::<PluginQuery>(10);
```
alilleybrinker, line 189: In the future we'll want this queue size to be configurable.

hipcheck/src/plugin/types.rs
```diff
@@ -271,46 +273,103 @@ impl TryFrom<Query> for PluginQuery {
...
+		if let Some(msgs) = self.backlog.remove(&id) {
+			return Ok(Some(msgs));
+		}
+		// No backlog message, need to operate the receiver
+		loop {
+			let Some(raw) = self.rx.message().await? else {
+				// gRPC channel was closed
+				return Ok(None);
+			};
+			let raw_id = raw.id;
```
alilleybrinker, line 300: Why do an assignment here instead of using `raw.id` later?

alilleybrinker (changes requested):
Just minor nits. Otherwise LGTM!

hipcheck/src/plugin/types.rs
j-lanson: Changed it. It was just because the `insert()` was going to require a `clone()` anyway.

hipcheck/src/plugin/types.rs
j-lanson: I'll make an issue

alilleybrinker (approved):

PR #299, opened 2024-08-23
fix quiet bug regression

patrickjcasey:
When running `hc check --format json -v quiet https://github.com/mitre/hipcheck`, the following error occurs on `origin/main`

image

Here is the tip of `origin/main` with the policy file error omitted

image

The tip of this branch has nothing printed before the JSON is output

image

If anyone has an idea of how to test this/ensure I hid everything that should be hidden when `-v quiet` is passed, I would love to hear it!


alilleybrinker:
Regarding testing, we've talked about doing snapshot testing with `cargo-insta` and that's what I'd like to do. Can you open an issue for it in the next milestone?

PR #308, opened 2024-08-28
feat: Update proto def to pass 'buf lint'

alilleybrinker:
This commit modifies the protobuf definition for Hipcheck plugins to meet the expectations of 'buf lint', which checks protobuf definitions against a set of best practices. The main goal here is to make the service definition more defensive against forward-compatibility issues.

This also includes updates to the code that interacts with the generated code to make sure it still compiles.


j-lanson:
I imagine this will break the working communication with the dummy plugin via `hc plugin`

alilleybrinker:
Ah, you're right that I'll need to update that as well. I'll amend the patch with that.

alilleybrinker:
Julian has some in-flight changes to the demo plugin with stuff that we'll be pulling out into our Rust plugin SDK, and I'd rather that get merged first and then rebase this on top. So this'll wait until that's ready.

alilleybrinker:
Resolved merge conflicts, now to get it to compile again haha

hipcheck/src/plugin/types.rs
```diff
@@ -275,34 +331,66 @@ impl TryFrom<Query> for PluginQuery {
...
+/// `query` field on the message can technically be missing.
+///
+/// This case is handled in the `message` method to flatten
+/// that kind of error so code consuming the stream doesn't
+/// have to worry about it.
+type QueryStream = Box<dyn Stream<Item = StdResult<PluginQuery, Status>> + Send + Unpin + 'static>;
+
 impl MultiplexedQueryReceiver {
-	pub fn new(rx: Streaming<PluginQuery>) -> Self {
+	pub fn new(rx: QueryStream) -> Self {
```
j-lanson, line 359: Was it changes from `buf lint` that caused the need for this to become `Box<dyn Stream>` or was it a deficiency in my existing code?

alilleybrinker: Good catch! This was necessary for compilation at one point during the refactoring, but I just checked and I was able to remove it without issue in the final version.

alilleybrinker:
@j-lanson removed boxed query stream. Ready for re-review

j-lanson (approved):
Looks good!

PR #309, opened 2024-08-28
DRAFT: fix: dummy plugin refactored to use "Session" tracking

j-lanson:
Previous implementation of receiver multiplexing for plugins had inherent bug that manifested when `dummy_rand_data` handler tried to request data from another plugin. It's `recv()` functionality would check if anything was in the backlog, else lock and listen on the underlying receiver object. The new-query accepting task already captures the message that the handler is waiting for, and puts in the backlog. Thus, the handler will hang waiting for a new message that will not necessarily come while the thing it wants is already in the backlog.

In order to overcome this, I refactored the `MultiplexedQueryReceiver` to create dependent `QuerySession` objects. The backlog was transitioned to a session-tracking hashmap. `QuerySession` objects implement `Drop` and use this to remove themselves from the session tracker when they go out of scope.

This PR also modifies `dummy_rand_data` to do something different (take a number, modulo by N, request the hash of the remainder, then add the original number to the hash vec), so as to require requesting data from another plugin (newly-added `dummy_sha256`). The use of the modulo operator is to allow us to test memoization, as only N distinct queries will be made to `dummy_sha256` which guarantees collisions and opportunity for memo-ization).

This PR also undoes the "bind to port to see if its available" trick for assigning plugin ports, as there is a hidden race condition if plugins are started asynchronously (two `bind()` tests to the same port will return true because the first plugin hasn't started yet). Deferred to asking the OS what port to use.


hipcheck/src/engine.rs
```diff
@@ -37,6 +37,7 @@ fn query(
 	};
 	// Initiate the query. If remote closed or we got our response immediately,
 	// return
+	println!("Querying {plugin}::{query} with key {key:?}");
```
alilleybrinker, line 40: Whenever this code stops being experimental, these will need to become logging calls.

plugins/dummy_rand_data/src/hipcheck_transport.rs
```diff
@@ -141,86 +161,86 @@ impl HcTransport {
...
+		while res_s_handle.is_err() {
+			res_s_handle = self.sessions.try_lock();
+		}
+		let mut s_handle = res_s_handle.unwrap();
+		eprintln!("Got lock for drop finally");
+		let Some(_channel) = s_handle.remove(&raw_id) else {
+			panic!("Dropping session that does not exist in sessions");
+		};
+	}
+}
```
alilleybrinker, line 189: Another option here would be to have a message you can send to the session tracker task telling it that it can drop the session, and then not having the session itself wait for that cleanup, and completing drop here.

plugins/dummy_sha256/src/hipcheck_transport.rs
```diff
@@ -0,0 +1,226 @@
...
+	rx: Arc<Mutex<MultiplexedQueryReceiver>>,
+}
+impl HcTransport {
+	pub fn new(rx: Streaming<PluginQuery>, tx: mpsc::Sender<Result<PluginQuery, Status>>) -> Self {
+		HcTransport {
+			rx: Arc::new(Mutex::new(MultiplexedQueryReceiver::new(rx))),
+			tx,
+		}
+	}
+	pub async fn send(&self, query: Query) -> Result<()> {
```
alilleybrinker, line 81: This is just a style thing, but I think in general having empty newlines between functions and in many other places would 1) match the style of the rest of the Hipcheck codebase and 2) be easier to read.

alilleybrinker (changes requested):
My only big comment is in the `Drop` impl for an alternative we ought to consider.

plugins/dummy_sha256/src/hipcheck_transport.rs
j-lanson: I can see that, but unless we make HcSessionSocket itself spawn a task to watch this "drop" message, the best we can do is have HcSessionSocket look for any such messages after `let Some(raw) = rx_handle.message().await?` in `listen()`. Which do you prefer? I personally don't stylistically like the idea of spawning a task inside HcSessionSocket, but up to you

j-lanson: I've pushed an example of how it would work without spawning an internal listener task. Let me know what you think.

j-lanson: Whoops, I realized I've been replying to the wrong comment. Sorry about that. I'm referring to the above session tracker discussion.

alilleybrinker: Thanks, this does seem cleaner to me.

alilleybrinker (approved):
LGTM! Just needs a squash and to fix the commit message to pass Conventional Commits checks.

PR #310, opened 2024-08-28
chore: Use consistent module folder structure.

alilleybrinker:
Over time we'd grown a mix of two styles of defining modules with their own children. Some were done with a `<module_name>.rs` file at the top of the `src` folder and a `<module_name>/` folder as well, others used the `<module_name>/mod.rs` pattern instead.

Over time I've come to prefer the `mod.rs` style, and this commit moves the `hipcheck` crate over to it. Mostly this is about consistency, making it easy to find things we need, more than anything else.


cstepanian (approved):
Looks good to me! No issues compiling on my machine.

PR #315, opened 2024-08-29
feat: Add JSON Pointer preprocessor for Policy Expression language

### LINKED ISSUE #261: Implement JSON-pointer like injection syntax for policy expressions

cstepanian:
Resolves #261

This PR adds a JSON Pointer preprocessor step before parsing and evaluation of Policy Expressions.
It adds a `context` parameter to the execution of Policy Expressions, which is used to lookup values referred to by the JSON Pointers.

It allows a limited subset of the full JSON Pointer syntax: only alphanumeric, '/' (slash), '~' (tilde), and '_' (underscore) characters are allowed.

This PR also adds a `MultipleErrors` variant of `policy_exprs::error::Error`. It is probably not the best long term design to solve the problem of tracking a list of errors found during parsing, checking, and execution of a language.

### DIFF
--- a/hipcheck/src/policy_exprs/error.rs
+++ b/hipcheck/src/policy_exprs/error.rs
@@ -6,8 +6,11 @@ use ordered_float::FloatIsNan;
 pub type Result<T> = std::result::Result<T, Error>;
 
 /// An error arising during program execution.
-#[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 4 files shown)


alilleybrinker (approved):

alilleybrinker:
Thanks @cstepanian!

PR #316, opened 2024-08-30
feat: added primitive support for plugin downloading, decompression a…

patrickjcasey:
This pull request adds the following:

- Preliminary support for downloading, decompressing and unarchiving plugins into the `HC_CACHE/plugins` folder
- Renamed several structs related to plugins to include `Plugin` as a prefix to avoid confusion
- Removed `hipcheck/plugin/parser`, as it started to contain code that was outside of the scope of just `KDL` parsing


mchernicoff (approved):
Looks good to me from the plugin structure side, but I'll wait to merge on @alilleybrinker  taking a look too.

alilleybrinker (approved):

PR #324, opened 2024-09-03
chore: now store architecture as an enum

patrickjcasey:
Updated `arch` fields in the download and plugin manifest structs to hold an enum of the architectures supported by hipcheck

### DIFF
--- a/hipcheck/src/plugin/supported_arch.rs
+++ b/hipcheck/src/plugin/supported_arch.rs
@@ -1,0 +1,44 @@
+use std::{fmt::Display, str::FromStr};
+
+use crate::hc_error;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+/// Officially supported target triples, as of RFD #0004
+///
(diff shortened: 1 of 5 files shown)


hipcheck/src/plugin/supported_arch.rs
```diff
@@ -0,0 +1,44 @@
+use std::{fmt::Display, str::FromStr};
+
+use crate::hc_error;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+/// Officially supported target triples, as of RFD #0004
+///
+/// NOTE: these architectures correspond to the offically supported Rust platforms
+pub enum SupportedArchitecture {
```
alilleybrinker, line 9: My only thought is this could probably be shortened to `SupportedArch`. Otherwise LGTM!

patrickjcasey: Updated it!

alilleybrinker (approved):

PR #327, opened 2024-09-04
feat: introduce plugin-based query infrastructure to scoring

j-lanson:
A roughly implemented PR that changes `score_results()` to use the plugin query infrastructure to drive analysis. Does not yet make use of `PolicyExpr` for scoring purposes. `AnalysisTree` ought to be the main object for driving execution/scoring going forward.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,9 @@
 # dotenv env configuration file
 .env
 
+# Vim swap files
+.*.swp
+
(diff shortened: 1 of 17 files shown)


alilleybrinker (approved):

alilleybrinker:
LGTM! Especially considering this is a stopgap while we transition fully to plugins. Thanks @j-lanson!

PR #331, opened 2024-09-04
fix: Copy plugins into containerfile build step

alilleybrinker:
Builds for Docker images to publish to Docker Hub were failing because we weren't copying in all the crates found in the top-level workspace `Cargo.toml` file, causing Cargo to complain. This fixes that by copying them into the builder, though they don't get built.


j-lanson (approved):
Looks good.

PR #333, opened 2024-09-04
fix: Adds weight field to policy file categories

mchernicoff:
Adds the missing `weight` field to the `PolicyCategory` struct. This field represents optional weights that can be given to categories of analyses in the policy file and was accidentally omitted in the PR that added that struct.


j-lanson (approved):
Looks good

PR #334, opened 2024-09-05
feat: implement PolicyFile --> AnalysisTree conversion, remove use of WeightTree in scoring

j-lanson:
Replaced use of WeightTree in scoring with AnalysisTree to help phase out that part of the infrastructure. Added `policy()` salsa function, currently returns an `Option` until we introduce the `ConfigFile-->PolicyFile` conversion so we always have a working policy file.

### DIFF
--- a/hipcheck/src/config.rs
+++ b/hipcheck/src/config.rs
@@ -4,9 +4,11 @@
 
 use crate::analysis::score::*;
 use crate::context::Context;
+use crate::engine::HcEngine;
 use crate::error::Result;
 use crate::hc_error;
-use crate::policy_exprs::{Expr, Primitive};
+use crate::policy::policy_file::{PolicyAnalysis, PolicyCategory, PolicyCategoryChild};
+use crate::policy::PolicyFile;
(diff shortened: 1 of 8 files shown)


mchernicoff (approved):

hipcheck/src/config.rs
```diff
@@ -806,7 +833,7 @@ impl WeightTreeNode {
 	// @Temporary - until policy file impl'd and integrated, we hard-code
 	// the policy for our analyses
 	pub fn with_hardcoded_expr(&self) -> AnalysisTreeNode {
-		let expr = Expr::Primitive(Primitive::Bool(false));
+		let expr = "true".to_owned();
```
cstepanian, line 836: Note that `true` is not valid Policy Expression syntax; `#t`/`#f` is the syntax for a literal boolean.
I'm not sure how temporary this code is expected to be; maybe this will never be evaluated.

cstepanian: I've submitted a PR (#338) to fix this.

PR #336, opened 2024-09-05
chore: Converts config file to policy file under the hood

mchernicoff:
Resolves issue #335 

If no KDL policy file is given by the user, but a deprecated TOML config file is found, Hipcheck will convert the config file to a policy file structure before running.

Note that this forces certain default choices on Hipcheck, because the older config file format does not contain all the same information a policy file.

### DIFF
--- a/hipcheck/src/config.rs
+++ b/hipcheck/src/config.rs
@@ -325,7 +325,7 @@ pub struct IdentityConfig {
 	#[default = 1]
 	pub weight: u64,
 
-	/// A percentage of commits permitted to have a mismatch between committer and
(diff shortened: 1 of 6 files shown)


hipcheck/src/config.rs
```diff
@@ -973,19 +977,15 @@ fn add_category(
 
 pub fn analysis_tree(db: &dyn WeightTreeProvider) -> Result<Rc<AnalysisTree>> {
 	// @Todo - once ConfigFile-->PolicyFile implemented, deprecate else block
```
j-lanson, line 979: Remove associated Todo comment

j-lanson (approved):

PR #340, opened 2024-09-06
fix: Adds code to retrieve config information from policy file analyses.

mchernicoff:
Adds code that allows config file locations provided in a policy file to be read into Salsa in `config.rs`


mchernicoff:
Code still has dead code warnings. Someone will need to clean those up.

alilleybrinker (approved):

mchernicoff:
This is a very ugly implementation of this feature. With more time, someone (probably me) should write some helper functions in the `policy` module to avoid all these nasty nested `if` and `match` blocks.

PR #341, opened 2024-09-06
Jlanson/add concern

j-lanson:
add `concern` field to proto definition. verified with `hc plugin` that the results make it back in one piece.

@alilleybrinker does this require a new version number for the `.proto` file?


alilleybrinker:
No new version number needed for the proto file, as it is backwards compatible. Version numbers in gRPC/protobuf are only for when you break compatibility, and adding new fields is non-breaking

PR #342, opened 2024-09-06
feat: Added PolicyFile::get_config method

alilleybrinker:
This lets us get configuration for a specific analysis by name, with a little bit of recursive searching.

### DIFF
--- a/hipcheck/src/config.rs
+++ b/hipcheck/src/config.rs
@@ -852,32 +852,33 @@ fn langs_file_rel(_db: &dyn LanguagesConfigQuery) -> Rc<String> {
 
 fn langs_file(db: &dyn LanguagesConfigQuery) -> Result<Rc<PathBuf>> {
 	if let Some(config_dir) = db.config_dir() {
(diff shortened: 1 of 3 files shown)


hipcheck/src/config.rs
```diff
@@ -852,32 +852,33 @@ fn langs_file_rel(_db: &dyn LanguagesConfigQuery) -> Rc<String> {
...
+	let policy_file = db.policy();
+	for category in &policy_file.as_ref().analyze.categories {
+		if category.name.eq("languages") {
+			for child in &category.children {
+				match child {
+					PolicyCategoryChild::Analysis(analysis) => {
+						if analysis.name.name == "linguist" {
+							if let Some(config) = &analysis.config {
+								if let Some(filepath) = config.clone().get("langs-file") {
+									return Ok(Rc::new(Path::new(&filepath).to_path_buf()))
```
j-lanson, line 870: isn't the point to use the `get_config()` function?

j-lanson: This can be addressed in a later commit

PR #344, opened 2024-09-06
chore: update legacy config salsa funcs to use policy file get_config()

j-lanson:
Change the legacy salsa functions that provide analysis-specific configuration to extract from policy file.


alilleybrinker:
Ah, you beat me to it for this bit while I was away for lunch. Thanks Julian!

PR #345, opened 2024-09-06
feat: plugin downloading and extraction added to `hc check`

patrickjcasey:
This pull request does the following:
- adds example `Hipcheck.kdl` to `config/`
- download and extract specified plugins (excluding legacy mitre/* plugins)
- updated RFD 4 to indicate `plugin.kdl` is the expected name of the plugin manifest for a plugin
- added SUPPORTED_ARCH constant that can be used to determine what rust triple version is currently running, used to determine which version of plugins to download and which entrypoints to use

### DIFF
--- a/config/Hipcheck.kdl
+++ b/config/Hipcheck.kdl
@@ -1,0 +1,43 @@
+plugins {
+    plugin "mitre/activity" version="0.1.0"
+    plugin "mitre/binary" version="0.1.0"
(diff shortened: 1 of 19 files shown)


alilleybrinker:
Thanks Patrick! I’ll review this first thing tomorrow

j-lanson:
Updated this branch with fixes to make it so that we can finish an entirely plugin-file based hipcheck execution

PR #349, opened 2024-09-07
feat: Implement big chunk of the plugin system.

alilleybrinker:


alilleybrinker:
Merging without review here since this is all stuff we'd previously reviewed when merging into this branch.

PR #351, opened 2024-09-07
feat: Additional impls toward plugin system

alilleybrinker:


hipcheck/src/analysis/score.rs
```diff
@@ -377,14 +396,14 @@ fn wrapped_query(
 			FUZZ_PHASE => db.fuzz_analysis(),
 			REVIEW_PHASE => db.review_analysis(),
 			TYPO_PHASE => db.typo_analysis(),
-			other => Err(hc_error!("Unrecognized legacy analysis '{other}'")),
+			_ => Err(hc_error!("Unrecognized legacy analysis '{other}'")),
```
j-lanson, line 397: isn't `other` still used in the error message?

alilleybrinker: Oh strange, it appears to compile just fine, and without the `_` it gave a warning. Strange.

j-lanson (approved):

PR #352, opened 2024-09-07
feat: update report to use investigate policy expression

j-lanson:
Update salsa functions and ReportBuilder to determine overall pass/fail based on policy expression


alilleybrinker (approved):
Looks great!

PR #354, opened 2024-09-07
feat: WIP implementation of plugin reporting

alilleybrinker:
This is missing some wiring, but is a partial implementation of plugin reporting.


alilleybrinker:
This is superseded by #361. Closing!

PR #362, opened 2024-09-09
fix: make HcEngine plugin lookup use {publisher}/{plugin} as key

j-lanson:
update `HcEngine` interface queries to use proper plugin key

update `hc plugin` dummy plugins to have `dummy` publisher


alilleybrinker (approved):

alilleybrinker:
Can merge as soon as CI passes!

PR #367, opened 2024-09-10
fix(release): temporarily shift mac sys deps to custom build setup

### LINKED ISSUE #366: Fix broken release build
This is a workaround for https://github.com/axodotdev/cargo-dist/issues/1407, since we don't want to wait for the `cargo-dist` folks to publish a fix (though they are working on it and I really appreciate that).

The basic idea will be to use the [`github-build-setup`](https://github.com/axodotdev/cargo-dist/blob/main/book/src/ci/customizing.md#customizing-build-setup) feature of `cargo-dist` to workaround this bug with `cargo-dist` itself trying to install our dependencies through Homebrew.

The basic fix will be:

- Comment out this code, so `cargo-dist` doesn't try to run Homebrew: https://github.com/mitre/hipcheck/blob/main/Cargo.toml#L56-L58
- Write a custom build setup that only triggers on macOS and installs `protobuf` with Homebrew.

Once we've done this, we can publish 3.6.2 as just `3.6.1` with the release automation fix.

ashleygwilliams:
fixes #366


alilleybrinker (approved):

alilleybrinker:
Thank you so much!

PR #372, opened 2024-09-10
chore: Revert 'cargo-dist' workaround

alilleybrinker:
We hit a bug in 'cargo-dist' yesterday that it turns out was already fixed in the latest version, but before we figured that out we implemented a workaround to enable us to get 3.6.2 out the door. Now that we know the latest 'cargo-dist' version doesn't need the workaround, we can revert it and use 0.22.1.


alilleybrinker:
Haha the `plan` step saved me multiple times here, catching errors that would have broken my release!

mchernicoff (approved):

PR #373, opened 2024-09-10
fix: Remove scoring panic.

alilleybrinker:
There was a `panic` call in some scoring code that meant we would panic if we hit an error evaluating a policy expression instead of returning it nicely. This converts that error into a nicer one.


j-lanson (approved):
Looks good!

PR #374, opened 2024-09-10
chore: Removes module analysis, data path, and scripts folder

### LINKED ISSUE #350: Remove "scripts" dir now that we don't do per-module analyses
The one script in the "scripts" directory was for per-module analysis support. Since we no longer do that, we can remove it. This also moves us closer to eliminating the need for a post-install step. Right now, our post-install `setup` command copies over this one script, plus a bundle of configuration files. With the move to the policy file concept, one of my goals is to reduce the need for all of these additional files, instead centralizing configuration in the KDL file (and having it be optional with defaults), so I think we can get to a place where `setup` isn't _required_, or at least is just used to generate a `Hipcheck.kdl` starter file in your current directory

mchernicoff:
Removes the module analysis and any of its dependent code. Because the `module-deps.json` file is no longer needed, this also removes any code, including in the CLI, related to specifying or locating the "data folder." And this removes the extraneous `scripts` folder, which resolves #350 .


mchernicoff:
> I think you'll _also_ need to remove mention of the `scripts` folder in the `cargo-dist` config and the `Containerfile`.

Oh good catch.

alilleybrinker (approved):

PR #379, opened 2024-09-10
chore: Removing dead code

alilleybrinker:
We have a lot of dead code left over from prior work, especially from the transition to the plugin system. This commit is the start of removing a lot of that dead code while ensuring everything still compiles and tests still pass.


alilleybrinker:
By my count, even after this PR, there are 80 instances of allow-unused or allow-dead-code in the codebase.

mchernicoff (approved):

PR #402, opened 2024-09-12
chore: WIP RFD for the plugin SDK

alilleybrinker:


site/content/rfds/0006-rust-plugin-sdk.md
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	/// Get all schemas for queries provided by the plugin.
+	fn schemas(&self) -> impl Iterator<Item = QuerySchema> {
+		self.queries().map(|query| QuerySchema {
+			query_name: query.name,
+			input_schema: query.inner.input_schema(),
+			output_schema: query.inner.output_schema(),
+		})
+	}
```
j-lanson, line 70: This implementation doesn't address the "default query explanation" gRPC call we added during the RFD4 impl

site/content/rfds/0006-rust-plugin-sdk.md
```diff
@@ -0,0 +1,285 @@
...
+	const NAME: &'static str;
+
+	/// Handles setting configuration.
+	fn set_config(&mut self, config: JsonValue) -> StdResult<(), ConfigError>;
+
+	/// Gets the plugin's default policy expression.
+	fn default_policy_expr(&self) -> Result<String>;
+
+	/// Get all the queries supported by the plugin.
+	fn queries(&self) -> impl Iterator<Item = NamedQuery>;
```
j-lanson, line 55: Should the `struct` definition of `NamedQuery` be anywhere in here? We explain why `Query` as a trait can't have `NAME: &'static str`, but then don't discuss how we address that problem with `NamedQuery`

j-lanson (changes requested):
Looks good to me, just a few comments

site/content/rfds/0006-rust-plugin-sdk.md
alilleybrinker: Good catch. I'll add that in.

site/content/rfds/0006-rust-plugin-sdk.md
alilleybrinker: Fair point. I can add that!

alilleybrinker:
@j-lanson updated and ready for re-review

j-lanson (approved):

PR #404, opened 2024-09-12
Add Salsa master branch as subtree under `hipcheck-incremental`

j-lanson:
Open to discussion about whether this is the right way to include Salsa's source code in our repo.

I'm not sure what it will look like when we try to pull down new changes from Salsa's main branch. This one adds the entire codebase as a single commit, presumably if/when we update to a newer Salsa commit, we'd want 1 commit that plays all the changes between the current version and the new version.


alilleybrinker:
Taking a look!

alilleybrinker:
Okay, some quick reviewing and I have a few thoughts:

- It looks like the merge commit for the subtree integration doesn't meet the Conventional Commits standard. That should be easy enough to fix.
- The Salsa crate we're importing is itself a workspace (it has a "main" top-level crate `Cargo.toml` that is _also_ a workspace-root manifest, with two more crates under the `components/` directory. In general, workspace-in-workspace doesn't work well, and in our case it looks like `cargo-dist` is upset about it.
- We'll likely need to define some tooling for managing the subtree. I _do_ think a subtree is probably the right thing, but I also want to investigate what Rust Analyzer does, because I know they also vendor Salsa in.

alilleybrinker:
It looks like Rust Analyzer just has a copy that they manage completely by hand and rarely sync from upstream. They also make their own edits to it. So I don't think they're good prior art for us, where we likely _don't_ want to make edits from upstream.

alilleybrinker:
@j-lanson I think it makes sense to close this. Once RFD 7 is approved, we'll plan to not publish to Crates.io for the 3.8.0 release, which will eliminate the need to vendor Salsa.

j-lanson:
Closing this in accordance with @alilleybrinker 's comment above.

PR #405, opened 2024-09-12
feat: Allow any `Expr` at the top level of a Policy Expression

cstepanian:
The existing parsing code requires that the top level expression is always a function call. I'm not sure if this is intentional or an oversight, but the evaluation code seems to have no problem evaluating other expressions at the top level.

This change allows any expression type at all in the parser.

This will enable testing policy expressions with a plain `#t` or `#f` to force success or failure, as well as permit a JSON Pointer at the top level.


mchernicoff (approved):

j-lanson:
Just a little confusion on my part - I dont' see how the changes contained actually implement the above described feature. It just looks like test cases were added to verify that the parser could handle it. Was the actual implementation done in another branch?

alilleybrinker:
This updated `parse` to call `parse_expr` instead of `parse_function`, so that's the only functional change, which does implement the described feature.

Issue #400, opened 2024-09-12
Add native support for date/time type


mchernicoff:

alilleybrinker:
Oh, you should check out the `jiff` crate for this

PR #419, opened 2024-09-18
feat: Adds datetime and span primitives to the policy expression syntax

mchernicoff:
Resolves Issue #400. Adds `Datetime` and `Span` types to the `Primitives` used in the policy expression syntax. Operators that use these types and an RFD explaining their intend use will be added in future commits.


hipcheck/src/policy_exprs/token.rs
```diff
@@ -34,6 +41,12 @@ pub enum Token {
...
 	Integer(i64),
 
+	#[regex(r"[0-9]{3,4}[^\s\)]+", lex_datetime)]
+	DateTime(Zoned),
+
+	#[regex(r"P[a-zA-Z0-9]+", lex_span)]
+	Span(Span),
+
 	#[regex("([a-zA-Z]+)", lex_ident)]
 	Ident(String),
```
j-lanson, line 51: Can we update the identifier regex to expect single quotes? This way there isn't a collision between `P` starting a span and being the first letter of something intended as a string identifier

hipcheck/src/policy_exprs/token.rs
```diff
@@ -68,6 +81,42 @@ fn lex_float(input: &mut Lexer<'_, Token>) -> Result<F64> {
...
+					LexingError::InvalidDatetime(s.to_string(), JiffError::new(err))
+				}),
+				// If the string provided does not parse to a valid date or datetime, return an error
+				Err(err) => Err(LexingError::InvalidDatetime(
+					s.to_string(),
+					JiffError::new(err),
+				)),
+			},
+		},
+	}
```
j-lanson, line 110: I think this is cleaner as a set of `if let Ok(_) = ` blocks (at the same identation level), instead of nesting the `matches`

hipcheck/src/policy_exprs/expr.rs
```diff
@@ -49,6 +50,12 @@ pub enum Primitive {
 
 	/// Boolean.
 	Bool(bool),
+
+	/// Date-time.
+	DateTime(Zoned),
+
+	/// Span of time
+	Span(Span),
```
j-lanson, line 58: Could you update these doc comments to specify the library the types come from and a summary of the string format they expect to parse from (I heard something about no fractional seconds?)

j-lanson: And also specify that it's always interpreted in UTC (if that is indeed the case)

mchernicoff: You can have fractional seconds, but not fractional minutes or hours (which ISO 8601 allows).

The time zone is always set to UTC by the parser to simplify things, but you can specify an offset that the parser will use to adjust the time to the correct UTC time. e.g.  {foo}T9:30-05 will parse to {foo}T14:30["UTC"].

hipcheck/src/policy_exprs/token.rs
mchernicoff: Oh right, I meant to do that. :)

hipcheck/src/policy_exprs/expr.rs
j-lanson: That sounds good. I just want to make sure people reading the API docs can get a clue as to what format is expected and why the string they pass is getting rejected

hipcheck/src/policy_exprs/token.rs
mchernicoff: Would it make sense to leave the last `match` block as a `match`, since we catch the error in that instance?

j-lanson: sounds good to me

hipcheck/src/policy_exprs/expr.rs
mchernicoff: Put some text in the comments. Let me know if it makes sense.

mchernicoff:
Looks like some tests are failing now, including ones unrelated to datetime and span. May have to do with the changes to the error type.

j-lanson:
> Looks like some tests are failing now, including ones unrelated to datetime and span. May have to do with the changes to the error type.

Failures seem related to adding `'` to regex for ident primitive

j-lanson (approved):
Looks good!

PR #426, opened 2024-09-19
feat: Make `xtask rfd new` include title in RFD file

### LINKED ISSUE #425: Make `cargo xtask rfd new` generate the RFD `.md` file with a YAML frontmatter
As-is, when this command is used to generate a new RFD, the file generated is missing a YAML frontmatter, which means it won't compile when built with `zola`. It could instead be given a simple YAML frontmatter which looks like:

```md
---
title: "<TITLE>"
---
```

Where `<TITLE>` is replaced with the title the user provided on the command line.

alilleybrinker:
Previously when making a new RFD, the `xtask` command would generate one without a YAML frontmatter, which would cause compilation with `zola` to fail. This fixes that, now including a YAML frontmatter with a `title` field filled with the title the user provided on the CLI.

Closes #425.

### DIFF
--- a/xtask/src/task/buf.rs
+++ b/xtask/src/task/buf.rs
@@ -1,3 +1,5 @@
+// SPDX-License-Identifier: Apache-2.0
+
 use crate::workspace;
 use anyhow::{Context, Result};
 use pathbuf::pathbuf;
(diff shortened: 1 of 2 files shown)


j-lanson (approved):
Looks good!

PR #427, opened 2024-09-19
feat: added preliminary SDK support to "rand_data" plugin

patrickjcasey:
This pull request is the first pass at using the `hipcheck-sdk` crate to make a plugin for `hipcheck`

The goal of the SDK is to make it as easy as possible for plugin authors to write new plugins for hipcheck!

The hidden `hc plugin` command no accepts another flag `--sdk` that will use the `dummy_rand_data_sdk` plugin, rather than the `dummy_rand_data` plugin

There are some outstanding issues that still need to be addressed, whether in this issue, or in another issue in the future:
- How are we handling third-party dependencies that are exposed in public facing types (`serde_json::Value`, `schemars::schema::SchemaObject`, ...)
- How much documentation should be present in the `hipcheck-sdk` crate
- How do we test the `hipcheck-sdk` crate to ensure it is producing the desired behaviors and integrates cleanly with what `hipcheck` is expecting
- Does it make sense to add a `prelude` module so a user would be able to `use hipcheck_sdk::prelude::*` and have the most common imports?

### DIFF
--- a/hipcheck/src/plugin/types.rs
+++ b/hipcheck/src/plugin/types.rs
@@ -478,6 +478,7 @@ impl PluginTransport {
 
 		// Send the query
 		let query: PluginQuery = query.try_into()?;
+
 		let id = query.id;
 		self.tx
 			.send(query)
@@ -499,6 +500,8 @@ impl PluginTransport {
 			while matches!(state, ReplyInProgress) {
 				// We expect another message. Pull it off the existing queue,
 				// or get a new one if we have run out
+				eprintln!("In progress");
+
 				let next = match msg_chunks.pop_front() {
 					Some(msg) => msg,
 					None => {
--- a/plugins/dummy_rand_data_sdk/src/main.rs
+++ b/plugins/dummy_rand_data_sdk/src/main.rs
@@ -1,0 +1,118 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::Result;
+use clap::Parser;
+use hipcheck_sdk::{
+	deps::{async_trait, from_str, JsonSchema, Value},
+	error::Error,
+	plugin_engine::PluginEngine,
+	plugin_server::PluginServer,
+	NamedQuery, Plugin, Query,
+};
+
+static GET_RAND_KEY_SCHEMA: &str = include_str!("../schema/query_schema_get_rand.json");
+static GET_RAND_OUTPUT_SCHEMA: &str = include_str!("../schema/query_schema_get_rand.json");
+
+fn reduce(input: u64) -> u64 {
+	input % 7
+}
+
+/// Plugin that queries hipcheck takes a `Value::Number` as input and performs following steps:
+/// - ensures input is u64
+/// - % 7 of input
+/// - queries `hipcheck` for sha256 of (% 7 of input)
+/// - returns `Value::Number`, where Number is the first `u8` in the sha256
+///
+/// Goals of this plugin
+/// - Verify `salsa` memoization is working (there should only ever be 7 queries made to `hipcheck`)
+/// - Verify plugins are able to query `hipcheck` for additional information
+#[derive(Clone, Debug)]
+struct RandDataPlugin;
+
+#[async_trait]
+impl Query for RandDataPlugin {
+	fn input_schema(&self) -> JsonSchema {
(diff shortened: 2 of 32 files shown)


alilleybrinker:
Thanks @patrickjcasey! I'll take a look tomorrow! :smile:

hipcheck/src/plugin/types.rs
```diff
@@ -499,6 +500,8 @@ impl PluginTransport {
 			while matches!(state, ReplyInProgress) {
 				// We expect another message. Pull it off the existing queue,
 				// or get a new one if we have run out
+				println!("In progress");
```
j-lanson, line 503: delete or switch to `eprintln`

plugins/dummy_rand_data_sdk/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+		from_str(self.schema.output_schema.as_str()).unwrap()
+	}
+
+	async fn run(
+		&self,
+		engine: &mut PluginEngine,
+		input: Value,
+	) -> hipcheck_sdk::error::Result<Value> {
+		let Value::Number(num_size) = input else {
+			return Err(Error::UnexpectedPluginResponseDataFormat);
```
j-lanson, line 47: the error type saying `Response` when this is checking the type of the "request" key feels wrong

sdk/rust/src/plugin_server.rs
```diff
@@ -0,0 +1,186 @@
...
+
+/// The result of running a query.
+pub type QueryResult<T> = StdResult<T, Status>;
+
+#[tonic::async_trait]
+impl<P: Plugin> PluginService for PluginServer<P> {
+	type GetQuerySchemasStream = RecvStream<QueryResult<GetQuerySchemasResp>>;
+	type InitiateQueryProtocolStream = RecvStream<QueryResult<InitiateQueryProtocolResp>>;
+
+	// @Todo - implement these calls by calling appropriate functions from `plugin`
```
j-lanson, line 67: Remove todo comment

sdk/rust/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+pub use tonic::async_trait;
+
+use crate::error::Error;
+use crate::error::Result;
+use error::ConfigError;
+use plugin_engine::PluginEngine;
+use serde_json::Value as JsonValue;
+use std::result::Result as StdResult;
+use std::str::FromStr;
+
```
j-lanson, line 27: To answer your re-export question, I would just make a `pub mod export/deps { }` or something similar in this `lib.rs` file that declares all the types we need to re-export. I defer to Andrew if he has any opinions on this

sdk/rust/src/lib.rs
```diff
@@ -0,0 +1,140 @@
+// SPDX-License-Identifier: Apache-2.0
+
+pub mod error;
+pub mod proto {
+	include!(concat!(env!("OUT_DIR"), "/hipcheck.v1.rs"));
+}
```
j-lanson, line 6: If we don't need this to be public I don't think we want it to be

j-lanson:
Re: a prelude crate

As far as I see, excluding the re-exports the only types users need access to is `QuerySchema, DynQuery, NamedQuery, Query, Plugin, PluginServer, PluginEngine`, plus the `error` module. And anyone using the SDK to write a plugin will need all of these. 

I'd be in favor of adding a `prelude` crate so that users that want to do so can simply import `*` as you have shown above.

Leave this and the reexport crate as a separate commit on top for now in case we need to walk it back

j-lanson:
As for documentation and testing, those are important, but they can be done after the SDK is merged

sdk/rust/src/lib.rs
patrickjcasey: for now I added a `deps` module that contains all of the types that need to be re-exported

alilleybrinker:
I concur on the idea of a `prelude` module.

sdk/rust/src/lib.rs
patrickjcasey: I added a prelude module, and updated the `dummy_rand_sdk` plugin to use the `use hipcheck_sdk::prelude::*` import

j-lanson (approved):
This looks good to me! Great work

PR #433, opened 2024-09-20
feat: add --arch flag to `hc check` subcommand to allow detection override

### LINKED ISSUE #415: Fails to compile on not-explicitly-listed architecture
See: https://github.com/mitre/hipcheck/blob/e0892bbffad9ffb1f49ae92a93cab86469e204b0/hipcheck/src/plugin/supported_arch.rs#L22

The current approach to the `CURRENT_ARCH` has two effects:

- Hipcheck can only compile on one of the listed architectures.
- Users can't _force_ the inference of a specific operation system.

Ideally, we'd replace this compile-time check with a runtime check, _and_ give the user a means via configuration to specify the architecture to use. If the user hasn't set one and we can't infer one, then error out at runtime.

j-lanson:
Resolves #415 .

We had previously encountered problems with cross-compilation to niche-but-valid architectures, where the compile-time target detection would fail, causing `CURRENT_ARCH` to be `()` instead of a valid arch enum variant. We changed this field to be type `Option<SupportedArch>` and also support a user-provided `--arch` flag to override the detected arch at runtime if the user has good reason to do so.

### DIFF
--- a/hipcheck/src/cli.rs
+++ b/hipcheck/src/cli.rs
@@ -7,6 +7,7 @@ use crate::{
 	error::Context,
 	error::Result,
 	hc_error,
+	plugin::SupportedArch,
 	session::pm,
 	shell::{color_choice::ColorChoice, verbosity::Verbosity},
 	source,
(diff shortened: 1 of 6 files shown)


alilleybrinker:
Thanks @j-lanson! I think this is part of what we need, and I'm happy to merge as-is. The part that's missing is expanding the `SupportedArch` enum to reflect more architectures we may not auto-detect, but which we can allow users to specify. In my specific case that would be `x86_64-unknown-linux-gnu" but there are certainly others we could reflect in the enum.

alilleybrinker (approved):

PR #434, opened 2024-09-20
feat: allow use of unknown arches through commandline

j-lanson:
This replaces `SupportedArch` with an `Arch` enum, which is further subdivided into `Unknown(String)` and the `KnownArch` enum.

Currently `DETECTED_ARCH` is `Option<KnownArch>`, but we could expand it to be just `Arch`, and try to get the target triple at compile time as a string to fill the `Arch::Unknown` variant. Without doing this, right now you'll be forced to specify `--arch` on the command line even if you build and run on the same arch, assuming its not one of the `KnownArch` types.

### DIFF
--- a/hipcheck/build.rs
+++ b/hipcheck/build.rs
@@ -5,8 +5,18 @@ use pathbuf::pathbuf;
 use tonic_build::compile_protos;
 
 fn main() -> Result<()> {
+	// Compile the Hipcheck gRPC protocol spec to an .rs file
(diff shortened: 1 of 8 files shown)


alilleybrinker (approved):
LGTM!

alilleybrinker (changes requested):
Just needs a squash actually, but otherwise LGTM!

alilleybrinker (approved):

PR #441, opened 2024-09-24
feat: Adds operators for datetime and span policy expression primitives

### LINKED ISSUE #401: Add date/time operators

mchernicoff:
Resolves #401 

Adds comparison and basic arithmetic operators for `DateTime` and `Span` policy expression primitives.

To handle comparing `Spans` without erroring or needing to provide a datetime for each span, `Spans` must now not include years or months as units. `Spans` may include weeks as units, even though `jiff` normally has issues with comparing raw `Spans` with weeks, because we convert all weeks to seven 24-hour days during parsing.

### DIFF
--- a/hipcheck/src/policy_exprs/env.rs
+++ b/hipcheck/src/policy_exprs/env.rs
@@ -55,6 +55,9 @@ impl<'parent> Env<'parent> {
 		env.add_fn("sub", sub);
 		env.add_fn("divz", divz);
 
+		// Additional datetime math functions
+		env.add_fn("duration", duration);
+
 		// Logical functions.
 		env.add_fn("and", and);
 		env.add_fn("or", or);
@@ -325,6 +328,13 @@ fn gt(env: &Env, args: &[Expr]) -> Result<Expr> {
 		(Int(arg_1), Int(arg_2)) => Ok(Bool(arg_1 > arg_2)),
 		(Float(arg_1), Float(arg_2)) => Ok(Bool(arg_1 > arg_2)),
 		(Bool(arg_1), Bool(arg_2)) => Ok(Bool(arg_1 > arg_2)),
+		(DateTime(arg_1), DateTime(arg_2)) => Ok(Bool(arg_1 > arg_2)),
+		(Span(arg_1), Span(arg_2)) => Ok(Bool(
+			arg_1
+				.compare(arg_2)
+				.map_err(|err| Error::Datetime(err.to_string()))?
+				.is_gt(),
+		)),
 		_ => unreachable!(),
 	};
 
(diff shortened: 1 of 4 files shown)


mchernicoff:
Some of the span comparisons got lost when I made some changes before pushing. Will go put those back in tomorrow. Leaving this as a drat until then.

hipcheck/src/policy_exprs/env.rs
```diff
@@ -395,32 +433,58 @@ fn neq(env: &Env, args: &[Expr]) -> Result<Expr> {
...
 		(Int(arg_1), Int(arg_2)) => Ok(Int(arg_1 + arg_2)),
 		(Float(arg_1), Float(arg_2)) => Ok(Float(arg_1 + arg_2)),
 		(Bool(_), Bool(_)) => Err(Error::BadType(name)),
+		(DateTime(arg_1), Span(arg_2)) => Ok(DateTime(
+			arg_1
+				.checked_add(arg_2)
+				.map_err(|err| Error::Datetime(err.to_string()))?,
+		)),
+		(DateTime(arg_1), DateTime(arg_2)) => Err(Error::BadType(name)),
+		(Span(arg_1), Span(arg_2)) => Err(Error::BadType(name)),
```
j-lanson, line 464: Semantically it seems fine to add two spans (e.g. 2 days plus 4 weeks = 4w2d). Does the `Span` API not allow for that?

hipcheck/src/policy_exprs/env.rs
```diff
@@ -446,6 +510,30 @@ fn divz(env: &Env, args: &[Expr]) -> Result<Expr> {
...
+}
+
+// Finds the difference in time between two datetimes, in units of hours (chosen for comparision safety)
+fn duration(env: &Env, args: &[Expr]) -> Result<Expr> {
+	let name = "duration";
+
+	let op = |arg_1, arg_2| match (arg_1, arg_2) {
+		(Int(_), Int(_)) => Err(Error::BadType(name)),
+		(Float(_), Float(_)) => Err(Error::BadType(name)),
+		(Bool(_), Bool(_)) => Err(Error::BadType(name)),
```
j-lanson, line 529: In general, is it possible to replace a lot of these `(Type(_), Type(_)) => Error::BadType` with a single `(_, _) => Error::BadType`?

j-lanson (changes requested):
Looks great! Just a couple requests.

hipcheck/src/policy_exprs/env.rs
mchernicoff: Oh we can do that. But you didn't list it in one of the features we wanted, so I didn't include it. I can add that in.

hipcheck/src/policy_exprs/env.rs
mchernicoff: I think that should work. Let me try it.

mchernicoff: Made the change. Let me know if that's what you were thinking.

j-lanson: Yeah looks good, there are still a few `(Bool(_), Bool(_)) => Err(Error::BadType(name)),` floating around, such as in the `add` function impl

mchernicoff: Great, I'll do the rest!

j-lanson (approved):
Just gotta wait for the CI to pass, then I'll merge! Thanks Michael

PR #445, opened 2024-09-24
feat: dummy sha256 plugin using hipcheck-sdk

### LINKED ISSUE #423: Transition dummy plugins to use the SDK

patrickjcasey:
This branch implements the `dummy/sha256` plugin with the `hipcheck-sdk`

This will close #423

### DIFF
--- a/plugins/dummy_sha256_sdk/src/main.rs
+++ b/plugins/dummy_sha256_sdk/src/main.rs
@@ -1,0 +1,89 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use clap::Parser;
+use hipcheck_sdk::prelude::*;
+use sha2::{Digest, Sha256};
+
+static SHA256_KEY_SCHEMA: &str = include_str!("../schema/query_schema_sha256.json");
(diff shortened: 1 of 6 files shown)


j-lanson:
@mchernicoff can you take a look?

plugins/dummy_sha256_sdk/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+		Ok(())
+	}
+
+	fn default_policy_expr(&self) -> hipcheck_sdk::prelude::Result<String> {
+		Ok("".to_owned())
+	}
+
+	fn default_query(&self) -> Option<DynQuery> {
+		Some(Box::new(Self))
+	}
```
j-lanson, line 68: I don't think you're supposed to implement this function yourself. The default query will be the `NamedQuery` with `name.empty()`. This plugin won't have a default query, unless you want to change `NamedQuery` below and update `dummy_rand_data_sdk` to just query `dummy/sha256`.

patrickjcasey: I think it's cleaner to just not implement this function

j-lanson (approved):
Looks good! Thanks Patrick

Issue #387, opened 2024-09-11
🛤️ Policy Expressions 2.0 - Tracking Issue


alilleybrinker:
- [x] #457
- [x] #471
- [x] Implement late-binding for JSON pointers
- [x] #400
- [x] #401

PR #451, opened 2024-09-26
fix: policy expr type errors due to X.0 floating points in JSON being treated as ints

### LINKED ISSUE #449: Identity analysis in config file format has invalid policy expression
**Describe the bug**
When running Hipcheck with identity analysis turned on, using an older-style config file, the config-to-policy conversion results in a type error for `lte` in the default policy expression. Turning off `identity` analysis _does_ stop the issue, confirming that it's specific to _that_ analysis.

**To Reproduce**
1. Install latest `hc`
2. Run an analysis with `analysis.practices.identity.active = true` in the config TOML file

j-lanson:
Resolves #449 .

Pernicious bug in policy expression code. JSON has only "Number" type, doesn't distinguish between float and int. When a float that ends in `.0` gets serialized then injected into a policy expression, it is treated as an integer. Our policy expression evaluation logic for arithmetic/comparison requires operands be `int + int` or `float + float`, so when this bug occurs we get a type error because we do `int + float`. 

This fix updates `binary_primitive_op` to detect when the operands are `int+float` or `float+int` and implicitly upcasts the `int` to a `float`. It also removes other type checking from `binary_primitive_op`, as it was bad design. For example, although we updated `fn add` to allow `DateTime + Span`, the type checking in `binary_primitive_op` did not allow this to occur and we had not tested to encounter this bug before.

Added tests to ensure int+float arithmetic/comparisons return floats, and ensure `(add datetime span)` does not error.

In a more correct AST design, we would do a first-pass eval of the AST and insert "cast" nodes around integer primitives to make them floats. This fix approach works but doesn't afford per-op control, highlighting the need for #387 .

### DIFF
--- a/hipcheck/src/engine.rs
+++ b/hipcheck/src/engine.rs
@@ -231,7 +231,6 @@ pub fn start_plugins(
 	)?;
 
 	let current_arch = get_current_arch();
-	println!("CURRENT ARCH: {}", current_arch);
 
 	// retrieve, verify and extract all required plugins
 	let required_plugin_names = retrieve_plugins(&policy_file.plugins.0, plugin_cache)?;
--- a/hipcheck/src/policy_exprs/env.rs
+++ b/hipcheck/src/policy_exprs/env.rs
@@ -142,6 +142,10 @@ fn partially_evaluate(fn_name: &'static str, arg: Expr) -> Result<Expr> {
 	Ok(lambda)
 }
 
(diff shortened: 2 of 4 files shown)


mchernicoff (dismissed):

alilleybrinker (approved):
Thanks Julian, and yeah I agree on the need for a refactor of this code to address stuff like this more preemptively.

PR #452, opened 2024-09-26
feat: Adds query sdk proc-macro

mchernicoff:
Adds a Hipcheck SDK proc-macro that can mark functions to have a struct `impl Query` created.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ members = [
     "plugins/dummy_sha256",
     "plugins/dummy_sha256_sdk",
     "sdk/rust",
+    "hipcheck-sdk-macros",
 ]
 
 # Make sure Hipcheck is run with `cargo run`.
--- a/hipcheck-sdk-macros/src/lib.rs
+++ b/hipcheck-sdk-macros/src/lib.rs
@@ -1,0 +1,240 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use convert_case::Casing;
+use proc_macro::TokenStream;
+use proc_macro2::Span;
+use std::ops::Not;
+use std::sync::{LazyLock, Mutex};
+use syn::spanned::Spanned;
+use syn::{parse_macro_input, Error, Ident, ItemFn, Meta, PatType};
+
+static QUERIES: LazyLock<Mutex<Vec<NamedQuerySpec>>> = LazyLock::new(|| Mutex::new(vec![]));
+
+#[allow(unused)]
+#[derive(Debug, Clone)]
+struct NamedQuerySpec {
+	pub struct_name: String,
+	pub function: String,
+	pub default: bool,
+}
+
+struct QuerySpec {
+	pub function: Ident,
+	pub input_type: syn::Type,
+	pub output_type: syn::Type,
+	pub default: bool,
+}
+
+/// Parse Path to confirm that it represents a Result<T: Serialize> and return the type T
+fn parse_result_generic(p: &syn::Path) -> Result<syn::Type, Error> {
+	use syn::GenericArgument;
+	use syn::PathArguments;
+	// Assert it is a Result
+	// Panic: Safe to unwrap because there should be at least one element in the sequence
+	let last = p.segments.last().unwrap();
+	if last.ident != "Result" {
+		return Err(Error::new(
+			p.span(),
+			"Expected return type to be a Result<T: Serialize>",
+		));
+	}
+	match &last.arguments {
+		PathArguments::AngleBracketed(x) => {
+			let Some(GenericArgument::Type(ty)) = x.args.first() else {
+				return Err(Error::new(
+					p.span(),
+					"Expected return type to be a Result<T: Serialize>",
+				));
+			};
+			Ok(ty.clone())
+		}
+		_ => Err(Error::new(
+			p.span(),
+			"Expected return type to be a Result<T: Serialize>",
+		)),
+	}
+}
+
+/// Parse PatType to confirm that it contains a &mut PluginEngine
+fn parse_plugin_engine(engine_arg: &PatType) -> Result<(), Error> {
+	if let syn::Type::Reference(type_reference) = engine_arg.ty.as_ref() {
+		if type_reference.mutability.is_some() {
+			if let syn::Type::Path(type_path) = type_reference.elem.as_ref() {
+				let last = type_path.path.segments.last().unwrap();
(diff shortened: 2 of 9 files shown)


j-lanson:
I made sure that `hipcheck-sdk-macros` doesn't actually depend on `hipcheck-sdk` and removed it as a dependency.

Cargo.toml
```diff
@@ -19,7 +19,7 @@ members = [
     "plugins/dummy_rand_data",
     "plugins/dummy_sha256",
     "plugins/dummy_sha256_sdk",
-    "sdk/rust",
+    "sdk/rust", "hipcheck-sdk-macros",
```
j-lanson, line 22: on a newline if the formatter will allow

hipcheck-sdk-macros/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+		}
+	};
+
+	println!("TO FOLLOW:\n{}", to_follow);
+	let named_spec = NamedQuerySpec {
+		struct_name: struct_name.to_string(),
+		function: spec.function.to_string(),
+		default: spec.default,
+	};
+	println!("SPEC: {named_spec:?}");
```
j-lanson, line 238: remove these print lines

plugins/dummy_sha256_sdk/Cargo.toml
```diff
@@ -10,3 +10,6 @@ clap = { version = "4.5.18", features = ["derive"] }
 hipcheck-sdk = { path = "../../sdk/rust" }
 sha2 = "0.10.8"
 tokio = { version = "1.40.0", features = ["rt"] }
+hipcheck-sdk-macros = { path = "../../hipcheck-sdk-macros", version = "0.1.0" }
+serde_json = "1.0.128"
+schemars = "0.8.21"
```
j-lanson, line 15: one of the things Andrew mentioned wanting was for `hipcheck_sdk` to re-export the `hipcheck_sdk_macros` proc macro. Please try to make this work, and then see if we can make it so `serde_json` and `schemars` are not also explicit dependencies (add anything we need for the macro to the `hipcheck_sdk::prelude` module maybe?)

plugins/dummy_sha256_sdk/src/main.rs
```diff
@@ -14,6 +15,13 @@ fn sha256(content: &[u8]) -> Vec<u8> {
 	hasher.finalize().to_vec()
 }
 
+#[query(default = false)]
+fn altsha256(_engine: &mut PluginEngine, content: Vec<u8>) -> Result<Vec<u8>> {
```
j-lanson, line 19: this function should need to be `async` so it *could* call `engine::query` if need be. 

also, replace use of `sha256` and its `Query` impl. rename alt func to `sha256`.

hipcheck-sdk-macros/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+	pub output_type: syn::Type,
+	pub default: bool,
+}
+
+/// Parse Path to confirm that it represents a Result<T: Serialize> and return the type T
+fn parse_result_generic(p: &syn::Path) -> Result<syn::Type, Error> {
+	use syn::GenericArgument;
+	use syn::PathArguments;
+	// Assert it is a Result
+	let first = p.segments.first().unwrap();
```
j-lanson, line 33: This will fail if the return type Path is `hipcheck_sdk::prelude::Result` instead of just `Result`.
```
error: Expected return type to be a Result<T: Serialize>
  --> plugins/dummy_sha256_sdk/src/main.rs:22:6
   |
22 | ) -> hipcheck_sdk::prelude::Result<Vec<u8>> {
```

I believe you want use `segments.last()` to get the type-name part of the path

j-lanson (changes requested):
There are some things that should be addressed before we merge

sdk/rust/src/lib.rs
```diff
@@ -170,3 +172,6 @@ pub trait Plugin: Send + Sync + 'static {
 		})
 	}
 }
+
+// Re-export macros
+pub use hipcheck_sdk_macros::query;
```
patrickjcasey, line 177: should this be added to the prelude import, so it is imported with `use hipcheck_sdk::prelude::*`

j-lanson: good question. I think the answer is yes, but @alilleybrinker might have opinions on default-importing proc macros. lets say yes to including it for now

mchernicoff: That would make sense. I can move it.

patrickjcasey: it could be put behind a macros feature flag and be imported if `macros` is enabled

j-lanson: actually yeah I like that. michael can you implement that?

mchernicoff: Let me know if it looks good now.

sdk/rust/Cargo.toml
```diff
@@ -15,7 +15,11 @@ tokio = { version = "1.39.2", features = ["rt"] }
...
 tonic = "0.12.1"
 schemars = "0.8.21"
+hipcheck-sdk-macros = { path = "../../hipcheck-sdk-macros", version = "0.1.0" }
 
 [build-dependencies]
 anyhow = "1.0.86"
 tonic-build = "0.12.1"
+
+[features]
+macros = []
```
patrickjcasey, line 25: this can be updated to conditionally depend on hipcheck-sdk-macros

it's something like:

```rust
[dependencies]
...
hipcheck-sdk-macros = { path = "../../hipcheck-sdk-macros", version = "0.1.0", optional = true }

[features]
macros = ["hipcheck-sdk-macros"]

Comment (patrickjcasey) [reply sdk/rust/Cargo.toml:25]: this has the benefit of only compiling the `hipcheck-sdk-macros` crate, if the user specified to use the `macro` crate

File: sdk/rust/src/lib.rs [hunk 8]
```diff
@@ -9,6 +9,8 @@ use serde_json::Value as JsonValue;
 use std::result::Result as StdResult;
 use std::str::FromStr;
 
+extern crate hipcheck_sdk_macros;
```
j-lanson, line 12: This also has to be gated with `cfg(feature = "macros")`

j-lanson (changes requested):
One more thing

j-lanson (approved):
Thanks Michael, looks great!

PR #476, opened 2024-10-08
feat: add "mock_engine" feature to SDK to allow offline query testing

### LINKED ISSUE #431: Rust SDK should include a test mechanism to mock out query responses
If we want plug-in authors to be able to test their plugins, we should provide a test API that lets the plug-in author write in responses for queries that let them test their own code.

j-lanson:
Resolves #431 .

Adds a feature flag to the Rust SDK called `"mock_engine"` that gates `PluginEngine` functionality and some imports. 

Adds a field to the `PluginEngine` called `mock_responses` that contains a lookup table. When `query()` is called, if the `"mock_engine"` feature is enabled, it will use that table instead of querying out to the Hipcheck core. This allows us to test query logic without spinning up a whole Hipcheck analysis. 

The SDK version of the `rand_data` plugin is updated to use the SDK macro system, and a test case has been added to demonstrate how plugins can be tested locally with `mock_responses`.

The `cfg` compiler flags have been carefully applied to ensure no "unused" warnings whether running with `"mock_engine"` or not.

I've temporarily left the commits split between what @patrickjcasey had done before he went on leave and what I have since added.

### DIFF
--- a/plugins/dummy_rand_data_sdk/Cargo.toml
+++ b/plugins/dummy_rand_data_sdk/Cargo.toml
@@ -8,6 +8,9 @@ publish = false
 [dependencies]
 anyhow = "1.0.87"
 clap = { version = "4.5.18", features = ["derive"] }
-hipcheck-sdk = { path = "../../sdk/rust" }
+hipcheck-sdk = { path = "../../sdk/rust", features = ["macros"] }
 rand = "0.8.5"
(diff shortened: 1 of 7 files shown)


mchernicoff (approved):

j-lanson:
Squashed

PR #477, opened 2024-10-10
feat(refactor): Created plugin version of git queries

mchernicoff:
Resolves issue #472 

This is mainly a refactor of extant code from `hipcheck/src/data/{mod.rs, git::*}`. The Salsa queries have been changed to query functions compatible with the Hipcheck plugin query system.

We have not refactored data structures and functions related to commit signers, as no Hipcheck analysis currently uses those. These can be added in a future update if desired.

Note that queries involving datetimes accept and return datetime values as plain Strings in the correct formatting for `jiff::Timestamp.` Other plugins should parse and display datetimes accordingly when using the functions in this plugin.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1041,6 +1041,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "git_sdk"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "clap",
+ "hipcheck-sdk",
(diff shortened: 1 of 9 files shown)


mchernicoff:
Draft because several of the design choices I made are open to discussion before final merging.

j-lanson (approved):
Looks good to me

PR #481, opened 2024-10-10
fix: update language on handling of lambdas

### LINKED ISSUE #464: Ambiguity with which op of lambda is replaced
Looking in the `policy_exprs` code, it implicitly puts the "lambda" op as the first in a binary pair, but when I was writing documentation for policy exprs and writing the default exprs in a policy file, I misunderstood and wrote them as if the missing op would go in the second op slot. Rectify the language in documentation / policy file and validate that everything works as expected.

j-lanson:
Resolves #464


alilleybrinker (dismissed):

alilleybrinker:
@j-lanson looks good; needs rebase. Ping me for re-review when ready.

PR #484, opened 2024-10-10
feat: Reorg libs and update protobuf defs

alilleybrinker:
This commit moves the two Hipcheck "macro" libraries (`hipcheck-macros` and `hipcheck-sdk-macros`) under the `libraries/` folder, which going forward will be used as the directory for any support library crates we define in the Hipcheck project.

It also deduplicates the protobuf definition files, ensuring that all users of them are synced with the central definition, and updates them to comply to `buf`'s recommended best practices for protobuf files. This includes updates to all users of the definitions to comply with the new specification. Note that these changes are backwards incompatible, as they involve changing the names of existing calls in the gRPC interface.

This is a breaking change, but given that the plugin interface is still experimental and without real operational users, this is the right time to get the interface definition right without letting ourselves be locked in to compatibility.


libraries/hipcheck-sdk-macros/src/lib.rs
```diff
@@ -265,10 +265,6 @@ pub fn queries(_item: TokenStream) -> TokenStream {
 		};
 		agg.extend(out);
 	}
-	eprintln!(
-		"Auto-generating Plugin::queries() with {} detected queries",
-		q_lock.len()
-	);
```
j-lanson, line 271: Since `queries! {}` relies on `#[query]` functions preceding its invocation, I had this print line as a sanity check for users to ensure the number of queries it sees is the right amount

j-lanson:
```
./target/debug/hc plugin
Starting HcPluginCore
Failed to create engine: status: Unimplemented, message: "", details: [], metadata: MetadataMap { headers: {"date": "Fri, 11 Oct 2024 13:11:35 GMT", "content-type": "application/grpc", "content-length": "0"} }
```

j-lanson (changes requested):
Bug needs addressing

alilleybrinker:
I really need to add an automated smoke test for this kind of failure.

libraries/hipcheck-sdk-macros/src/lib.rs
alilleybrinker: I am generally against printing during macro compilation by default. If we instead wanted to check an environment variable like `HC_DEBUG` and print if it's present, that would be fine.

alilleybrinker:
Rebased; have not yet addressed the other comments.

libraries/hipcheck-sdk-macros/src/lib.rs
j-lanson: Using a debug env var works for me

alilleybrinker:
Added smoke tests, identified test failure; now working on resolving.

libraries/hipcheck-sdk-macros/src/lib.rs
alilleybrinker: Done!

alilleybrinker:
Figured out the bug. It was actually not always able to find the binaries it was trying to run during a test. Added some logging to the plugin spawning code which let me track the issue down. Looks like this needs a rebase, which I'll take care of, but at least now I feel confident in the changes.

alilleybrinker:
Agh, tests still failing here. It seems that the problem here is that this test relies on having compiled the plugins so they're available to run. When running an integration test Cargo makes sure the crate you're attached to has been built, but it doesn't make sure the whole _workspace_ has been built. So that's my next thing to resolve then.

Easiest solution would be to call Cargo to do the build from inside the test, though I'm not sure if that's the best solution.

alilleybrinker:
Deciding to close this. I still want to bring in the tests proposed here, and some of the reorg stuff I think is useful, but this is too many changes in one PR, and it's drifted substantially from `main` now.

PR #487, opened 2024-10-11
feat: hc build emits target schemas, sdk interprets with typify

j-lanson:
Our plugins need to use types from `hipcheck::target::types.rs`. A simple example of this is that each top-level analysis's default query needs to accept an instance of `Target`. We don't want to have to re-define these types in each of our simple plugins, but we also don't want to take them directly from Hipcheck - we should enforce a code-separation boundary between our SDK and the Hipcheck core, both for hygiene and so that we don't treat the Rust SDK overly deferentially and "better" than other potential language SDKs. 

The solution herein contained is to have `hipcheck` auto-generate schemas for these types as part of its build, and place them in `sdk/schema`, a directory accessible by any SDK. 

The Rust SDK has been changed to include a module `types.rs` which uses the `typify` crate on the generated schema to produce Rust struct definitions. We then make this module accessible to plugins using the SDK. 

To make it so `hipcheck::target::types.rs` doesn't rely on other parts of the hipcheck crate, I moved a few traits from `types.rs` into `mod.rs` of the same module.


alilleybrinker:
Is there a reason a plugin would need any of the schemas besides the `Target` schema?

j-lanson:
> Is there a reason a plugin would need any of the schemas besides the `Target` schema?

As an example, the `git` data plugin uses only `LocalGitRepo`, and the `github_api` uses only `RemoteGitRepo`. There might be supporting data analyses that only provide info for certain classes of targets (like packages), where the analysis is not applicable otherwise

mchernicoff (approved):

PR #497, opened 2024-10-14
feat: add `patch` field to policy file to allow config of dependent plugins

### LINKED ISSUE #495: Add `patch {}` section to policy file for passing configuration to dependent plugins
Add a patch {} section to the policy file that is only necessary when a dependency needs configuration
```rust
patch {
        plugin "mitre/gitlab_api" {
                api-token-var "HC_GITHUB_TOKEN"
        }
}
```
The way I think this would work, if a dependency plugin returns with a configuration error, we kick back to the user to add/update a deps {} section to configure the plugin. We should also disallow a plugin specified in `patch{}` and in `plugin{}`.

j-lanson:
Resolves #495 .

### DIFF
--- a/config/Hipcheck.kdl
+++ b/config/Hipcheck.kdl
@@ -8,7 +8,11 @@ plugins {
     plugin "mitre/entropy" version="0.1.0"
     plugin "mitre/churn" version="0.1.0"
 }
-
+patch {
+	plugin "mitre/github_api" {
+		api-token-var "HC_GITHUB_TOKEN"
+	}
+}
 analyze {
     investigate policy="(gt 0.5 $)"
     investigate-if-fail "mitre/typo" "mitre/binary"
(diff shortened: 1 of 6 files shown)


mchernicoff:
Should we have some kind of error if the user erroneously puts config information the patch section when it belongs to one of the plugins in the plugin section?

j-lanson:
> Should we have some kind of error if the user erroneously puts config information the patch section when it belongs to one of the plugins in the plugin section?

Yes, definitely. But having looked at the way things are written currently, I think the `PolicyFile` parsing needs a substantial overhaul anyway. Currently, we look through the whole analysis tree for each config. We should be maintaining a map of plugin --> config that synthesized from the `analysis` and `patch` sections, which would make detecting the scenario you described a lot easier. I'd rather wait to add the sanity check until we've done this refactor.

mchernicoff (approved):

PR #498, opened 2024-10-14
feat: Add activity plugin.

### LINKED ISSUE #485: Activity

mchernicoff:
Resolves #485 

Adds activity plugin, mostly by refactoring existing code from `analysis.rs` and `metric\activity.rs`.

Changes git plugin to use types from the Rust SDK.

Changes the Rust SDK query and mock query functions to accept inputs that `impl Serialize` instead of `Into<JsonValue>`.


plugins/activity/src/main.rs
```diff
@@ -0,0 +1,131 @@
...
+	const PUBLISHER: &'static str = "mitre";
+
+	const NAME: &'static str = "activity";
+
+	fn set_config(&self, _config: Value) -> std::result::Result<(), ConfigError> {
+		Ok(())
+	}
+
+	fn default_policy_expr(&self) -> hipcheck_sdk::prelude::Result<String> {
+		Ok("".to_owned())
```
j-lanson, line 59: We should implement a default policy expression. Usually its "less than X weeks". I'd like the plugin to optionally take a config field that provides the value for `X`. If it is supplied in `set_config`, then we return the corresponding default policy expr, otherwise we return `Ok("")`

plugins/activity/src/main.rs
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	fn explain_default_query(&self) -> hipcheck_sdk::prelude::Result<Option<String>> {
+		Ok(None)
```
j-lanson, line 63: Please add an explanation

j-lanson (changes requested):
A few changes, but overall looks very good!

j-lanson (approved):

PR #500, opened 2024-10-15
feat: add fuzz plugin

### LINKED ISSUE #482: Fuzz

j-lanson:
Resolves #482 .

This "analysis" is pretty simple, just forwards the "fuzz" query response from the `github_api` plugin.

I updated the SDK `mock.rs` to expect a response type of `Result<impl Serialize>` instead of `Result<Into<Value>>` in order to operate with the same expectations as the "query value" field.

### DIFF
--- a/plugins/fuzz/src/main.rs
+++ b/plugins/fuzz/src/main.rs
@@ -1,0 +1,97 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use clap::Parser;
+use hipcheck_sdk::{prelude::*, types::Target};
+use serde_json::Value;
+use std::result::Result as StdResult;
+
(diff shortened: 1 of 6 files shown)


plugins/fuzz/src/main.rs
```diff
@@ -0,0 +1,99 @@
...
+			package: None,
+		}
+	}
+
+	fn mock_responses() -> StdResult<MockResponses, Error> {
+		let target = target();
+		let known_remote = target.remote.as_ref().unwrap().clone();
+		let output = true;
+
+		// when calling into query, the input repo gets passed to `last_commit_date`, lets assume it returns the datetime `output`
```
mchernicoff, line 85: Comment should be updated for this plugin.

mchernicoff (approved):

PR #505, opened 2024-10-18
allow manifest field of plugins to be a path

patrickjcasey:
This pull request accomplishes 2 main goals:

1. It centralizes the location where all of the LEGACY "plugins" are managed, making it easier in the future to update the codebase to reflect when these features have been updated to use the plugin APIs
2. Adds the ability to run locally built plugins via a provided filepath, allowing for testing plugins, without the need for relying on downloading plugins and their manifests. This is especially useful right now, as the infrastructure for downloading plugins is not currently in place.

This pull request also cleans up a couple of small items in the codebase:

1. The `activity` plugin removed `mock_engine` from its `dependencies` and added it to its `dev-dependencies`, as it is only a dependency for testing and should not be running in the context of a normal `activity` plugin binary
2. Cleaned up some log messages to more accurately reflect the current state of the codebase 
3. Adds some missing SPDX license identifiers

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2577,6 +2577,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "review_sdk"
+version = "0.1.0"
+dependencies = [
+ "clap",
+ "hipcheck-sdk",
+ "log",
+ "schemars",
+ "serde",
+ "serde_json",
+ "tokio",
+ "url",
+]
+
+[[package]]
 name = "ring"
 version = "0.17.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ members = [
     "plugins/fuzz",
     "plugins/entropy",
     "plugins/linguist",
+    "plugins/review"
 ]
 
 # Make sure Hipcheck is run with `cargo run`.
--- a/plugins/github_api/plugin.kdl
+++ b/plugins/github_api/plugin.kdl
@@ -1,0 +1,10 @@
+publisher "mitre"
+name "github_api"
+version "0.1.0"
+license "Apache-2.0"
+entrypoint {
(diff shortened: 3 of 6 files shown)


hipcheck/src/analysis/score.rs
```diff
@@ -229,18 +217,16 @@ fn wrapped_query(
...
 			return Err(hc_error!("legacy analyses only have a default query"));
 		}
 		match plugin.as_str() {
-			ACTIVITY_PHASE => db.activity_analysis(),
 			AFFILIATION_PHASE => db.affiliation_analysis(),
 			BINARY_PHASE => db.binary_analysis(),
 			CHURN_PHASE => db.churn_analysis(),
 			ENTROPY_PHASE => db.entropy_analysis(),
 			IDENTITY_PHASE => db.identity_analysis(),
-			FUZZ_PHASE => db.fuzz_analysis(),
```
j-lanson, line 243: I would actually restore these two deleted lines from the match statement. the above `&& MITRE ...` does the check.

j-lanson:
Have you run an analysis including `fuzz` and `activity` using this version of the codebase?

patrickjcasey:
I have not tried to run an analysis with `fuzz` and `activity` enabled

patrickjcasey:
Here is the output from running the following command `cargo run -- check --policy Hipcheck.kdl https://github.com/mitre/hipcheck.git`

Here is the `Hipcheck.kdl` used:
cargo run -- check --policy Hipcheck.kdl https://github.com/mitre/hipcheck.git

Both `fuzz` and `activity` were enabled for this test and had no issues

j-lanson (changes requested):
See comment

patrickjcasey:
added them to the repo!

j-lanson:
legacy

On my side the analysis completed even with the missing files, and my print line shows its still using the legacy analysis function instead of the plugin.

j-lanson:
Please make it so that a missing manifest file results in a failed analysis, and double-check that you are running a plugin and not a legacy analysis function

PR #516, opened 2024-10-22
feat: Adds affiliation plugin.

mchernicoff:
Resolves issue #499 

Adds an affiliation plugin. While based on the existing code in `hipcheck/src/metric/affiliation.rs` and `hipcheck/src/analysis/mod.rs`, we redesigned the affiliation algorithm slightly to give a better count of affiliated contributors (e.g. not treating "Both" as its own contributor).

This plugin includes a parser for Organization Specification KDL files, to be included as config information when running this plugin. A short example of such a file is included as `example_orgs.kdl`. A future PR may add a larger file, similar to the existing `Orgs.toml` file.


plugins/affiliation/src/main.rs
```diff
@@ -0,0 +1,573 @@
...
+			)
+			.unwrap();
+
+		Ok(mock_responses)
+	}
+
+	#[test(tokio::test)]
+	async fn test_affiliation() {
+		let conf = Config {
+			orgs_file: pathbuf![&env::current_dir().unwrap(), "src", "example_orgs.kdl"],
```
j-lanson, line 552: Should make a separate dir `test` on the same level as `src` and move the example orgs KDL there

plugins/affiliation/src/main.rs
```diff
@@ -0,0 +1,573 @@
...
+	collections::{HashMap, HashSet},
+	fmt::{self, Display, Formatter},
+	path::PathBuf,
+	result::Result as StdResult,
+	sync::OnceLock,
+};
+
+#[derive(Debug, Deserialize)]
+struct Config {
+	orgs_file: PathBuf,
```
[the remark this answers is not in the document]
mchernicoff: Can do. I had thought about that but wasn't sure if the parsing should go in the main query or not.

plugins/affiliation/src/main.rs
```diff
@@ -0,0 +1,573 @@
...
+	}
+
+	fn default_policy_expr(&self) -> Result<String> {
+		match CONFIG
+			.get()
+			.ok_or_else(|| {
+				log::error!("tried to access config before set by Hipcheck core!");
+				Error::UnspecifiedQueryState
+			})?
+			.count_threshold
```
j-lanson, line 372: On my side clippy marks this:

```
warning: in a `match` scrutinee, avoid complex blocks or closures with blocks; instead, move the block or closure higher and bind it with a `let`
   --> plugins/affiliation/src/main.rs:368:19
    |
368 |               .ok_or_else(|| {
    |  ____________________________^
369 | |                 log::error!("tried to access config before set by Hipcheck core!");
370 | |                 Error::UnspecifiedQueryState
371 | |             })?
    | |_____________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#blocks_in_conditions
    = note: `#[warn(clippy::blocks_in_conditions)]` on by default
```

j-lanson (changes requested):
Few changes :) looks good though

plugins/affiliation/src/main.rs
mchernicoff: Still not marking it on mine. Explains why you had to fix it for the other plugin when you did your latest MR.

patrickjcasey: @mchernicoff maybe do a `rustup update`, there was recently a new version of Rust put out (1.82)

mchernicoff: Ah hah. I was working on Windows, not WSL like I usually do, to make sure we were OS filesystem agnostic, and I was still on 1.81 there.

j-lanson:
Please add a `plugin.kdl` file at `plugins/affiliation/`

j-lanson (approved):

PR #528, opened 2024-10-25
feat: improved plugin dependency resolution

patrickjcasey:
This pull request accomplishes 3 main goals:
1. It centralizes the location where all of the LEGACY "plugins" are managed, making it easier in the future to update the codebase to reflect when these features have been updated to use the plugin APIs
2.  Adds the ability to run locally built plugins via a provided filepath, allowing for testing plugins, without the need for relying on downloading plugins and their manifests. This is especially useful for running plugins that are actively being developed.
3. It centralizes all of the logic for retrieving plugins in `hipcheck/src/plugin/retrieval.rs`, both local and remote, and fixes an issue where plugin dependencies were not getting downloaded, nor started.

This pull request also cleans up a couple of small items in the codebase:
- The activity plugin removed mock_engine from its dependencies and added it to its dev-dependencies, as it is only a dependency for testing and should not be running in the context of a normal activity plugin binary
- Cleaned up some log messages to more accurately reflect the current state of the codebase
- Adds some missing SPDX license identifiers
- Fixed a couple of types and minor clippy lints

This pull request also sets the `fuzz` analysis to be run through the `plugin` infrastructure. There was a bug with `entropy` that was only present in the new `plugin` version of `entropy` `"not enough commits to calculate entropy"` was occurring and there is still work on converting the JSON pointer to a span. Once these two issues are resolved, I anticipate it being trivial to convert these analyses to use the `plugin` infrastructure


hipcheck/src/engine.rs
```diff
@@ -68,8 +64,12 @@ fn default_query_explanation(
 ) -> Result<Option<String>> {
 	let core = db.core();
 	let key = get_plugin_key(publisher.as_str(), plugin.as_str());
-	let Some(p_handle) = core.plugins.get(&plugin) else {
-		return Err(hc_error!("No such plugin {}", key));
+	let Some(p_handle) = core.plugins.get(&key) else {
+		return Err(hc_error!(
+			"No such plugin {}, {:?}",
+			key,
+			core.plugins.keys().collect::<Vec<&String>>()
```
j-lanson, line 71: Do we really want to be listing all of the running plugins in the error message? This seems more like something for `log::debug()`

hipcheck/src/plugin/download_manifest.rs
```diff
@@ -277,95 +264,6 @@ impl ParseKdlNode for DownloadManifestEntry {
...
 }
 
-impl DownloadManifestEntry {
-	/// This function does the following:
-	/// 1. Download specified plugin
-	/// 1. Verify its size and hash
-	/// 1. Extract plugin into plugin-specific folder
-	/// 1. Finds `plugin.kdl` inside plugin-specific folder and calls this recursively
-	pub fn download_and_unpack_plugin<'a>(
-		&self,
```
j-lanson, line 287: Do I understand correctly that all the deleted code in this file is superceded by the additions in `retrieval.rs`?

hipcheck/src/engine.rs
patrickjcasey: That is left over debugging code. I will update that tomorrow!

hipcheck/src/plugin/download_manifest.rs
patrickjcasey: Yes all of the logic has moved to `retrieval.rs`

hipcheck/src/engine.rs
j-lanson: I'd like to get this merged in today, could you just remove it quickly? Otherwise I'll merge the branch as-is and you can submit a fix as a separate PR.

patrickjcasey: I can push something in a couple of minutes!

j-lanson (approved):
Looks good, I'll merge when CI passes

PR #540, opened 2024-10-28
chore: update analysis config to allow non-string values

### LINKED ISSUE #539: Validate Identity plugin execution

j-lanson:
Resolves #539 .

Ran a simplified analysis with only identity analysis enabled. 

- The config for `identity` is `percent-threshold` which should be a float, but we only allow Strings currently. I updated the policy file definition to treat analysis config as `HashMap<String, Value>` instead. 
- Report building was still trying to legacy stuff for identity analysis output manipulation (expecting a float instead of an array), updated `get_legacy()` to check the `MITRE_LEGACY_PLUGINS` array so "plugin-ified" analyses will use the proper policy-expr handling code.


patrickjcasey:
Looks good to me!

patrickjcasey (approved):

PR #541, opened 2024-10-28
feat: Support Spans and DateTimes in JSON Pointer injection

### LINKED ISSUE #524: Add support for Spans/Datetimes in JSON pointer injection
We've added two new primitive types `Span` and `Datetime` to the policy expression language, each with a string representation unambiguous from the other.

However, the JSON pointer injection code that that inserts analysis data in policy expressions doesn't know how to detect these primitives and turn them into proper `enum Primitive` variants. 

Update the `Value --> Expr` parsing to handle these cases, either by detecting from string content, or as an extension, updating `Span` and `Datetime` schema in the SDK to emit as a "format"-specified string, and doing custom parsing from Value on the Hipcheck core side.

cstepanian:
Resolves #524

This PR adds initial support for reading type-annotated data from JSON objects returned by a Plugin.

Two new data types are supported: `jiff::Zoned` and `jiff::Span`.

This implementation expects the JSON to follow a particular format: the data will consist of a JSON object with two required fields, `format` and `data`. `format` must contain a string that references a known format; this PR adds support for `"date-time"` and `"duration"`. These names are adopted to match JSON Schema's naming. `data` must contain the data; for both of these added data types, they are represented as JSON strings.

References
 - https://json-schema.org/understanding-json-schema/reference/string#dates-and-times

### DIFF
--- a/hipcheck/src/policy_exprs/json_pointer.rs
+++ b/hipcheck/src/policy_exprs/json_pointer.rs
@@ -3,12 +3,13 @@
 #![deny(unused)]
 
 use crate::policy_exprs::{
-	error,
-	error::{Error, Result},
+	error::{self, Error, JiffError, Result},
 	expr::{Array, Expr, Primitive},
-	ExprMutator, JsonPointer,
+	ExprMutator, JsonPointer, LexingError,
 };
+use jiff::{tz::TimeZone, Span, Timestamp, Zoned};
 use ordered_float::NotNan;
+use serde::{Deserialize, Serialize};
 use serde_json::Value;
 
 /// Policy Expression stage that looks up JSON Pointers from the JSON `context`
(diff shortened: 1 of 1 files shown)


hipcheck/src/policy_exprs/json_pointer.rs
```diff
@@ -56,6 +57,46 @@ fn lookup_json_pointer<'val>(pointer: &str, context: &'val Value) -> Result<&'va
 	}
 }
 
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(tag = "format", content = "data")]
+#[serde(rename_all = "lowercase")]
+enum JSONData {
```
j-lanson, line 63: Maybe a more expressive name for this enum? It's specifically types that we have custom deserialization / expr-type detection for

j-lanson (dismissed):
Looks good, thanks Cal!

hipcheck/src/policy_exprs/json_pointer.rs
```diff
@@ -56,6 +57,46 @@ fn lookup_json_pointer<'val>(pointer: &str, context: &'val Value) -> Result<&'va
...
+	// Contains a string representing a jiff::Zoned
+	#[serde(rename = "date-time")]
```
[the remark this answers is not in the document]
cstepanian: Yes: the intention is to match up with the naming conventions used by JSON Schema. `date-time` and `duration` are both established values for `format` when defining a schema.

hipcheck/src/policy_exprs/json_pointer.rs
cstepanian: I agree that the name `JSONData` is too generic. But I couldn't think of a better name at the time.

Have any ideas?

How about `AnnotatedJSONData`, `TaggedJSONData`, `TaggedValue`, or some variation thereof?

cstepanian: I'm thinking `AnnotatedJSONValue`. While these two types are both represented as `String`s, and one can imagine most specially-handled data would be as well, it seems plausible to me that other JSON representations could be used.

j-lanson: Fine with me, please update it and I'll approve and merge

cstepanian: Renamed to `AnnotatedJSONValue`, and added detailed comments about the formats and relevant specs.

j-lanson (approved):

PR #550, opened 2024-10-30
feat: add RFD describing proposed batched query feature

### LINKED ISSUE #502: Investigate and write RFD for batching plugin queries
Sometimes a plugin will want to make a query for each element in a collection. In some cases, this has the potential to be very costly from a performance standpoint, as each request is a separate gRPC call. We should augment the Hipcheck protobuf definition to include a message type for a "batched" request, so Hipcheck knows to unload each element from a vector and make the `query()` for each one individually.

Investigate how this would work, and prepare an RFD describing the proposed change.

j-lanson:
Resolves #502 .

### DIFF
--- a/site/content/docs/rfds/0009-batch-queries.md
+++ b/site/content/docs/rfds/0009-batch-queries.md
@@ -1,0 +1,139 @@
+---
+title: Batched Plugin Queries
+weight: 9
+slug: 0009
+extra:
+  rfd: 9
+---
+
+# Batched Plugin Queries
+
+Now that we've made substantial progress in converting legacy Hipcheck analysis
+implementations to plugins, we've identified the need for a way to "batch"
+queries for performance reasons. For example, the entropy and churn analyses
+rely on a "linguist" functionality that evaluates whether a given file in a repo
+contains source code. The plugin design philosophy would dictate that we
+separate the `linguist` functionality into its own plugin, with an endpoint
+`is_likely_source_file(&Path) -> bool`, and have churn and entropy each query
+that. But making a `gRPC` request from the `entropy` plugin for each file in a
+potentially large repository is likely to incur runtime costs.
+
+Plugins can expose query endpoints that take a `Vec<_>` of objects and return a
+`Vec<_>` of reponses, but because `salsa` memo-ization operates on the entire
+query object, it would cache the entire `Vec<_>` as a key, and therefore any
+future `Vec`-based queries would not benefit from the memo-ization unless they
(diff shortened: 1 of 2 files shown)


alilleybrinker:
Thanks for this! Two comments:

- On the SDK changes, I think it should be possible to still have a single `query` function which uses a new trait to permit calling it with some a singular `&'static str` and `&[&'static str]` (or anything that derefs to that).
- Could the submit chunking actually be split into a separate RFD? It can be a small one. I'd view this as an oversight in the original design, I just think putting it here may make it hard to find later.

alilleybrinker (changes requested):
(Sorry, forgot to give my comments as a review: see my comments above)

j-lanson:
> * On the SDK changes, I think it should be possible to still have a single `query` function which uses a new trait to permit calling it with some a singular `&'static str` and `&[&'static str]` (or anything that derefs to that).

Can you elaborate on this point? As I understand, you're suggesting not having an explicit `batch_query()` function. I think interpreting whether a user wants to batch or not dependent on the type of the passed variable is wrong, because there may be some query endpoints that still take a `Vec` of objects. `query(Vec<Value>)` means "send this as a single request to the endpoint", `batch_query(Vec<Value>)` means "I don't want to waste time sending `hc` core each value individually, but they each represent a different query to X plugin"

alilleybrinker:
> > * On the SDK changes, I think it should be possible to still have a single `query` function which uses a new trait to permit calling it with some a singular `&'static str` and `&[&'static str]` (or anything that derefs to that).
> 
> Can you elaborate on this point? As I understand, you're suggesting not having an explicit `batch_query()` function. I think interpreting whether a user wants to batch or not dependent on the type of the passed variable is wrong, because there may be some query endpoints that still take a `Vec` of objects. `query(Vec<Value>)` means "send this as a single request to the endpoint", `batch_query(Vec<Value>)` means "I don't want to waste time sending `hc` core each value individually, but they each represent a different query to X plugin"

Ah, you're right. Another thought though has occurred to me. It may be useful to, in addition to a singular `batch_query` method, also expose a `batch()` method which returns a handle with its own `query()` method, with the handle being consumed and the query sent with a `send()` (or some similar name) final method.

j-lanson:
I added to the proposed `PluginEngine` interface changes to include the `batch()` function, and split the Submit chunking proposal into its own RFD

alilleybrinker (approved):
LGTM! :tada:

PR #555, opened 2024-10-31
feat: improve experience when using local plugins

patrickjcasey:
This change set improves the ergonomics of run 'local' plugins. Rather than running them at the path reachable in a provided Policy file, hipcheck now copies the entrypoints and plugin manifest files to the same cache locations that downloaded plugins would use.

Here is what `tree ~/Library/Caches/hipcheck/plugins` looks like on macOS after running hipcheck:

image

I deleted everything in `~/Library/Caches/hipcheck/plugins` before this run

I used the config on this branch in `config/Hipcheck.kdl` for testing this branch

This branch also added manifest paths to `binary` and `churn` to `config/Hipcheck.kdl` and updated the policy expression for `review`, in order to allow `hipcheck` to successfully complete its analysis with the config in the repo

### DIFF
--- a/config/Hipcheck.kdl
+++ b/config/Hipcheck.kdl
@@ -1,12 +1,12 @@
 plugins {
     plugin "mitre/activity" version="0.1.0"
-    plugin "mitre/binary" version="0.1.0"
+    plugin "mitre/binary" version="0.1.0" manifest="./plugins/binary/plugin.kdl"
     plugin "mitre/fuzz" version="0.1.0" manifest="./plugins/fuzz/plugin.kdl"
     plugin "mitre/review" version="0.1.0" manifest="./plugins/review/plugin.kdl"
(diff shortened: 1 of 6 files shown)


config/Hipcheck.kdl
```diff
@@ -24,7 +24,7 @@ analyze {
 			binary-file-threshold 0
 		}
         analysis "mitre/fuzz" policy="(eq #t $)"
-        analysis "mitre/review" policy="(lte $ 0.05)"
+        analysis "mitre/review" policy="(lte (divz (count (filter (eq #f) $)) (count $)) 0.05)"
```
j-lanson, line 27: can replace this with 
```
analysis "mitre/review" {
    percent-threshold 0.05
}
```

j-lanson (approved):

Issue #545, opened 2024-10-30
Validate Activity plugin execution


j-lanson:

j-lanson:
This the tracking-issue version of the issue #543 you opened

PR #556, opened 2024-10-31
feat: added support for running "mitre/activity" as a plugin

patrickjcasey:
This pull requrest adds support for running `mitre/activity` via the new plugin infrastructure and tweaks the default config `config/Hipcheck.kdl` to enable running this plugin via the default config

During the development of this feature, I hit a couple of `unreachable!()` blocks in the code that evaluates policy expressions, so I added a new error type `MismatchedTypes(Box<Primitive>, Box<Primitive>, String)` to `hipcheck/src/policy_exprs/errors.rs` to make it clearer when mismatched types are compared

To make it easier for users to work with timestamp data, specifically when working with `jiff::Span` and `jiff::Zoned` and will allow us to more easily enable custom JSON objects that are opaque to the user in the future

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1430,6 +1430,7 @@ dependencies = [
  "futures",
  "hipcheck-sdk-macros",
  "indexmap 2.6.0",
+ "jiff",
  "log",
  "prost",
  "rand",
--- a/config/Hipcheck.kdl
+++ b/config/Hipcheck.kdl
@@ -1,5 +1,5 @@
 plugins {
-    plugin "mitre/activity" version="0.1.0"
+    plugin "mitre/activity" version="0.1.0" manifest="./plugins/activity/plugin.kdl"
     plugin "mitre/binary" version="0.1.0" manifest="./plugins/binary/plugin.kdl"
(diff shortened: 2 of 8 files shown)


hipcheck/src/policy_exprs/env.rs
```diff
@@ -347,7 +347,11 @@ fn gt(env: &Env, args: &[Expr]) -> Result<Expr> {
 				.map_err(|err| Error::Datetime(err.to_string()))?
 				.is_gt(),
 		)),
-		_ => unreachable!(),
+		(a, b) => Err(Error::MismatchedTypes(
+			Box::new(a),
+			Box::new(b),
+			name.to_string(),
+		)),
```
j-lanson, line 354: I think I would suggest reverting all of these changes associated with the `MismatchedTypes` variant, and remove the variant. The PR #514 will make this change obsolete, and merging it now will just cause conflicts with that PR.

j-lanson (changes requested):
Looks good, just one change because of broader project context

j-lanson:
Resolves #545 .

patrickjcasey:
Should be good to go!

j-lanson (approved):

PR #557, opened 2024-11-01
feat(site): Substantial site redesign

alilleybrinker:
This redesign is still a draft with pieces unfinished, but it's at the point where we can probably start to have others contribute to it.


alilleybrinker:
Run `cargo xtask site serve` locally to see what this all looks like.

alilleybrinker:
One meaningful change in here for functionality: this updates the default policy expressions for all top-level plugins to apply defaults for optional values rather than not having a default policy expression at all if config items are omitted. This should match the behavior prior to the transition to the plugin system.

PR #558, opened 2024-11-01
chore: Makes changes needed for affiliation plugin to run

### LINKED ISSUE #549: Validate Affiliation plugin execution

mchernicoff:
Resolves #549 

Updates affiliation plugin to run properly.

Because we have not yet implemented batching for queries, this PR adds temporary "local" functions to the git plugin, which do the same thing as existing async query functions but are not queries. It also adds new "batch" queries that call these local functions instead of relying on nested queries and can process more than one commit or contributor at once.

The code in the affiliation plugin has been updated to use these temporary queries instead of making looped calls to the regular queries. Once batching has been implemented, we can remove the temporary functions and change the execution code for affiliation back.

### DIFF
--- a/plugins/affiliation/src/main.rs
+++ b/plugins/affiliation/src/main.rs
@@ -27,53 +27,54 @@ use std::{
 	sync::OnceLock,
 };
 
+pub static ORGSSPEC: OnceLock<OrgSpec> = OnceLock::new();
+
 #[derive(Debug, Deserialize)]
 struct Config {
 	orgs_spec: OrgSpec,
-
 	// Maximum number of concerningly affilaited contributors permitted in a default query
-	count_threshold: Option<u16>,
+	count_threshold: Option<u64>,
 }
 
-#[derive(Deserialize)]
+#[derive(Debug, Deserialize)]
 struct RawConfig {
-	orgs_file_var: Option<String>,
-	count_threshold_var: Option<u16>,
+	#[serde(rename = "orgs-file-path")]
+	orgs_file_path: Option<String>,
+	#[serde(rename = "count-threshold")]
+	count_threshold: Option<u64>,
 }
 
 impl TryFrom<RawConfig> for Config {
 	type Error = ConfigError;
 	fn try_from(value: RawConfig) -> StdResult<Config, ConfigError> {
-		if let Some(ofv) = value.orgs_file_var {
+		if let Some(ofv) = value.orgs_file_path {
 			// Get the Orgs file path and confirm it exists
 			let orgs_file = PathBuf::from(&ofv);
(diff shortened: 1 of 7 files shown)


plugins/affiliation/src/main.rs
```diff
@@ -206,113 +221,131 @@ impl AffiliatedType {
...
-		};
-		let view_value = engine
-			.query("mitre/git/contributors_for_commit", detailed_repo)
-			.await
-			.map_err(|e| {
-				log::error!("failed to get contributors for commit {}: {}", hash, e);
-				Error::UnspecifiedQueryState
-			})?;
-		let commit_view: CommitContributorView = serde_json::from_value(view_value)
-			.map_err(|_| Error::UnexpectedPluginQueryInputFormat)?;
```
j-lanson, line 258: Leave this in as a comment block perhaps so the switch back to looping is easy?

plugins/affiliation/src/main.rs
```diff
@@ -206,113 +221,131 @@ impl AffiliatedType {
...
-		};
-		let view_value = engine
-			.query("mitre/git/contributors_for_commit", commit_repo)
-			.await
-			.map_err(|e| {
-				log::error!("failed to get contributors for commit {}: {}", hash, e);
-				Error::UnspecifiedQueryState
-			})?;
-		let commit_view: CommitContributorView = serde_json::from_value(view_value)
-			.map_err(|_| Error::UnexpectedPluginQueryInputFormat)?;
```
j-lanson, line 290: Same comment as above

plugins/affiliation/src/main.rs
```diff
@@ -322,64 +355,56 @@ async fn affiliation(engine: &mut PluginEngine, key: Target) -> Result<i64> {
 
 	log::info!("completed affiliation metric");
 
-	Ok(count)
+	Ok(affiliations)
```
j-lanson, line 358: It's hard to tell with the logic, but it doesn't appear that this function returns 1 bool per commit the way that its header comment says it does. Can you verify that `affiliations.len() == hashes.len()`?

j-lanson (changes requested):

plugins/affiliation/src/main.rs
mchernicoff: I was thinking of doing that. Will do.

plugins/affiliation/src/main.rs
mchernicoff: That would be because I forgot to update the header. Affiliations should have 1 bool per contributor, not commit. Testing confirms that. Will update the header comment.

j-lanson (approved):
Thanks Michael!

Issue #568, opened 2024-11-05
Rename `github_api` and `npm_dependencies` plugins


j-lanson:
`github_api` --> `github`,
`npm_dependencies` --> `npm`

Don't forget any of the `plugin.kdl` files that rely on this, please make sure the analysis runs.

PR #571, opened 2024-11-05
rename github and npm plugins

patrickjcasey:
This branch renames 2 of the current `mitre` plugins and has been broken into a commit for each of the plugins that was renamed. 

I also noticed the main struct used in `mitre/github` was called `GitlabAPIPlugin`, when it should really be `GithubAPIPlugin` to be less confusing


patrickjcasey:
Resolves #568

j-lanson (approved):
Looks good, thanks Patrick!

PR #573, opened 2024-11-05
chore: moved dummy plugins into test-plugins folder

### LINKED ISSUE #569: Move dummy plugins to a `test` dir
Make sure that `hc plugin` still works as a subcommand.

patrickjcasey:
This branch moves `plugin/dummy-*` to `test-plugins`

`hc plugin` still works with these changes

resolves #569


Cargo.toml
```diff
@@ -15,10 +15,10 @@ members = [
...
     "hipcheck-macros",
     "xtask",
-    "plugins/dummy_rand_data_sdk",
-    "plugins/dummy_rand_data",
-    "plugins/dummy_sha256",
-    "plugins/dummy_sha256_sdk",
+    "test-plugins/dummy_rand_data_sdk",
+    "test-plugins/dummy_rand_data",
+    "test-plugins/dummy_sha256",
+    "test-plugins/dummy_sha256_sdk",
```
j-lanson, line 21: I know this wasn't explicit in the issue, but while you're at it can you delete the non-SDK dummy plugins and remove the `--sdk` flag from the `plugin` subcommand CLI and logic?

j-lanson (changes requested):
Looks good, just requesting additional changes not part of original issue.

Cargo.toml
patrickjcasey: No problem!

patrickjcasey:
Removed the non-sdk dummy plugins and removed the -sdk suffix from the sdk-compatible plugins and removed the `--sdk` flag from `hc plugin`

j-lanson (approved):
Looks good!

Issue #570, opened 2024-11-05
`git` diff parser can't handle hyphens in `stats` section


j-lanson:
Sometimes when a non-source file is added, the resulting git diff commant will emit this:
```
0       4       site/content/_index.md
136     2       site/content/install/_index.md
-       -       site/static/images/homepage-bg.png
2       2       site/tailwind.config.js
```
but our stats parsing code expects the two first columns to always contain numbers.

PR #577, opened 2024-11-05
Diff parser bugfixes

### LINKED ISSUE #574: `git` parser can't handle patchless diff
In one of the commits in this repo, there is a diff that looks like this:
```
diff --git a/hipcheck/src/analysis/session/spdx.rs b/hipcheck/src/session/spdx.rs
similarity index 100%
rename from hipcheck/src/analysis/session/spdx.rs
rename to hipcheck/src/session/spdx.rs
```
This is all metadata, and our parser expects that afterwards there will be a line starting with `@@` to denote where the diff "content" occurs. However, this is a rename-only change, so there is no content to emit.

alilleybrinker:
Fixes #574, #570

### DIFF
--- a/plugins/git/src/main.rs
+++ b/plugins/git/src/main.rs
@@ -458,7 +458,7 @@ mod test {
 
 	#[test]
 	fn test_hyphens_in_diff_stats() {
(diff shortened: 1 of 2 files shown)


alilleybrinker:
Oh let me collapse Julian's test case commits down to one.

j-lanson (approved):
Thanks Andrew!

PR #589, opened 2024-11-07
fix: Fix injection logic for datetime and span

alilleybrinker:
The previous implementation of JSON pointer injection for datetime and span data was incorrect. In the prior implementation, the datetime or span data needed to be provided as an object with two fields: `field` and `data`, where `field` specifies the type to convert the `data` field into. The correct implementation is to simply try converting into one, then the other, and to then fail if neither succeeds. They have completely distinct string representations, so there is not a possibility of confusable conversion into the *wrong* type with this approach.

In the future we can and should update this implementation to incorporate the JSON schema data on what type is intended. I believe confusion on the use of JSON schema here (which *does* use the `field`/`data` pattern described above, but in the _schema_ not the actual data) was the source of the original mistaken implementation.

### DIFF
--- a/hipcheck/src/policy_exprs/json_pointer.rs
+++ b/hipcheck/src/policy_exprs/json_pointer.rs
@@ -1,15 +1,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
-#![deny(unused)]
-
 use crate::policy_exprs::{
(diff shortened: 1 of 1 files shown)


j-lanson (approved):
Looks good!

PR #590, opened 2024-11-07
fix: backoff_interval adjusted for MacOS

aamohd:
The `backoff_interval` was causing failures to run on MacOS with a `Reached max spawn attempts for plugin...` error. This fix increases the `backoff_interval_micros` value to allow enough time for a connection to made for the plugin.

### DIFF
--- a/hipcheck/src/engine.rs
+++ b/hipcheck/src/engine.rs
@@ -219,7 +219,7 @@ pub fn start_plugins(
 		/* max_spawn_attempts */ 3,
 		/* max_conn_attempts */ 5,
 		/* port_range */ 40000..u16::MAX,
-		/* backoff_interval_micros */ 1000,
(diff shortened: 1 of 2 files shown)


hipcheck/src/plugin/manager.rs
```diff
@@ -107,7 +107,6 @@ impl PluginExecutor {
 				if !(0.0..=2.0).contains(&jitter_percent) {
 					panic!("Math error! We should have better guardrails around PluginExecutor field values.");
 				}
-				// sleep_duration = (backoff * conn_attempts) * (1.0 +/- jitter_percent)
```
j-lanson, line 110: Please restore this comment, it's documentation not dead code

j-lanson (changes requested):
One fix, otherwise looks good

j-lanson (approved):
Looks good!

PR #594, opened 2024-11-08
feat: Improve error reporting from typo and review

alilleybrinker:
This commit modifies the error reporting type in the SDK to enable more expressive errors from plugins, and updates 'typo' and 'review' to use it.


sdk/rust/src/error.rs
```diff
@@ -59,8 +59,30 @@ pub enum Error {
 
 	#[error("invalid format for QueryTarget")]
 	InvalidQueryTargetFormat,
+
+	#[error(transparent)]
+	Unspecified { source: DynError },
+}
```
j-lanson, line 65: Should `UnspecifiedQueryState` be removed from the enum in favor of this? Of course that would be done in a future PR

plugins/typo/src/main.rs
```diff
@@ -66,34 +67,25 @@ impl TryFrom<RawConfig> for Config {
...
 	// Get the repo's dependencies
 	let value = engine
-		.query("mitre/npm/dependencies", local)
+		.query("mitre/npm/dependencies", value.local)
 		.await
-		.map_err(|e| {
-			log::error!("failed to get dependencies for typo query: {}", e);
-			Error::UnspecifiedQueryState
-		})?;
+		.map_err(Error::any)?;
```
j-lanson, line 79: In the `review` plugin changes, you used `.context()` to explain "failed to get pull request reviews". Why not do the analogous thing here?

sdk/rust/src/error.rs
alilleybrinker: It could. I haven’t gone through and removed all of them, but that’s certainly the goal

plugins/typo/src/main.rs
alilleybrinker: Oh fair point. I can make that edit

j-lanson (approved):

PR #599, opened 2024-11-08
fix: bugs in detecting empty git diffs and end of patch meta block

j-lanson:
Resolves a few bugs in the `plugins/git` diff parser that I encountered while trying to run `plugins/entropy` on `numpy@no version`.

1. Our format specifier for `get_commits()` caused GPG key info to be printed and parsed, but locally I was getting an error in the GPG processing C code in `git`. Other team member got a different error, with git seemingly trying to do validation against GPG signatures in the git history when GPG was not set up on his machine.

2. Some diffs that just delete a file do not have the 1-2 lines prefixed with `+++` or `---` between the patch metadata and patch chunks. We were naively using `opt(line)` to capture these, which would eat up the start of another `diff` in some cases. Now we check for those lines explicitly. 

3. Some other commits (such as those just used to force the CI to re-run) have zero output in our customized `git log` invocation used for parsing diffs. Without any changes, we have no way to detect these commits, and our # parsed diffs != # num parsed commits. My solution was to add `~~~` to this argument `"--pretty=tformat:`, causing each diff to have a header of `~~~` printed whether or not there is anything else to print for that diff. This way, we can at least count empty diffs. Now we successfully parse all commit diffs for a large repo like `numpy` which has 28,000+ commits

### DIFF
--- a/plugins/git/src/main.rs
+++ b/plugins/git/src/main.rs
@@ -469,4 +469,12 @@ mod test {
 		let (leftover, _) = crate::parse::patch(input).unwrap();
 		assert!(leftover.is_empty());
 	}
+
+	#[test]
+	fn test_patch_without_triple_plus_minus() {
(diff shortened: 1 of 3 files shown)


alilleybrinker (approved):
LGTM!

PR #606, opened 2024-11-08
feat: improve error handling if plugin entrypoint does not exist

### LINKED ISSUE #583: Improve error message when a plugin isn't found by checking for presence first
Right now if a plugin can't be found to spawn, it errors out with a "max spawn attempts reached" error. We should improve that by using `which` to check for the plugin first and erroring out if we can't find it.

patrickjcasey:
Closes #583 

With the current codebase and plugins that we have, this check works, but if an entrypoint is not kicking off a binary and runs a docker image or runs `python3 -m ...` or does anything outside of how our plugins work today, this will likely fail. Maybe this should just log a warning that it could not find the provided file and still try to spawn the plugin?

### DIFF
--- a/hipcheck/src/plugin/manager.rs
+++ b/hipcheck/src/plugin/manager.rs
@@ -82,6 +82,15 @@ impl PluginExecutor {
 			// ports in the desired range are already bound
 			let port = self.get_available_port()?;
 			let port_str = port.to_string();
+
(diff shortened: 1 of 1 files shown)


alilleybrinker (changes requested):
I think we may be better off just warning for a likely error rather than hard-erroring here.

alilleybrinker (approved):

PR #608, opened 2024-11-08
chore: Preparing to use 'dist' for more releases.

alilleybrinker:
This relies on the upgrade to the latest version of 'dist'. Since we now have plugins that we want to produce pre-built binaries for, this updates our build configuration to support that. It'll need to be tested.


alilleybrinker:
Update: The `dist` folks are investigating this error; trying to figure out why `plan` is complaining, which it shouldn't be doing.

alilleybrinker:
After some debugging, I was able to at least eliminate the duplication of packages found by `cargo-dist`. Turns out my `workspace.members` field in the `dist-workspace.toml` file was wrong; when including a Cargo virtual workspace with many distable packages in it, you only need to include the root, not each package individually.

alilleybrinker:
Update: `cargo-dist` team is working on resolving the bug that's blocking this PR. Basically, an internal setting called `needs_coherence` is being incorrectly set to `true` for `cargo dist plan`. Once the fix is out I'll update this to use the patched version of `cargo-dist` and we can proceed.

alilleybrinker:
@j-lanson, this is now running a prerelease version of `cargo-dist` which incorporates the fixes for tag incoherence we've discussed. If all looks good to you, we can merge and then proceed with running a release for the new plugins!

plugins/activity/plugin.kdl
```diff
@@ -2,12 +2,14 @@ publisher "mitre"
...
+
 entrypoint {
-    on arch="aarch64-apple-darwin" "./target/debug/activity"
-    on arch="x86_64-apple-darwin" "./target/debug/activity"
-    on arch="x86_64-unknown-linux-gnu" "./target/debug/activity"
-    on arch="x86_64-pc-windows-msvc" "./target/debug/activity.exe"
+  on arch="aarch64-apple-darwin" "activity"
+  on arch="x86_64-apple-darwin" "activity"
+  on arch="x86_64-unknown-linux-gnu" "activity"
+  on arch="x86_64-pc-windows-msvc" "activity.exe"
```
j-lanson, line 10: The modification to the path is necessary for `download-manifest`-based plugin installs, but breaks the system we put in place for local plugin testing, as the plugin resolver relies on this path to know where to grab the locally-built plugin.

I think we do make the change, but we'll need to come up with a solution for supporting both ASAP

j-lanson (dismissed):

j-lanson (approved):

PR #624, opened 2024-11-14
chore: deleted activity analysis from hipcheck core

bmarr-mitre:
Addresses  https://github.com/mitre/hipcheck/issues/623


j-lanson (approved):
Looks good, thanks Brian

PR #626, opened 2024-11-14
fix: handling of plugin entrypoint string

j-lanson:
The existing code treated the `entrypoint` string as a `PathBuf`, when in reality it could be a cmdline string with spaces and arguments. 

This PR updates the handling to reflect this, and also modifies the `retrieve_local_plugin()` behavior to only copy the plugin entrypoint binary if it is a path. For instance, if the entrypoint started as `docker ...`, we wouldn't copy the `docker` binary because does not exist as a path. 

Also updated the warning in `start_plugin()` to warn if the binary component of `entrypoint` cannot be found with `which`.

### DIFF
--- a/hipcheck/src/plugin/manager.rs
+++ b/hipcheck/src/plugin/manager.rs
@@ -3,7 +3,7 @@
 use crate::{
 	hc_error,
 	hipcheck::plugin_service_client::PluginServiceClient,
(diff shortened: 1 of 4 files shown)


patrickjcasey (approved):
Looks good to me!

Issue #600, opened 2024-11-08
hc fails to start plugins on macOS


patrickjcasey:
![Image](https://github.com/user-attachments/assets/95c145d2-71cc-4b9b-bbf1-2c3ae6d05e75)

patrickjcasey:
This issue seems to reliably manifest itself when running `hc check` if another instance of `hc check` is running

alilleybrinker:
Hm. This is true for you even after the tuning we did yesterday for macOS spawn timing?

patrickjcasey:
Yes this is triggering with the latest version of `main`. To add more detail, the plugins are able to all bind to the ports handed to them by `hc`, but the function call  `PluginServiceClient::connect(format!("http://127.0.0.1:{port_str}")).await` in `hipcheck/src/plugin/manager.rs`  fails with a generic `transport error` from tonic

alilleybrinker:
Exciting! I think this is probably a race condition on automatic port assignment in macOS. We'll want to dig in further, and this _does_ evince the need to establish some sort of exclusion through locking the shared resources (repo cache, plugin cache) that Hipcheck has; alongside avoiding a data race on port assignment from the OS.

PR #628, opened 2024-11-14
fix: reduce unnecessary filesytem copies when spawning plugins

patrickjcasey:
Here is a video of this branch having two macOS `hc check` commands work at the same time!

https://github.com/user-attachments/assets/9c4a6355-8817-421e-85e4-00f903140946

To trigger this bug on macOS, the following steps were the easiest way I could find to reliably trigger this
1. Have a plugin running
2. While it is still running, copy over the running binary location on disk
3. Attempt to run this new binary (it will fail)

To avoid this issue, `hc` now checks the `sha256` of the source and destination before copying the files and skips the copy if the hashes are identical

### DIFF
--- a/hipcheck/src/plugin/retrieval.rs
+++ b/hipcheck/src/plugin/retrieval.rs
@@ -134,12 +134,19 @@ fn retrieve_local_plugin(
 			// unwrap is safe here, we just updated the entrypoint for current arch
 			let new_bin = try_get_bin_for_entrypoint(&new_entrypoint).0.unwrap();
 
+			// path where the binary for this plugin will get cached
+			let binary_cache_location = plugin_cache.plugin_download_dir(&plugin_id).join(new_bin);
+
 			// @Note - sneaky potential for unexpected behavior if we write local plugin manifest
 			// to a cache dir that already included a remote download
-			std::fs::copy(
-				&original_entrypoint,
-				plugin_cache.plugin_download_dir(&plugin_id).join(new_bin),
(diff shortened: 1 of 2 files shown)


j-lanson:
Resolves #600 .

hipcheck/src/plugin/retrieval.rs
```diff
@@ -134,12 +134,22 @@ fn retrieve_local_plugin(
...
+			// to only copy the binary to its new location if the hash of the original entrypoint is DIFFERENT than
+			// what is currently stored in the cache
+			if file_sha256(&original_entrypoint)?
+				!= file_sha256(&binary_cache_location).unwrap_or_default()
+			{
+				// @Note - sneaky potential for unexpected behavior if we write local plugin manifest
+				// to a cache dir that already included a remote download
+				std::fs::copy(
+					&original_entrypoint,
+					plugin_cache.plugin_download_dir(&plugin_id).join(new_bin),
```
j-lanson, line 150: re-use `&binary_cache_location` instead of recomputing

j-lanson (changes requested):
Looks good, just one small nit

patrickjcasey:
https://github.com/user-attachments/assets/3cd26b75-a576-4e9a-923a-e0642fe00fd8

`hc check` running simultaneously on the latest code in this branch

j-lanson (approved):
Looks good, thanks a bunch Patrick!

PR #650, opened 2024-11-22
chore: remove churn, entropy, typo from hipcheck core

### LINKED ISSUE #640: Delete legacy code for entropy plugin

### LINKED ISSUE #641: Delete legacy code for churn plugin

### LINKED ISSUE #642: Delete legacy code for typo plugin

bmarr-mitre:
Resolves #640
Resolves #641 
Resolves #642


j-lanson (changes requested):
PR is not passing CI checks, view the failing CI test, it will tell you what you need to change. You can also run `cargo clippy` locally to view changes, and something like `cargo clippy fix` to fix the changes automatically. but be careful because in my experience sometimes the reported changes on the CI is a larger set.

hipcheck/src/metric/math.rs
```diff
@@ -1,34 +1 @@
 // SPDX-License-Identifier: Apache-2.0
```
j-lanson, line 1: If the file now only contains the comment header, just delete the whole file and remove it as a module

j-lanson (approved):
Looks good!

PR #662, opened 2024-11-25
Aamohd/file based config

aamohd:


hipcheck/src/cli.rs
```diff
@@ -104,6 +105,16 @@ struct PathArgs {
...
 	)]
 	policy: Option<PathBuf>,
+
+	/// Path to the exec config file
+	#[arg(
+		short = 'e',
+		long = "exec",
+		global = true,
+		help_heading = "Path Flags",
+		long_help = "Path to the exec file."
```
j-lanson, line 115: --> `"Path to the execution config file"`

hipcheck/src/cli.rs
```diff
@@ -296,6 +315,9 @@ impl CliConfig {
 				policy: std::env::current_dir()
 					.ok()
 					.map(|dir| pathbuf![&dir, "Hipcheck.kdl"]),
+				exec: env::current_dir()
+					.ok()
+					.map(|dir| pathbuf![&dir, "config/Config.kdl"]),
```
j-lanson, line 320: Remove `config/`

hipcheck/src/executor.rs
j-lanson, line 1: Rename to `exec.rs`

hipcheck/src/main.rs
```diff
@@ -503,12 +506,24 @@ fn cmd_plugin(args: PluginArgs) {
...
+	let exec_config = if let Some(p) = config.exec() {
+		ExecConfig::from_file(p)
+		.context("Failed to load the exec config. Please make sure the exec config file is in the provided location and is formatted correctly.")
+	} else {
+		Err(hc_error!(
+			"No exec file found. Please provide an exec config file before running Hipcheck."
+		))
+	};
+
+	let plugin_data = exec_config.unwrap().plugin_data;
```
j-lanson, line 518: Instead of unwrapping, change the function to return an `ExitCode` and use `Shell::print_error()` like `cmd_setup()` and `cmd_check()`

hipcheck/src/plugin/types.rs
```diff
@@ -160,6 +160,9 @@ pub struct PluginContext {
 
 	/// The child process in which the plugin is running.
 	pub proc: Child,
+
+	/// The size of the gRPC buffer
+	pub grpc_buffer: usize,
```
j-lanson, line 165: rename to `grpc_query_buffer_size`

j-lanson (changes requested):

j-lanson:
As I mentioned in one of the requested changes, we shouldn't take`./config/Exec.kdl` as the backup location for the exec file, it should just be `Exec.kdl`. But in git version control we are storing the file at `./config/Exec.kdl`, which means people would need to pass `--exec` every time they want to run Hipcheck. 

I think the solution is that everyone developing on Hipcheck copies `./config/Exec.kdl` to `./Exec.kdl` at the project root so it gets picked up as the default location, and this will be everyone's personal `Exec.kdl` file that doesn't get tracked in git. We can also add a rule to the `.gitignore` to ensure no root-level `Exec.kdl` file gets commited. 

@alilleybrinker thoughts? or should the "config" folder be un-deprecated as a concept and use the resolution of the config dir to determine the expected location of the Exec.kdl file?

hipcheck/src/cli.rs
```diff
@@ -221,6 +231,12 @@ impl CliConfig {
 		self.deprecated_args.config.as_deref()
 	}
 
+	/// Get the path to the exec config file
+	pub fn exec(&self) -> Option<&Path> {
+		// Question - should I remove the path from backups
```
j-lanson, line 236: Remove comment

hipcheck/src/main.rs
```diff
@@ -503,12 +506,42 @@ fn cmd_plugin(args: PluginArgs) {
...
+				break;
+			}
+			if let Some(parent) = curr_dir.parent() {
+				curr_dir = parent.to_path_buf();
+			} else  {
+				log::info!("Using a default Exec Config");
+				config = ExecConfig::default();
+				break;
+			}
+		}
```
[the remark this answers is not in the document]
j-lanson: Also, can you add a `log::info()` line to report where the `Exec.kdl` gets resolved from or if it uses an in-memory one

hipcheck/src/main.rs
```diff
@@ -503,12 +506,42 @@ fn cmd_plugin(args: PluginArgs) {
...
-		/* max_spawn_attempts */ 3,
-		/* max_conn_attempts */ 5,
+		/* max_spawn_attempts */ plugin_data.max_spawn.attempts,
+		/* max_conn_attempts */ plugin_data.max_conn.attempts,
 		/* port_range */ 40000..u16::MAX,
-		/* backoff_interval_micros */ 100000,
-		/* jitter_percent */ 10,
+		/* backoff_interval_micros */ plugin_data.backoff.micros,
+		/* jitter_percent */ plugin_data.jitter.percent,
+		/*grpc_buffer*/ plugin_data.grpc_buffer.size,
```
j-lanson, line 544: Since all fields of `PluginExecutor` now come from the exec config file, I would also add a `ExecFile::get_plugin_executor() -> PluginExecutor ` function instead of creating it like this.

j-lanson (changes requested):
Just a couple more things, but looking good!

hipcheck/src/session/mod.rs
```diff
@@ -265,13 +291,45 @@ pub fn load_policy_and_data(policy_path: Option<&Path>) -> Result<(PolicyFile, P
...
+	// Resolve the path to the exec config file.
+	let exec_config = match exec_path {
+		Some(p) => {
+			// Use the path provided
+			if !p.exists() {
+				hc_error!("Failed to load exec config. Please make sure the path set by the --exec flag exists.");
+			}
+			ExecConfig::from_file(p)
+				.context("Failed to load the exec config. Please make sure the exec config file is in the provided location and is formatted correctly.")
+				.unwrap()
```
j-lanson, line 318: should pass up the error with `?` instead of `.unwrap()` here and below

hipcheck/src/exec.rs
```diff
@@ -289,6 +311,19 @@ impl ExecConfig {
 		}"#;
 		Self::from_str(data)
 	}
+
+	pub fn get_plugin_executor(plugin_data: PluginConfig) -> PluginExecutor {
```
j-lanson, line 315: this function should just take `&self` and use `self.plugin_data`. We already have a `PluginConfig` as one of the struct members.

hipcheck/src/exec.rs
```diff
@@ -289,6 +311,19 @@ impl ExecConfig {
...
+		// let plugin_data = exec_config.plugin_data;
```
j-lanson, line 316: Delete comment

hipcheck/src/session/mod.rs
```diff
@@ -224,14 +223,16 @@ impl Session {
...
+		let executor = ExecConfig::get_plugin_executor(plugin_data.clone());
+
+		// let executor = PluginExecutor::new(
+		// 	/* max_spawn_attempts */ plugin_data.max_spawn.attempts,
+		// 	/* max_conn_attempts */ plugin_data.max_conn.attempts,
+		// 	/* port_range */ 40000..u16::MAX,
+		// 	/* backoff_interval_micros */ plugin_data.backoff.micros,
+		// 	/* jitter_percent */ plugin_data.jitter.percent,
+		// 	/* grpc_buffer */ plugin_data.grpc_buffer.size,
+		// )?;
```
j-lanson, line 235: Delete comment

hipcheck/src/exec.rs
```diff
@@ -278,6 +279,27 @@ impl ExecConfig {
...
+				return Self::from_file(target_ref)
+			}
+			if let Some(parent) = curr_dir.parent() {
+				curr_dir = parent.to_path_buf();
+			} else  {
+				// If file not found, use default values
+				log::info!("Using a default Exec Config");
+				return Self::default();
+			}
+		}
```
j-lanson: Looking at it, this logic doesn't exactly do what is described here: https://doc.rust-lang.org/cargo/reference/config.html

We should first look in `env::current_dir()`, but when we look in ancestor directories we should look for `.hipcheck/Exec.kdl`

hipcheck/src/session/mod.rs
```diff
@@ -320,26 +321,9 @@ fn load_exec_config(exec_path: Option<&Path>) -> Result<ExecConfig> {
...
-					curr_dir = parent.to_path_buf();
-				} else  {
-					// If file not found, use default values
-					log::info!("Using a default Exec Config");
-					exec_config = ExecConfig::default().unwrap();
-					break;
-				}
-			}
+			exec_config = ExecConfig::find_file()
+				.context("Failed to load the exec config. Please make sure the exec config file is in the provided location and is formatted correctly.")
```
j-lanson: --> "Please ensure the `Exec.kdl` is in the current directory or in `.hipcheck/Exec.kdl` of a parent directory."

hipcheck/src/session/mod.rs
```diff
@@ -320,26 +321,9 @@ fn load_exec_config(exec_path: Option<&Path>) -> Result<ExecConfig> {
...
+				.unwrap();
```
j-lanson: `.unwrap()` --> `?`

hipcheck/src/exec.rs
```diff
@@ -289,6 +311,19 @@ impl ExecConfig {
...
+		// let plugin_data = exec_config.plugin_data;
+		PluginExecutor::new(
+			/* max_spawn_attempts */ plugin_data.max_spawn.attempts,
+			/* max_conn_attempts */ plugin_data.max_conn.attempts,
+			/* port_range */ 40000..u16::MAX,
+			/* backoff_interval_micros */ plugin_data.backoff.micros,
+			/* jitter_percent */ plugin_data.jitter.percent,
+			/*grpc_buffer*/ plugin_data.grpc_buffer.size,
+		)
+		.unwrap()
```
j-lanson, line 325: `.unwrap()` --> `?` and change function signature to return a `Result<PluginExecutor>`. In general we should not use unwrap unless we have a good reason why we know it won't fail.

hipcheck/src/main.rs
```diff
@@ -503,14 +506,17 @@ fn cmd_plugin(args: PluginArgs) {
...
-		/* jitter_percent */ 10,
-	)
-	.unwrap();
+	let exec_config = if let Some(p) = config.exec() {
+		ExecConfig::from_file(p)
+			.context("Failed to load the exec config. Please make sure the exec config file is in the provided location and is formatted correctly.")
+	} else {
+		ExecConfig::find_file()
+			.context("Failed to load the exec config. Please make sure the exec config file is in the provided location and is formatted correctly.")
+	}.unwrap();
```
j-lanson, line 515: Like other comments, replace `unwrap` with doing `Shell::print_error` on failure.

hipcheck/src/main.rs
```diff
@@ -503,14 +506,17 @@ fn cmd_plugin(args: PluginArgs) {
...
-		/* backoff_interval_micros */ 100000,
-		/* jitter_percent */ 10,
-	)
-	.unwrap();
+	let exec_config = if let Some(p) = config.exec() {
+		ExecConfig::from_file(p)
+			.context("Failed to load the exec config. Please make sure the exec config file is in the provided location and is formatted correctly.")
+	} else {
+		ExecConfig::find_file()
+			.context("Failed to load the exec config. Please make sure the exec config file is in the provided location and is formatted correctly.")
```
j-lanson, line 514: It doesn't make sense to use the exact same error message here. The message makes sense for the `from_file()` call because the user gave us a path to a file and we failed to load that. In `find_file()` it could just be that no file exists in the places we searched so "the provided location" is vague and inaccurate.

j-lanson (changes requested):
Don't forget to squash your commits into one to pass the CI check

j-lanson (changes requested):

hipcheck/src/main.rs
```diff
@@ -503,14 +506,28 @@ fn cmd_plugin(args: PluginArgs) {
...
+		},
+		Err(e) => {
+			Shell::print_error(&hc_error!("Failed to resolve the exec config {}", e), Format::Human);
+			return ExitCode::FAILURE;
+		}
+	};
+
+	let plugin_executor = ExecConfig::get_plugin_executor(&exec_config)
+		.context("Failed to resolve the Plugin Executor.")
+		.unwrap();
```
j-lanson, line 529: `.unwrap()` -> `Shell:print_error` and `return ExitCode::FAILURE`.

j-lanson (changes requested):
Please carefully check for any remaining erroneous uses of `unwrap()`.

config/Config.kdl
j-lanson, line 1: This file should be called `Exec.kdl`

j-lanson (changes requested):
One additional change (file rename), and previous comment has not been addressed yet.

j-lanson (changes requested):
Squash down to one commit and we're probably good to go

j-lanson (changes requested):
Can you please rebase off of main? Also, Rust 1.84 released today and our CI automatically uses it so you might see new clippy warnings in the CI that cause a test failure. You will need to fix those locally if they appear.

j-lanson (approved):

Issue #575, opened 2024-11-05
explore replacing `plugins/git` `nom` parser with `libgit2` function calls 


j-lanson:
To get commit diffs for a repo, our `git` plugin currently uses a configured `git log` message to get the entire diff history the repo, and uses a `nom` parser to extract data into `Diff` objects. The log data we parse does not include commit hashes, so if there are any corner cases, the parser will fail and `commits.len() != diffs.len()`. Rather than relying on the parser, we should see if we can use the Rust `libgit2` bindings to synthesize `CommitDiff` objects in a much more reliable way. 

[This link](https://libgit2.org/libgit2/ex/HEAD/log.html) shows example code in C of how to use the libgit2 API to get `git log` -like info. The task will be to see if we can do something similar to get the diff info we need to construct `CommitDiff` objects, and preferably remove the need to `zip` two independently-derived `Vec`s together.

PR #663, opened 2024-11-26
 feat: use gix for parsing git information in mitre/git plugin

patrickjcasey:
This pull request removes the need for a custom `nom`-based parser for parsing git information in the `mitre/git` plugin, by swapping over to using `gix` https://github.com/GitoxideLabs/gitoxide for parsing git information. As a part of this change, I got some data from the following repos:
- https://github.com/mitre/hipcheck
- https://github.com/numpy/numpy

testing on `https://github.com/linux/torvalds` was skipped for time reasons

I investigated using `libgit2` first, but the performance on larger repos was not ideal, and it was difficult to integrate caching into the code. `gix` (https://github.com/GitoxideLabs/gitoxide) has better support for caching and is written in pure Rust which is also nice for memory safety reasons.

for determining execution speed, the following command was run with the `release` version of `hc` on my M1 Macbook Pro with 32 GB RAM, the `mitre/git` plugin was also configured to run the `release` version:

`time ./target/release/hc check --policy config/Hipcheck.kdl <REPO>`

for testing `main`, commit `4f205afd6ba0afd98d0ba91dc742f2038f6ddc89` was used

|  Branch   | Repo | # of Commits | # of CommitDiffs | # of Contributors | Last Commit |Execution Speed |
|  :-------: |  :--------------: | :----------: | :--------: | :--------------: | :-----------: | :-------------: |
| `main`    | `mitre/hipcheck` |      529         |    529               |         20 |    2024-12-09T22:58:27Z      |  9.19 s  |
| `gix`  | `mitre/hipcheck` |       529         |    529               |         20 |    2024-12-09T22:58:27Z      |  9.98 s  |
| `main`    | `numpy/numpy` | **28436**  | **28436** |    **1898**| 2024-12-09T20:49:11Z| 2m 49s |
| `gix`  | `numpy/numpy` |       37617       |       37617         |      2243   |    2024-12-09T20:49:11Z  |  4m 38s    |

`git rev-list --count HEAD` results:
- `mitre/hipcheck`: 529 is returned, which matches both implementations
- `numpy/numpy`: 37617 is returned, which matches the `gix` implementation

`git log --pretty="%an %ae%n%cn %ce" | LC_ALL=C sort -u | wc -l` results:
- `mitre/hipcheck`: 20 is returned, which matches both implementations
- `numpy/numpy`: 2243 is returned, which matches `gix` implementation

The main highlight of this pull request is the fact that the parsing of git-related information is handled by a git library, rather than our `nom`-based parser, which was missing commits and contributors for larger repos. At some point, a pass can be made to dive into the performance of the `git` plugin to find places for optimization


patrickjcasey:
Closes #575

hipcheck-common/src/chunk.rs
```diff
@@ -223,9 +223,7 @@ mod test {
 		let res = match chunk_with_size(query, 10) {
 			Ok(r) => r,
 			Err(e) => {
-				println!("{e}");
-				assert!(false);
-				return;
+				panic!("{e}");
```
j-lanson, line 226: Is `panic!` really correct to use in test cases?

plugins/git/src/git.rs
```diff
@@ -0,0 +1,267 @@
...
+use gix::revision::walk::Sorting;
+use gix::revision::Walk;
+use gix::traverse::commit::simple::CommitTimeOrder;
+use gix::Repository;
+use jiff::Timestamp;
+use std::path::Path;
+use std::sync::OnceLock;
+
+// a lot of the data related to commits can be derived from a Vec<RawCommit>, so cache this and reuse it when possible
+static ALL_RAW_COMMITS: OnceLock<Result<Vec<RawCommit>>> = OnceLock::new();
```
j-lanson, line 24: Aren't all of these commits tied to a specific repo? The `git` plugin needs to be idempotent. Seems like if `get_all_raw_commits` is called on Repo A, then on Repo B, the execution for Repo B will just re-use the commits from Repo A.

j-lanson: Instead you probably want to store a static `(PathBuf, HeadCommit, Vec<RawCommit>)` and use that if possible, but invalidate and replace it when a new `repo` is requested.

hipcheck-common/src/chunk.rs
patrickjcasey: assert!(false) is raising a clippy warning. If there is an error and the test panics, the test harness will catch it and display the error to the user

plugins/git/src/git.rs
patrickjcasey: I can look at it, but I'm pretty sure OnceLock runs once, each time the binary is invoked. Since the `git` plugin is invoked every time `hc check` is run, it will get populated once for each repo. If it wasn't working, then the number of commits reported would be the same on two different repos, as `get_all_raw_commits` is called internally by multiple functions

j-lanson: That's under our current usage of the `git` plugin, but someone else might write and run an analysis plugin that doesn't know about this hidden constraint on the `git` plugin. We should design it to be an optimization, but not restrict how the different `git` plugin API endpoints get called.

patrickjcasey: Ok sounds good. I will fix this tomorrow

j-lanson (approved):
Thanks Patrick!

PR #676, opened 2024-12-02
`hc-common` refactor and Rust SDK benchmarking feature

### LINKED ISSUE #649: Refactor `PluginQuery` struct ecosystem and chunking functionality into `hipcheck_common` crate
To reduce code duplication, we can move the struct definitions and chunking algorithm into a crate with a name like `hipcheck_common` that makes clear it's not really for use by anything but Hipcheck.

j-lanson:
Resolves #649 .

This PR moves the `Query --> PluginQuery --> Query` conversion code and chunking algorithm / size estimation code into a common crate that both the Rust SDK and Hipcheck core can use. 

Also, for testing if my changes slowed down execution, I added `benchmarking.rs` from Hipcheck core to the Rust SDK and added a single usage to record how long a query takes, ignoring the time to send/receive its query messages to/from the core.


Cargo.toml
```diff
@@ -31,7 +31,7 @@ members = [
     "plugins/identity",
     "plugins/linguist",
     "plugins/review",
-    "plugins/typo"]
+    "plugins/typo", "hipcheck-common"]
```
mchernicoff, line 34: *Very* minor thing, but should this be on two lines for consistency?

mchernicoff (approved):

PR #680, opened 2024-12-03
chore: remove hc_github_token from hipcheck core

### LINKED ISSUE #673: Remove `HC_GITHUB_TOKEN` requirement from `hc` core
The Hipcheck core used to need users to define an `HC_GITHUB_TOKEN` env var. Now the github token is required by the `github` plugin, but the `hipcheck/` crate and commands like `hc ready` still look for this env var. Remove checks/references to `HC_GITHUB_TOKEN` from the Hipcheck core crate and documentation.

@alilleybrinker perhaps we should introduce a way for plugins to report if they are ready?

bmarr-mitre:
Resolves #673

### DIFF
--- a/hipcheck/src/main.rs
+++ b/hipcheck/src/main.rs
@@ -362,13 +362,10 @@ struct ReadyChecks {
 	npm_version_check: StdResult<String, VersionCheckError>,
 	cache_path_check: StdResult<PathBuf, PathCheckError>,
 	policy_path_check: StdResult<PathBuf, PathCheckError>,
-	github_token_check: StdResult<(), EnvVarCheckError>,
 }
 
 impl ReadyChecks {
 	/// Check if Hipcheck is ready to run.
-	///
-	/// We don't check `github_token_check`, because it's allowed to fail.
 	fn is_ready(&self) -> bool {
 		self.hipcheck_version_check.is_ok()
 			&& self.git_version_check.is_ok()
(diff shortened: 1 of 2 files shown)


sdk/rust/src/types.rs
```diff
@@ -12,9 +12,5 @@ typify_macro::import_types!(
 			type = "string",
 			format = "date-time"
 		} = jiff::Zoned,
-		{
-			type = "string",
-			format = "duration"
-		} = jiff::Span,
```
alilleybrinker, line 18: Is there a reason to delete this in this fix, or is this a mistaken carry over change from some other work?

alilleybrinker (changes requested):

sdk/rust/src/types.rs
bmarr-mitre: This is related to removing the hc_github_token from the hc ready check.  Previously there were 3 elements in the 'ok' message tuple.  Now there are 2.  In the previous commit, this line was causing the Ubuntu test to fail.

alilleybrinker: I'll check it out locally myself when I can. It doesn't make sense to me why this change to an environment variable would require removing the schema-generation mapping from the duration format to the `jiff::Span` type.

j-lanson: Yeah this is certainly not correct, but I can't see what the failing ubuntu test was to diagnose what the proper fix is. Try restoring these lines?

alilleybrinker: Agree with @j-lanson. I haven't been able to replicate the issue locally, but I'm on macOS. I don't see why this would fail on Ubuntu specifically though.

bmarr-mitre: No problem.  I'll put the lines back in and we can look at the ubuntu results.

alilleybrinker: It looks like it passed Ubuntu. Whatever error there had been must have been because of something else. I'll approve and merge.

alilleybrinker (approved):
LGTM!

PR #681, opened 2024-12-03
chore: Updates file path tags in example policy file

mchernicoff:
The example policy file `config/Hipheck.kdl` was still using the older `foo-file-path` label for some of the additional file locations, instead of `foo-file`. This fixes that so the example policy file works correctly.

### DIFF
--- a/config/Hipcheck.kdl
+++ b/config/Hipcheck.kdl
@@ -29,13 +29,13 @@ analyze {
 
     category "attacks" {
         analysis "mitre/typo" {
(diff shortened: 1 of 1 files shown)


ninaagrawal (commented):
Looks good

ninaagrawal (commented):
Nothing to change, looks fine

ninaagrawal (approved):
Looks good

ninaagrawal:
Just merged the branch

Issue #572, opened 2024-11-05
`config_to_policy` conversion references non-existent URLs for plugin endpoints


j-lanson:
when running `hc check https://github.com/mitre/hipcheck`, since no policy file is provided, HC loads the old config file and converts it to a policy file. Looks like there are references to remote URLs like `.../hipcheck/blob/main/plugin/dist/mitre-activity.kdl` which don't currently exist. Update the `config_to_policy` conversion to properly use the local `plugin.kdl` files as we have done to be able to run local plugins.

PR #709, opened 2024-12-09
fix: update config to policy file conversion and associated tests

j-lanson:
config --> policy file conversion and test files needed updates to match changes to per-plugin config/version expectations


alilleybrinker (approved):

j-lanson:
Resolves #572 .

PR #715, opened 2024-12-12
plugin download/execution improvements

### LINKED ISSUE #706: add plugin cache dir to child PATH (last) when spawning
`docker` entrypoints for plugins will probably have references to artifacts in the plugin cache dir (such as a container file), but we don't execute the plugin from that directory so these files will fail to be found. Update the plugin child process's PATH at spawn time by pushing `canon_working_dir`.

### LINKED ISSUE #707: delete archive file and uncompressed dir from plugin dir
When a plugin is retrieved from remote and stored in a plugin cache subdir, the compressed file (`.tar.xz` / `.zip` / etc) and the uncompressed folder stick around inside the subdir. This is unnecessary and they should be deleted as part of the unpacking process / after the unpacking process by the `retrieval.rs` code.

j-lanson:
Resolves #706 . Resolves #707 .

Some quality of life changes for plugins.

- Add the plugin's cache path to the plugin's process PATH to help plugin process find artifacts when running
- Add `force: bool` to plugin downloading function, and use existing plugin with right version in cache if exists and `force=false` instead of pulling and decompressing every time
- Change plugin bundle extraction to only copy up artifacts from a subdir if it matches name of the bundle minus the compression extension. Previous impl could accidently copy unrelated files if passed `extract_dir` contained other subdirs already.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1333,7 +1333,7 @@ checksum = "7f24254aa9a54b5c858eaee2f5bccdb46aaf0e486a595ed5fd8f86ba55232a70"
 
 [[package]]
 name = "hipcheck"
-version = "3.7.0"
+version = "3.8.0"
 dependencies = [
  "anyhow",
  "async-stream",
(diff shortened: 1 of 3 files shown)


patrickjcasey (approved):
Looks good and functions as expected on my machine!

PR #729, opened 2024-12-17
chore: delete deprecated proto top-level folder

patrickjcasey:
All .proto files are now stored in `hipcheck-common/proto/hipcheck/v1`


j-lanson:
I think Andrew wants the more broken down version of the `proto` dir going forward. Could you instead replace `hipcheck-common/proto` with `proto` and make sure everything compiles / runs properly?

j-lanson (changes requested):
See above comment

alilleybrinker:
We also need to validate this doesn’t break release compilation for anything. One of the challenges with the proto definitions is that they need to be inside the directory of anything released on crates.io. I recommend doing dry runs of `cargo publish` on all library crates we have to make sure they don’t break. The solution for it is to symlink into the directory so there’s one copy but its replicated into the right places.

patrickjcasey:
I ran the following and they were all successful:

- `cargo publish --dry-run -p hipcheck-sdk`
- `cargo publish --dry-run -p hipcheck-macros`
- `cargo publish --dry-run -p hipcheck-sdk-macros`

j-lanson:
I wanted us to keep the version of `proto` with the `messages` subdir, not the monolithic one

j-lanson (changes requested):
See above comment

Issue #730, opened 2024-12-17
RFD10 changes to Hipcheck .proto files


patrickjcasey:

patrickjcasey:
closed by #737

PR #737, opened 2024-12-20
feat: implemented submit chunking for RFD #0010

patrickjcasey:
- Added support for chunking Query submissions
- Added test for submission chunking

I just want to ensure that adding an enum variant to `QueryState` is not considered a breaking change. Also since this changes `hipcheck-common`, does its version need be incremented or do we just handle that when it comes time to cut a release

### DIFF
--- a/hipcheck-common/src/chunk.rs
+++ b/hipcheck-common/src/chunk.rs
@@ -36,11 +36,20 @@ fn estimate_size(msg: &PluginQuery) -> usize {
 }
 
 pub fn chunk_with_size(msg: PluginQuery, max_est_size: usize) -> Result<Vec<PluginQuery>> {
-	// Chunking only does something on response objects, mostly because
-	// we don't have a state to represent "SubmitInProgress"
-	if msg.state == QueryState::Submit as i32 {
-		return Ok(vec![msg]);
-	}
+	// in_progress_state - the state the PluginQuery is in for all queries in the resulting Vec,
+	// EXCEPT the last one
+	//
+	// completion_state - the state the PluginQuery is in if it is the last chunked message
+	let (in_progress_state, completion_state) = match msg.state() {
+		// if the message gets chunked, then it must either be a reply or submission that is in process
+		QueryState::Unspecified => return Err(anyhow!("msg in Unspecified query state")),
+		QueryState::SubmitInProgress | QueryState::SubmitComplete => {
+			(QueryState::SubmitInProgress, QueryState::SubmitComplete)
+		}
+		QueryState::ReplyInProgress | QueryState::ReplyComplete => {
+			(QueryState::ReplyInProgress, QueryState::ReplyComplete)
+		}
+	};
 
 	let mut out: Vec<PluginQuery> = vec![];
 	let mut base: PluginQuery = msg;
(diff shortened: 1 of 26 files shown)


patrickjcasey:
closes #730 #732

hipcheck-common/src/chunk.rs
```diff
@@ -162,7 +176,9 @@ impl QuerySynthesizer {
 					.map_err(|_| Error::UnspecifiedQueryState)?;
 				match state {
 					QueryState::Unspecified => return Err(Error::UnspecifiedQueryState),
-					QueryState::Submit => return Err(Error::ReceivedSubmitWhenExpectingReplyChunk),
+					QueryState::SubmitInProgress | QueryState::SubmitComplete => {
+						return Err(Error::ReceivedSubmitWhenExpectingReplyChunk)
+					}
```
j-lanson, line 181: I believe the `QuerySynthesizer` needs to be modified as well to support de-chunking chunked `Submit` queries.

hipcheck-common/src/chunk.rs
```diff
@@ -209,23 +225,40 @@ mod test {
...
+			assert_eq!(
+				res.iter()
+					.filter(|x| x.state() == intermediate_state)
+					.count(),
+				3
+			);
+			// ensure last one is ...Complete
+			assert_eq!(res.last().unwrap().state(), final_state);
+			assert_eq!(res.len(), 4);
+		}
```
j-lanson, line 262: To ensure the above `QuerySynthesizer` is implemented correctly, perhaps there should be an additional stage to this test that tries to reassemble the original message from the fragments.

sdk/rust/src/error.rs
```diff
@@ -71,6 +71,7 @@ impl From<hipcheck_common::error::Error> for Error {
 		use hipcheck_common::error::Error::*;
 		match value {
 			UnspecifiedQueryState => Error::UnspecifiedQueryState,
+			UnexpectedRequestInProgress => Error::UnexpectedReplyInProgress,
```
j-lanson, line 74: Should add another variant to `Error` instead of re-using `UnexpectedReplyInProgress`

j-lanson (changes requested):
Missing some logic in the QuerySynthesizer, but the rest is looking good!

j-lanson:
On the subject of versioning, I would do the following:

1. Bump `hipcheck-common` version to `0.2.0`
2. Update `rust-sdk` to take the new version, and bump it's version to `0.3.0`
3. Update the plugins to take SDK `0.3.0`

We can add RFD10 to `hipcheck-common-v0.2.0` without another version bump since we haven't released anything yet.

On the next release, we will release new versions of the rust SDK and all plugins.

patrickjcasey:
I applied the fixes you suggested and fixed a couple of minor issues with usage of `serde_json`. The `from_str` and `to_string` methods were being used when  directly assigning to a `serde_json::Value` and calling `serde_json::Value::from_value` was more appropriate

patrickjcasey:
This is not quite done yet, still working through a couple of things

hipcheck-common/src/chunk.rs
```diff
@@ -71,44 +80,47 @@ pub fn chunk_with_size(msg: PluginQuery, max_est_size: usize) -> Result<Vec<Plug
...
-				query.concern.push(concern);
+				chunked_query.concern.push(concern);
 				remaining -= c_bytes;
 				made_progress = true;
 			}
 			// since we use `swap_remove`, whether or not we stole a concern we know the element
 			// currently at `i` is too big for `remainder` (since if we removed, the element at `i`
 			// now is one we already passed on)
-			l -= 1;
+			i += 1;
```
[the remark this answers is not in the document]
patrickjcasey: I was attempting to maintain concern ordering, but the original approach is better because it is able to potentially pack non-consecutive concerns in one message, potentially reducing total # of messages. I reverted this portion of the commit to use the original logic

patrickjcasey:
just pushed to this branch again. It seems like there were some new `clippy` warnings that popped up, so I also fixed those

alilleybrinker:
> just pushed to this branch again. It seems like there were some new `clippy` warnings that popped up, so I also fixed those

Makes sense, Rust released 1.84.0 yesterday, which we pick up immediately in CI. New Rust versions usually mean new Clippy lints.

j-lanson (approved):
Thanks Patrick!

PR #750, opened 2024-12-30
CLRU added for git plugin

aamohd:
12/30 - The PR has changes to the `git` plugin that uses an LRU cache functionality. The configurability of the size of the cache would come after the #662 changes are pulled.

### DIFF
--- a/plugins/git/src/git.rs
+++ b/plugins/git/src/git.rs
@@ -2,6 +2,8 @@
 
 use crate::data::*;
 
+use crate::Error;
+use crate::CACHE;
 use anyhow::Context;
 use anyhow::Result;
 use gix::bstr::ByteSlice;
@@ -18,13 +20,12 @@ use gix::traverse::commit::simple::CommitTimeOrder;
 use gix::ObjectId;
 use gix::Repository;
 use jiff::Timestamp;
+use lru::LruCache;
 use std::path::Path;
 use std::path::PathBuf;
-use std::sync::Mutex;
 
 /// used to cache all of the `RawCommit` from the last repo/HEAD combination analyzed by this
-/// plugin
-type GitRawCommitCache = Option<(PathBuf, ObjectId, Vec<RawCommit>)>;
+pub type GitRawCommitCache = LruCache<(PathBuf, ObjectId), Vec<RawCommit>>;
 
 /// retrieve a handle to the git repo at this path, as well as determine the commit hash of
 /// HEAD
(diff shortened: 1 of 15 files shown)


plugins/git/src/git.rs
```diff
@@ -18,13 +19,17 @@ use gix::traverse::commit::simple::CommitTimeOrder;
...
 use std::path::PathBuf;
+use std::sync::LazyLock;
 use std::sync::Mutex;
 
 /// used to cache all of the `RawCommit` from the last repo/HEAD combination analyzed by this
 /// plugin
-type GitRawCommitCache = Option<(PathBuf, ObjectId, Vec<RawCommit>)>;
+type GitRawCommitCache = CLruCache<(PathBuf, ObjectId), Vec<RawCommit>>;
+static CACHE: LazyLock<Mutex<GitRawCommitCache>> =
+	LazyLock::new(|| Mutex::new(CLruCache::new(NonZero::new(1).unwrap())));
```
j-lanson, line 32: The point of this PR is to make the `1` you have here a configurable parameter

plugins/git/src/git.rs
```diff
@@ -4,6 +4,7 @@ use crate::data::*;
 
 use anyhow::Context;
 use anyhow::Result;
+use clru::CLruCache;
```
j-lanson, line 7: Is there a reason that the `clru` crate is better for this than `lru`? https://docs.rs/lru/latest/lru/

j-lanson (changes requested):

plugins/git/src/git.rs
```diff
@@ -108,6 +109,7 @@ where
 {
 	let commit_walker = get_commit_walker(repo, head_commit)?;
 	let commits = walk_commits(repo, commit_walker, &get_raw_commit, None)?;
+	// let raw_commits: GitRawCommitCache =
```
j-lanson, line 112: Delete comment

plugins/git/src/git.rs
```diff
@@ -122,24 +124,23 @@ pub fn get_all_raw_commits<P>(repo_path: P) -> Result<Vec<RawCommit>>
 where
 	P: AsRef<Path>,
 {
-	// used to cache all of the RawCommits from the last repository analyzed
-	static ALL_RAW_COMMITS: Mutex<GitRawCommitCache> = Mutex::new(None);
-
 	let (repo, head_commit) = initialize_repo(repo_path.as_ref())?;
-	let mut cache = ALL_RAW_COMMITS.lock().unwrap();
+	let unlocked_cache = CACHE.get().ok_or(Error::UnspecifiedQueryState)?;
+	let mut cache = unlocked_cache.lock().unwrap();
```
j-lanson, line 129: replace the `unwrap()` with `?` and map the error if need-be.

plugins/git/src/main.rs
```diff
@@ -428,16 +436,52 @@ async fn commit_contributors(
...
+
+impl TryFrom<RawConfig> for Config {
+	type Error = hipcheck_sdk::error::ConfigError;
+	fn try_from(value: RawConfig) -> StdResult<Config, Self::Error> {
+		let opt_commit_cache_size = value.commit_cache_size;
+		Ok(Config {
+			opt_commit_cache_size,
+		})
+	}
+}
```
j-lanson, line 457: This pattern is unnecessary for your purpose. Note that `RawConfig` and `Config` are the exact same struct in terms of fields. Just remove `Config` and `TryFrom<RawConfig>`, and rename `RawConfig` to `Config`. The raw-to-validated pattern is only necessary if there are programmatic reasons why a value may or may not be valid. In this case, any usize is correct (we might say 0 is incorrect, but the code technically still works).

plugins/git/src/main.rs
```diff
@@ -428,16 +436,52 @@ async fn commit_contributors(
 	Ok(commit_contributors)
 }
 
-#[derive(Clone, Debug)]
+#[derive(Deserialize)]
+struct RawConfig {
+	#[serde(rename = "commit-cache-size")]
+	commit_cache_size: Option<usize>,
```
j-lanson, line 442: You can use `#[serde(default = "<VALUE>")]` and remove the need for an `Option` here. This way if the field is not specified, it gets a default value. This will clean up your `set_config()` code below.

site/content/docs/guide/plugins/mitre-git.md
```diff
@@ -8,3 +8,9 @@ extra:
 
 Provides access to Git commit history data. Does not define a default query
 and can't be used as a top-level plugin in a policy file.
+
+## Configuration
+
+| Parameter           | Type    | Explanation   |
+|:--------------------|:--------|:--------------|
+| `commit-cache-size` | `Integer` | Number of repositories to retain in cache. |
```
j-lanson, line 16: `Number of repositories to retain in cache.` --> `Optional number of repositories to retain in cache. Defaults to one.`

j-lanson (changes requested):
A few logic changes to make the code cleaner. Also, you'll need to rebase off main and address the `git/Cargo.toml` merge conflict.

j-lanson (changes requested):
Please just change the commit message to `feat: configurable commit cache size in git plugin`

j-lanson (approved):

PR #751, opened 2025-01-03
fix minor issues in Containerfile for v3.9.0

AevaOnline:
While attempting to locally build and run the docker image from the v3.9.0 release tag, I found several minor issues, addressed below. I also found one that I'm not sure how to fix, and will open an Issue for it.

- Both the `ENTRYPOINT` and `CMD` lines should be written in JSON notation to allow interactive use of "docker run". The current non-JSON form prevents user from running any CMD other than "help".

- in the source code on the v3.9.0 release tag, the version string in `dist/Containerfile` is still 3.7.0

- Between v3.7.0 and v3.8.0, the installed path changes from $HOME/.cache to $HOME/.local

After correcting the above issues, a default policy file is not found, but this is easily remedied now that `CMD` uses JSON notation and interactive `docker run` commands are possible:
```
docker run -it --rm sha256:9cae59257cca559025ebbbc958fa5b614396ab7e7de8c32cad016e90b077a056  --policy /home/hc_user/.config/hipcheck/Hipcheck.kdl ready
Hipcheck Version: hipcheck 3.9.0
Git Version:      git version 2.39.5
NPM Version:      10.9.2
Cache Path:       /home/hc_user/.cache/hipcheck
Policy Path:      /home/hc_user/.config/hipcheck/Hipcheck.kdl
Hipcheck is ready to run!
``

### DIFF
--- a/dist/Containerfile
+++ b/dist/Containerfile
@@ -2,7 +2,7 @@
 
 FROM node:bookworm-slim
 
-ARG HC_VERSION="3.7.0"
+ARG HC_VERSION="3.9.0"
 
 WORKDIR /app
 
@@ -17,8 +17,8 @@ USER hc_user
 
 RUN set -eux \
     && curl --proto '=https' --tlsv1.2 -LsSf https://github.com/mitre/hipcheck/releases/download/hipcheck-v${HC_VERSION}/hipcheck-installer.sh | sh \
-    && $HOME/.cargo/bin/hc setup
+    && $HOME/.local/bin/hc setup
 
-ENV HC_CONFIG="$HOME/.config/hipcheck"
-ENTRYPOINT "$HOME/.cargo/bin/hc"
-CMD "help"
+ENV HC_CONFIG="/home/hc_user/.config/hipcheck"
+ENTRYPOINT ["/home/hc_user/.local/bin/hc"]
+CMD ["help"]

### CODE REVIEW

File: dist/Containerfile [hunk 1]
```diff
@@ -17,8 +17,8 @@ USER hc_user
...
 RUN set -eux \
     && curl --proto '=https' --tlsv1.2 -LsSf https://github.com/mitre/hipcheck/releases/download/hipcheck-v${HC_VERSION}/hipcheck-installer.sh | sh \
-    && $HOME/.cargo/bin/hc setup
+    && $HOME/.local/bin/hc setup
 
-ENV HC_CONFIG="$HOME/.config/hipcheck"
-ENTRYPOINT "$HOME/.cargo/bin/hc"
-CMD "help"
+ENV HC_CONFIG="/home/hc_user/.config/hipcheck"
+ENTRYPOINT ["/home/hc_user/.local/bin/hc"]
```
j-lanson, line 23: Just curious, is there a reason `$HOME` no longer works in place of `/home/hc_user` here?

AevaOnline: Reading through docker's docs on this, it appears to be a limitation of the "exec form" of CMD and ENTRYPOINT directives. To use $HOME with the JSON format, documentation suggests changing to explicit invocation of a bash shell. 

https://docs.docker.com/reference/dockerfile/#variable-substitution

AevaOnline: I wonder if a more general fix -- both for this and the plugin issue -- would be to install the `hc` binary in `/usr/local/bin/` and copy the configs, plugins, cache, etc, into a common directory, such as `/app`, that could be overridden with a bindmount by the user?

Such a final invocation might then take this form:

`docker run -ti --volume ./my_config:/app mitre/hipcheck:latest --policy /app/Hipcheck.kdl check $URL`

j-lanson: The cache (containing repos and plugins) is already configuable, so a user who wanted their image to access a persistent cache on the host would just need to volume mount and then use `--cache <PATH_TO_CACHE_IN_IMAGE>` in the `check` invocation.

The problem in #752 is likely that we simply need to copy those config files into the image when we build it. I'm working on installing podman to be able to investigate locally, but it requires me to update my WSL Ubuntu from 20 to 22 which is taking a while.

AevaOnline: Thanks, that actually resolved the issue in #752! 

I created a local directory `docker_volume` and copied in the config and plugin files, mounted it and specified it with `--cache` and then 3.9.0 docker image works :) 

```
❯ docker run -it --rm --volume ./docker_volume:/app -e HC_GITHUB_TOKEN='...' mitre/hipcheck:local-3.9.0  --cache /app/ --policy /app/config/Hipcheck.kdl check https://github.com/mitre/hipcheck

           Analyzing GitHub repo mitre/hipcheck from https://github.com/mitre/hipcheck
                Done loading policy and data files (0 seconds)
                Done resolving git repository target (0 seconds)
                Done analyzing and scoring results (8 seconds)

            Analyzed hipcheck (main)
                     using Hipcheck 3.9.0
                     on Sat January 4, 2025 at 12:49am
...
```

j-lanson: Glad its working for you now :)

That being said, we want the Containerfile to still work without volume mount, and based on your invocation I see your mount also includes the `config` subdir which explains why now the plugins are happy. There are still things to run to ground on our side so that users can go the Containerfile route without these hoops.

j-lanson (changes requested):
You'll need to reword the commit message to match the Conventional Commits spec in order to pass the CI

j-lanson:
I think the solution to the distributed/containerized Hipcheck will be based on implementing #764.

This will allow us to write a policy/config file that specifies the paths of plugin-specific config files relative to itself. So if `Hipcheck.toml` is in `/home/hc_user/.config/hipcheck/`, the fact that we run `hc` from a totally different location will not cause the above errors with plugins unable to find their config files.

So, starting with the 3.10 release, we will release a modified version of the policy/config files in `config/` with this new macro.

j-lanson:
Closed in favor of #766 . Marked @AevaOnline as co-author in commit message.

Issue #256, opened 2024-08-06
Hipcheck errors without proper explanation when given an NPM or PyPI package version not tagged in the repository


mchernicoff:
When Hipcheck is given an NPM or PyPI package to check with a version, it uses that version number as a tag refspec for the corresponding Git repository. If that repository does not have a tag corresponding to that version (e.g. the tag for an older release has been deleted) or the tag is not formatted exactly as a plain semver string (e.g. `v.1.2.3` instead of `1.2.3`), Hipcheck will quit with the following error message:

> revspec '1.2.3' not found; class=Reference (4); code=NotFound (-3)

As a short term fix, we should change the error message to explain what has happened and suggest corrective action for the user (e.g. try again with no version number and an optional `--ref` flag that matches an existing tag). Longer term, we may wish to handle a missing version tag more gracefully.

alilleybrinker:
Probably try some common patterns for matching tags, and otherwise error out and tell the user to use the `--ref` flag.

To match version `X.Y.Z` we should check:

- `vX.Y.Z`
- `packagename-X.Y.Z`
- `packagename-vX.Y.Z`
- `packagename_X.Y.Z`
- `packagename_vX.Y.Z`

j-lanson:
Closed by #754 .

PR #754, opened 2025-01-03
feat: improved version-to-git-tag matching scheme

### LINKED ISSUE #753: Add fuzzy version matching for repos derived from packages

j-lanson:
Resolves #753, #755, and #256 .

If during resolution we have a `Package` with a non-null version, we try to resolve to a tag in the repo with a variety of combinations of version specifier and package name. Verified that this works against `-t nmp duplexer2@0.1.4`, which previously gave us problems. Our previous dirty fix of checking `v<VERSION>` inside of `git::checkout()` has been removed.

Edit: Based on discussion in the Slack channel, I've augmented the logic. If a package is specified with no version and no ref, we will use a forgiving semver regex to try and find the "latest" version tag in the repo.


mchernicoff (dismissed):

mchernicoff:
Needs a rebase, but otherwise LGTM.

mchernicoff (approved):

PR #762, opened 2025-01-06
feat: implemented list function for plugins

ninaagrawal:
first implementation of the list function for plugins

### DIFF
--- a/hipcheck/src/cache/plugin.rs
+++ b/hipcheck/src/cache/plugin.rs
@@ -1,23 +1,145 @@
 // SPDX-License-Identifier: Apache-2.0
-
-use std::path::{Path, PathBuf};
-
+#![allow(unused)]
+use crate::{cache::repo, error::Result, plugin::PluginId, StdResult};
 use pathbuf::pathbuf;
+use std::{
+	borrow::Borrow,
+	env,
+	path::{Component, Path, PathBuf},
+	time::{Duration, SystemTime},
+};
+use tabled::{Table, Tabled};
+use walkdir::{DirEntry, WalkDir};
+//use super::super::cli::CliConfig;
+use regex::Regex; //used for regex
+use semver::{Version, VersionReq};
 
-use crate::plugin::PluginId;
+///enums used for sorting are the same as the ones used for listing repo cache entries except for
+/// does not include Largest
+#[derive(Debug, Clone)]
+pub enum PluginCacheSort {
+	OldestDate, //same as "Oldest" in repo.rs
+	Alpha,
+	NewestVersion, //sorts the versions of the plugins from newest to oldest unless inverted
+}
+
+#[derive(Debug, Clone)]
+pub enum PluginCacheDeleteScope {
+	All,
+	Group {
+		sort: PluginCacheSort,
+		invert: bool,
+		n: usize,
+	},
+}
+
+#[derive(Debug, Clone)]
+pub struct PluginCacheListScope {
+	pub sort: PluginCacheSort,
+	pub invert: bool,
+	pub n: Option<usize>,
+}
+
+#[derive(Debug, Clone, Tabled)]
+pub struct PluginCacheEntry {
+	pub publisher: String,
+	pub name: String,
+	pub version: String,
+	#[tabled(display_with("Self::display_modified", self), rename = "last_modified")]
+	pub modified: SystemTime,
+}
+
+impl PluginCacheEntry {
+	//copied the function from repo for simplicity
+	pub fn display_modified(&self) -> String {
+		let Ok(dur) = self.modified.duration_since(SystemTime::UNIX_EPOCH) else {
+			return "<DISPLAY_ERROR>".to_owned();
+		};
+		let Some(dt) = chrono::DateTime::<chrono::offset::Utc>::from_timestamp(
+			dur.as_secs() as i64,
+			dur.subsec_nanos(),
+		) else {
+			return "<DISPLAY_ERROR>".to_owned();
+		};
+		let chars = dt.to_rfc2822().chars().collect::<Vec<char>>();
+		// Remove unnecessary " +0000" from end of rfc datetime str
+		chars[..chars.len() - 6].iter().collect()
+	}
+}
+
+struct HcPluginCacheIterator {
+	wd: Box<dyn Iterator<Item = StdResult<DirEntry, walkdir::Error>>>,
+}
+
+impl HcPluginCacheIterator {
+	fn new(path: &Path) -> Self {
+		HcPluginCacheIterator {
+			wd: Box::new(
+				WalkDir::new(path) //builds a new iterator using WalkDir
+					.min_depth(3) //makes sure to get the publisher, name, and version folders
+					.max_depth(3)
+					.into_iter()
+					.filter_entry(|e| e.path().is_dir()), //makes sure the path is a directory
+			),
+		}
+	}
+	fn path_to_plugin_entry(&self, path: &Path) -> Result<PluginCacheEntry> {
+		let components: Vec<String> = path
+			.components()
+			.filter_map(|component| {
+				if let Component::Normal(name) = component {
+					//extracts components separated by "/" from the path
+					name.to_str().map(|s| s.to_string()) //converts the string slices into Strings
+				} else {
+					None
+				}
+			})
+			.collect(); //collects filepath into vector
+			   //Todo: add error handling for when the path length is less than 3
+		let relevant_components: &[String] = &components[components.len() - 3..];
+		let plugin_publisher = relevant_components[0].to_owned();
+		let plugin_name = relevant_components[1].to_owned();
+		let plugin_version = relevant_components[2].to_owned();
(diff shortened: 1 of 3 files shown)


hipcheck/src/cache/plugin.rs
```diff
@@ -1,21 +1,154 @@
...
+use semver::{Version, VersionReq};
+use crate::cache::repo;
 
-use std::path::{Path, PathBuf};
+///enums used for sorting are the same as the ones used for listing repo cache entries except for
+/// does not include Largest
+#[derive(Debug, Clone)]
+pub enum PluginCacheSort {  
+	Oldest,
+	Alpha,
```
j-lanson, line 24: I think we also want a sort for highest-to-lowest version numbers

hipcheck/src/cache/plugin.rs
```diff
@@ -1,21 +1,154 @@
...
+struct HcPluginCacheIterator {
+	wd: Box<dyn Iterator<Item = StdResult<DirEntry, walkdir::Error>>>
+}
+
+impl HcPluginCacheIterator { 
+	fn new(path: &Path) -> Self {
+		HcPluginCacheIterator{
+			wd: Box::new(
+				WalkDir::new(path) //builds a new iterator using WalkDir
+					.min_depth(3) //makes sure to get the publisher, name, and version folders
```
j-lanson, line 81: probably want to add `.max_depth(4)` so we only get exactly the plugin dirs.

hipcheck/src/cache/plugin.rs
```diff
@@ -1,21 +1,154 @@
...
+					.into_iter()
+					.filter_entry(|e| e.path().is_dir()), //makes sure the path is a directory
+			)
+		}	
+	}
+	fn path_to_plugin_entry(&self, path: &Path) -> Result<PluginCacheEntry> { //does function need to work when you clone a repo?
```
j-lanson, line 87: Delete comment here

hipcheck/src/cache/plugin.rs
```diff
@@ -1,21 +1,154 @@
...
+	   let components: Vec<String> = path
+        .components()
+        .filter_map(|component| {
+            if let Component::Normal(name) = component { //extracts components seperated by "/" from the path
```
j-lanson, line 91: `seperated` --> `separated`

hipcheck/src/cache/plugin.rs
```diff
@@ -1,21 +1,154 @@
...
+		let last_modified: SystemTime = repo::get_last_modified_or_now(path);
+		Ok(PluginCacheEntry{
+							publisher:plugin_publisher, 
+							name:plugin_name, 
+							version: plugin_version,
+							modified: last_modified
+							}) 
+		
+	}
+
```
j-lanson, line 113: Some of this formatting seems a bit weird, can you make sure to run `cargo fmt` locally to run through the Rust formatter?

hipcheck/src/cache/plugin.rs
```diff
@@ -1,21 +1,154 @@
...
+impl Iterator for HcPluginCacheIterator {
+	type Item = PluginCacheEntry;
+	fn next(&mut self) -> Option<Self::Item> { 
+		if let Some(Ok(e)) = self.wd.next() {
+			if let Ok(ce) = self.path_to_plugin_entry(e.path()) {
+				Some(ce)
+			} 
+			else {
+				None
+			}
```
j-lanson, line 129: this `if else` structure can be replaced with doing `.ok()` on the output of `self.path_to_plugin_entry(e.path())`, which does the same thing. Turns `Result<T>` into `Option<T>`, discarding error. Same with the enclosing `if else`.

hipcheck/src/cache/plugin.rs
```diff
@@ -1,21 +1,154 @@
...
-use pathbuf::pathbuf;
 
-use crate::plugin::PluginId;
 
 /// Plugins are stored with the following format `<path_to_plugin_cache>/<publisher>/<plugin_name>/<version>`
 pub struct HcPluginCache {
-	/// path to the root of the plugin cache
-	path: PathBuf,
+	pub path: PathBuf, //path to the root of the plugin cache
+	pub entries: Vec<PluginCacheEntry>, //needs to store a vector of PluginEntries
```
j-lanson, line 142: Instead of making fields these `pub`, I would make a special constructor for `HcPluginCache` marked with `#[cfg(test)]` so you can only create it that way in the testing code.

hipcheck/src/cache/plugin.rs
```diff
@@ -34,4 +167,379 @@ impl HcPluginCache {
...
+				(PluginCacheSort::Oldest, true) => {
+					|a, b| b.modified.partial_cmp(&a.modified).unwrap()
+				}
+			};
+			entries.sort_by(|a1: &A, a2: &A| sort_func(a1.borrow(), a2.borrow()));
+		}
+
+	///lists all the plugin entries. Works the same as the function in repo.rs except for 
+	///user can filter plugins by version and publisher
+	/// TODO: split into inner function that you can call, outer function can call inner and display on the result
```
j-lanson, line 189: TODO comment can be removed now

hipcheck/src/cache/plugin.rs
```diff
@@ -34,4 +167,379 @@ impl HcPluginCache {
...
+			sort:PluginCacheSort::Alpha, 
+			invert:true, 
+			n:None,
+		};
+
+		let entries = &test_data();
+		let results = HcPluginCache::list_inner(entries, time_sort, None, None, None);
+		let expected_output = vec!["affiliation", "activity", "bugs bunny", "difference"];
+		let mut actual_output: Vec<&String> = Vec::new();
+		match results {
```
j-lanson, line 306: Instead of doing `match results {}`, you can just do `list_inner(...).unwrap()` on line 303 to deal with the error case. We don't care if test cases panic, because it is treated as a test failure. 

Also, to create `actual_output` you can do something like `results.map(|x| &x.name)` instead of your for-loop. Should cut down a lot on code lines. Don't forget to propagate these changes to the other tests below.

hipcheck/src/cache/repo.rs
```diff
@@ -64,7 +64,7 @@ impl RepoCacheEntry {
 			.unwrap_or("<DISPLAY_ERROR>".to_owned())
 			.to_string()
 	}
-	fn display_modified(&self) -> String {
+	pub fn display_modified(&self) -> String {
```
j-lanson, line 67: this likely doesn't need to be `pub` anymore because you copied the implementation wholesale

j-lanson (changes requested):
I appreciate the thorough testing Nino. I'm requesting some changes for clarity and code hygiene, but overall it is looking good.

hipcheck/src/cache/plugin.rs
ninaagrawal: what about max_depth(3)?

j-lanson: good point, that should work. We just want to make it so only Paths exactly 3 subdirs down are returned.

hipcheck/src/cache/plugin.rs
ninaagrawal: I ran cargo fmt in the hipcheck directory but please let me know if I need to do anything else

j-lanson: You should set up an automation with whatever IDE you are using to run `cargo fmt` whenever you save

hipcheck/src/cache/plugin.rs
```diff
@@ -1,23 +1,155 @@
...
+		}
+	}
+}
 
 /// Plugins are stored with the following format `<path_to_plugin_cache>/<publisher>/<plugin_name>/<version>`
 pub struct HcPluginCache {
-	/// path to the root of the plugin cache
-	path: PathBuf,
+	path: PathBuf,                  //path to the root of the plugin cache
+	entries: Vec<PluginCacheEntry>, //needs to store a vector of PluginEntries
```
j-lanson, line 130: the comment on `entries` feels a bit redundant

hipcheck/src/cache/plugin.rs
```diff
@@ -1,23 +1,155 @@
...
+	}
+	///This Constructor is only meant to be used when testing the code
+	#[cfg(test)]
+	pub fn new_for_test() -> Self {
+		let plugins_path = Path::new("C:/Users/ninaagrawal/AppData/Local/hipcheck/plugins");
+		let entries: Vec<PluginCacheEntry> = HcPluginCacheIterator::new(plugins_path).collect();
+		Self {
+			path: pathbuf![plugins_path],
+			entries,
+		}
```
j-lanson, line 151: Based on the way you are doing the tests below, just calling `list_inner()`, you can delete this function. When I made the suggestion to write this func, I was thinking of a different approach, but yours is completely fine and good.

j-lanson (changes requested):
Just two minor things, otherwise its looking good to be merged

j-lanson (approved):
Looks good!

PR #791, opened 2025-01-15
rfd9 implementation

patrickjcasey:
This pull request implements the majority of the guts of RFD #0009

- `hipcheck.proto` has been updated to make Query.key and Query.output
"repeated string fields"
- `hipcheck.proto` has been updated to add a new field, split, to Query
- chunking of GRPC messages now maintains initial ordering of fields
when chunking is required
- additional unit tests added for chunking algorithm
- added `batch_query` function to rust SDK

### DIFF
--- a/hipcheck-common/src/types.rs
+++ b/hipcheck-common/src/types.rs
@@ -12,8 +12,8 @@ pub struct Query {
 	pub publisher: String,
 	pub plugin: String,
 	pub query: String,
-	pub key: serde_json::Value,
-	pub output: serde_json::Value,
+	pub key: Vec<serde_json::Value>,
+	pub output: Vec<serde_json::Value>,
 	pub concerns: Vec<String>,
 }
 
@@ -51,17 +51,28 @@ impl TryFrom<PluginQuery> for Query {
 
 	fn try_from(value: PluginQuery) -> Result<Query, Self::Error> {
 		let direction = QueryDirection::try_from(value.state())?;
-		let key = serde_json::from_str(&value.key).map_err(Error::InvalidJsonInQueryKey)?;
-		let output =
-			serde_json::from_str(&value.output).map_err(Error::InvalidJsonInQueryOutput)?;
+
+		let mut keys = Vec::with_capacity(value.key.len());
+		for x in value.key.into_iter() {
+			let value = serde_json::from_str(x.as_str()).map_err(Error::InvalidJsonInQueryKey)?;
+			keys.push(value);
+		}
+
+		let mut outputs = Vec::with_capacity(value.output.len());
+		for x in value.output.into_iter() {
+			let value =
+				serde_json::from_str(x.as_str()).map_err(Error::InvalidJsonInQueryOutput)?;
+			outputs.push(value);
+		}
+
 		Ok(Query {
 			id: value.id as usize,
 			direction,
 			publisher: value.publisher_name,
 			plugin: value.plugin_name,
 			query: value.query_name,
-			key,
-			output,
+			key: keys,
+			output: outputs,
 			concerns: value.concern,
 		})
 	}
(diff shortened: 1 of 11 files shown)


hipcheck-common/src/types.rs
```diff
@@ -12,9 +12,10 @@ pub struct Query {
 	pub publisher: String,
 	pub plugin: String,
 	pub query: String,
-	pub key: serde_json::Value,
-	pub output: serde_json::Value,
+	pub key: Vec<serde_json::Value>,
+	pub output: Vec<serde_json::Value>,
 	pub concerns: Vec<String>,
+	pub split: bool,
```
j-lanson, line 18: Unless there is a reason why users should be able to set `split`, which I don't think there is, you can just remove `split` as a field from `Query`. It is a field necessary only for fragmented comms

hipcheck-common/src/types.rs
```diff
@@ -72,19 +86,30 @@ impl TryFrom<Query> for PluginQuery {
...
 			state: state as i32,
 			publisher_name: value.publisher,
 			plugin_name: value.plugin,
 			query_name: value.query,
-			key,
-			output,
+			key: keys,
+			output: outputs,
 			concern: value.concerns,
+			split: value.split,
```
j-lanson, line 112: and then make this `false`

hipcheck/src/score.rs
```diff
@@ -220,7 +220,10 @@ pub fn score_results(_phase: &SpinnerPhase, db: &dyn ScoringProvider) -> Result<
 			// Determine if analysis passed by evaluating policy expr
 			let passed = {
 				if let Ok(output) = &response {
-					std_exec(policy.clone(), Some(&output.value)).map_err(|e| hc_error!("{}", e))?
+					// by this time, the result cached should have evaluated to a single Value, so
+					// use the first value
+					std_exec(policy.clone(), Some(output.value.first().unwrap()))
+						.map_err(|e| hc_error!("{}", e))?
```
j-lanson, line 226: I don't see anything checking the size of `output.value` before this point. We should add some logic before callign `std_exec` to return an error if the plugin returned more than 1 output object

j-lanson (changes requested):
A few small changes, but looks great. If you feel like the chunking algorithm has been adequately tested I'm happy to merge after you make these changes.

hipcheck/src/score.rs
```diff
@@ -220,7 +220,14 @@ pub fn score_results(_phase: &SpinnerPhase, db: &dyn ScoringProvider) -> Result<
 			// Determine if analysis passed by evaluating policy expr
 			let passed = {
 				if let Ok(output) = &response {
-					std_exec(policy.clone(), Some(&output.value)).map_err(|e| hc_error!("{}", e))?
+					// by this time, the result cached should have evaluated to a single Value, so
+					// use the first value
+					if output.value.len() != 1 {
+						return Err(hc_error!("Analysis output had more than one value"));
```
j-lanson, line 226: We should specify which analysis in the error message

sdk/rust/src/engine.rs
```diff
@@ -273,6 +301,12 @@ impl PluginEngine {
 		inner(self, concern.as_ref())
 	}
 
+	#[cfg(feature = "mock_engine")]
+	#[cfg_attr(docsrs, doc(cfg(feature = "mock_engine")))]
+	pub fn get_concerns(&self) -> &[String] {
```
j-lanson, line 306: Please put a brief doc comment, something like "Exposes the current set of concerns recorded by the PluginEngine"

j-lanson (changes requested):

patrickjcasey:
I added the doc comment to `get_concerns` and made the error in `score.rs` include which plugin had the error

j-lanson (approved):
Thanks Patrick!

PR #796, opened 2025-01-16
chore(docs): clean up exposed symbols and doc comments for Rust SDK

### LINKED ISSUE #790: improve Rust SDK docs and packages
Add doc coments to introduce the crate, clean up doc comments on functions and types within the crate.

Audit the symbols and sub-modules exposed by the crate and potentially re-organize. 

[Omnibor](https://docs.rs/omnibor/latest/omnibor/#modules) may be a good model.

j-lanson:
Resolves #790 .

Updated exports interface, added overall description with links to website, cleaned up doc comments.

Supported `cargo doc` build configs:
- `stable`/`nightly`: `cargo doc`, `cargo doc --all-features`
- `nightly`:  `RUSTDOCFLAGS="--cfg docsrs" cargo doc --all-features` (adds tags showing which symbols/funcs are feature-gated, this is what `Docs.rs` will show)

Relevant points:
- Added `rustdoc-args = ["--cfg", "docsrs"]` to Cargo.toml and `#[cfg_attr(docsrs, doc(cfg(feature = "X")))]` to relevant exports so that when docs are built, will tag those exports as being only available with feature `X`
- Moved contents of `mock.rs` into `engine.rs` and made `mock` a module defined in `lib.rs`, so that we can hide it if `mock_engine` feature is not available while still making the contents available to `PluginEngine`
- I was able to test the docs generation roughly the way they would be built on `docs.rs` with `RUSTDOCFLAGS="--cfg docsrs" cargo doc --all-features`, but I still don't have perfect confirmation the docs build will succeed on release.


alilleybrinker:
Looking at this now! :smile:

alilleybrinker (approved):

PR #810, opened 2025-01-21
feat: added "batch" function to PluginEngine in rust SDK

patrickjcasey:
- added convenience function for instantiating a `QueryBuilder` to dynamically buildup a query and send to a target plugin endpoint
- completed RFD 0009 implementation

### DIFF
--- a/sdk/rust/src/engine.rs
+++ b/sdk/rust/src/engine.rs
@@ -14,12 +14,12 @@ use hipcheck_common::{
 	types::{Query, QueryDirection},
 };
 use serde::Serialize;
-use std::sync::Arc;
 use std::{
 	collections::{HashMap, VecDeque},
 	future::poll_fn,
 	pin::Pin,
 	result::Result as StdResult,
+	sync::Arc,
 };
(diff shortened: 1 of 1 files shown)


sdk/rust/src/engine.rs
```diff
@@ -547,3 +593,29 @@ impl MockResponses {
...
+			.unwrap();
+		let mut engine = PluginEngine::mock(mock_responses);
+		let mut builder = engine.batch("mitre/foo").unwrap();
+		let idx = builder.query("abcd".into());
+		assert_eq!(idx, 0);
+		let idx = builder.query("efgh".into());
+		assert_eq!(idx, 1);
+		let response = builder.send().await.unwrap();
+		assert!(response.contains(&1234.into()));
+		assert!(response.contains(&5678.into()));
```
j-lanson, line 619: Not a big deal, but we probably want to assert they are in the right order too, not just `contains()`

j-lanson (changes requested):

j-lanson (approved):

Issue #383, opened 2024-09-11
Adding english-language description of final policy expr boolean evaluation


j-lanson:
With RFD#0004, Hipcheck can handle an unlimited number of analyses, but we lose the hard-coded semantic information that made the report human-readable (i.e. the knowledge of concern format / output units and meaning was embedded into the code). When a plugin analysis fails by comparison, all we do currently is say "X was the output, the policy was Y". Printing these policies raw in the report output is not very ergonomic.

By necessity, the top of the AST for a policy expression must be an operation that returns a boolean. Investigate to what extent we can provide english language descriptions for this final operation, combined with the `default_query_explanation()` that each plugin provides, to give a generated explanation of why the policy fails. To start, I think limiting this to just the top-level boolean-returning op and its immediate operands is a good goal.

e.g.
entropy, `(eq 0 (count (filter (gt 8.0) $)))` --> "expected the `number of commit entropies` passed through `(count (filter (gt 8.0) $))` to equal `0`, was `X`"
review `(lte $ 0.05)` --> "expected the `percent of commits receiving review`  to be less than `0.05`, was `X`"

j-lanson:
This will likely rely on #371 , as we need a structured JSON-pointer type to know which operand of the top-level boolean expression contains the metric data.

Also, should be able to "collapse" any portion of the expression that does not rely on JSON-pointer data. So when we handle `(eq (sub 10 5) ..)` we should be able to evaluate the entire first operand

PR #816, opened 2025-01-22
feat: Add English language parsing of failure messages

mchernicoff:
Part of Issue #383 . Adds primitive English-language summaries of policy expressions for failed plugins.

The policy expression level English parser requires that the "top level" of a policy expression be a function that takes two arguments (as the comparison operator functions that return a Boolean all do) where at least one argument is a Primitive type. It then applies a recursive English parser to each component of the policy expression, drilling into nested functions and lambdas as needed. After that, it evaluates whichever of the top level function's arguments is not a Primitive as a policy expression using the value returned by the plugin, to get the Primitive that the top level function is comparing the other Primitive to. The English language description of the function and the returned Primitive are then used to construct a final String describing why the plugin failed.

To facilitate parsing functions to English, the `FunctionDef` in an `Env` `Binding` now includes both a short, internal-use name and a longer English description as strings for each function. The descriptions are added when creating the standard environment.

All of the current Hipcheck plugins have had their default query explanations updated to work with the current form of the English parser.

Future work for this issue will entail adding a similar parser for successful plugins and adding additional code to the English parsers to clean up the language and better handle expected special cases that currently parse into awkward sentences.

### DIFF
--- a/hipcheck/src/policy_exprs/mod.rs
+++ b/hipcheck/src/policy_exprs/mod.rs
@@ -75,18 +75,17 @@ impl FromStr for Expr {
 }
 
 /// Evaluates `deke` expressions.
-#[cfg(test)]
 pub struct Executor {
 	env: Env<'static>,
 }
 
-#[cfg(test)]
 impl Executor {
 	/// Create an `Executor` with the standard set of functions defined.
 	pub fn std() -> Self {
 		Executor { env: Env::std() }
 	}
 
+	#[cfg(test)]
 	/// Run a `deke` program.
 	pub fn run(&self, raw_program: &str, context: &Value) -> Result<bool> {
 		match self.parse_and_eval(raw_program, context)? {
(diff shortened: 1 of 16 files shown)


plugins/typo/src/main.rs
```diff
@@ -144,8 +144,7 @@ impl Plugin for TypoPlugin {
 
 	fn explain_default_query(&self) -> Result<Option<String>> {
 		Ok(Some(
-			"Returns whether each of the repository's package dependencies has a typo in its name"
-				.to_string(),
+			"the repository's dependecies flagged as typos".to_string(),
```
j-lanson, line 147: *dependencies

hipcheck/src/report/mod.rs
```diff
@@ -318,6 +332,16 @@ impl Analysis {
...
 	}
+
+	pub fn failing_explanation(&self) -> Result<String> {
+		let input = &self.policy_expr;
+		// Split the explanation message into components on commas, and reverse the order so it can serve as a stack
+		let message = &self.message;
+		let value = &self.value;
+		Ok(policy_exprs::parse_failing_expr_to_english(
+			input, message, value,
+		)?)
```
j-lanson, line 343: If `parse_failing_expr_to_english()` fails, we should just return the policy expression or do whatever we were doing before. With this logic, we crash if the function fails.

hipcheck/src/policy_exprs/mod.rs
```diff
@@ -156,6 +155,212 @@ impl ExprMutator for Env<'_> {
 	}
 }
 
+/// Return an English language explanation for what a failing plugin was expected to see and what it saw instead
+pub fn parse_failing_expr_to_english(
```
j-lanson, line 159: With `(eq #t $)` I get `Expected true to be equal to 'Does the target repo do fuzzing?' but it was false`. Can we do something to detect which side of the expression has the JSON value and put that one first in the output message so this kind of a message makes more sense?

plugins/entropy/src/main.rs
```diff
@@ -210,7 +210,7 @@ impl Plugin for EntropyPlugin {
 
 	fn explain_default_query(&self) -> Result<Option<String>> {
 		Ok(Some(
-			"The entropy calculation of each commit in a repo".to_owned(),
+			"The entropy calculation of each commit in the repository".to_owned(),
```
j-lanson, line 213: Should make `The` lowercase to match style of other messages

hipcheck/src/policy_exprs/mod.rs
mchernicoff: I was thinking of doing that in a later PR, but I can do it for this one. We already check for where the JSON value is, so should not be hard to flip the order when that happens.

hipcheck/src/policy_exprs/mod.rs
```diff
@@ -156,6 +155,230 @@ impl ExprMutator for Env<'_> {
...
+				.map(|p| self.visit_primitive(p).unwrap())
+				.collect::<Vec<String>>()
+				.join(", "),
+			true => {
+				let mut english_elts = elts[..max_length]
+					.iter()
+					.map(|p| self.visit_primitive(p).unwrap())
+					.collect::<Vec<String>>()
+					.join(", ");
+				english_elts.push_str(". . .");
```
j-lanson, line 369: I'd think it would look better without spaces like "..."

j-lanson (approved):
Looks good as an initial implementation!

PR #817, opened 2025-01-22
fix(setup): write configuration files directly from memory

### LINKED ISSUE #782: Embed default config files into `hc` binary, make `hc setup` write them to the config folder
Update the `hc` binary to include strings of the basic config files in `config/`, e.g. Orgs.kdl, Binary.kdl, etc. 

Update `hc setup` command to write these files from the binary to the resolved config folder, thus removing the need to fetch them from the internet.

KirilldogU:
During hc setup write config files from memory, without local dirs or remote source.

Resolves #782

### DIFF
--- a/hipcheck/src/setup.rs
+++ b/hipcheck/src/setup.rs
@@ -1,182 +1,35 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::{cli::SetupArgs, error::Result, hc_error, util::http::agent};
-use regex::Regex;
-use std::{
-	fs::File,
-	path::{Path, PathBuf},
-	sync::OnceLock,
-};
-use tar::Archive;
-use xz2::read::XzDecoder;
+use crate::error::Result;
+use std::io::Write;
+use std::{fs::File, path::Path};
 
-static R_HC_SOURCE: OnceLock<Regex> = OnceLock::new();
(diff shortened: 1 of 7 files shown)


hipcheck/src/setup.rs
```diff
@@ -1,182 +1,37 @@
...
+use crate::error::Result;
+use std::io::Write;
+use std::{fs::File, path::Path};
+
+static BINARY_TOML: &str = include_str!("../../config/Binary.toml");
+static EXEC_KDL: &str = include_str!("../../config/Exec.kdl");
+static HIPCHECK_KDL: &str = include_str!("../../config/Hipcheck.kdl");
+static HIPCHECK_TOML: &str = include_str!("../../config/Hipcheck.toml");
+static LANGS_TOML: &str = include_str!("../../config/Langs.toml");
+static LOCAL_HIPCHECK_KDL: &str = include_str!("../../config/local.Hipcheck.kdl");
```
mchernicoff, line 12: Do we want a `local.Hipcheck.kdl` file in everyone's version of Hipcheck, or is that just a developer tool we are using?

j-lanson: Good point, we should not have `local.Hipcheck.kdl` included

KirilldogU: Thanks for input, it's removed!

j-lanson (approved):

PR #818, opened 2025-01-22
feat: implemented delete

ninaagrawal:
Implemented the delete function.

### DIFF
--- a/hipcheck/src/cache/plugin.rs
+++ b/hipcheck/src/cache/plugin.rs
@@ -1,26 +1,32 @@
 // SPDX-License-Identifier: Apache-2.0
-#![allow(unused)]
-use crate::{cache::repo, error::Result, plugin::PluginId, StdResult};
+#![allow(unused)] // for ease of testing
+use crate::{
+	cache::repo,
+	error::Result,
+	hc_error,
+	plugin::{PluginId, PluginName, PluginPublisher, PluginVersion},
+	StdResult,
+};
+use dialoguer::Confirm;
 use pathbuf::pathbuf;
+use regex::Regex;
+use semver::{Version, VersionReq};
 use std::{
 	borrow::Borrow,
-	env,
+	env, fmt,
 	path::{Component, Path, PathBuf},
 	time::{Duration, SystemTime},
 };
 use tabled::{Table, Tabled};
 use walkdir::{DirEntry, WalkDir};
-//use super::super::cli::CliConfig;
-use regex::Regex; //used for regex
-use semver::{Version, VersionReq};
 
-///enums used for sorting are the same as the ones used for listing repo cache entries except for
+/// enums used for sorting are the same as the ones used for listing repo cache entries except for
 /// does not include Largest
 #[derive(Debug, Clone)]
 pub enum PluginCacheSort {
-	OldestDate, //same as "Oldest" in repo.rs
+	OldestDate, // same as "Oldest" in repo.rs
 	Alpha,
-	NewestVersion, //sorts the versions of the plugins from newest to oldest unless inverted
+	NewestVersion, // sorts the versions of the plugins from newest to oldest. Ex: 0.2.1 would come before 0.2.0
 }
 
 #[derive(Debug, Clone)]
@@ -44,13 +50,13 @@ pub struct PluginCacheListScope {
 pub struct PluginCacheEntry {
 	pub publisher: String,
 	pub name: String,
-	pub version: String,
+	pub version: Version,
 	#[tabled(display_with("Self::display_modified", self), rename = "last_modified")]
 	pub modified: SystemTime,
 }
 
 impl PluginCacheEntry {
-	//copied the function from repo for simplicity
+	// copied the function from repo for simplicity
 	pub fn display_modified(&self) -> String {
 		let Ok(dur) = self.modified.duration_since(SystemTime::UNIX_EPOCH) else {
 			return "<DISPLAY_ERROR>".to_owned();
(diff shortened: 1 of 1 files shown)


alilleybrinker:
Make sure to test, it should be safe to delete plugins since on a re-run they *should* be redownloaded, and if they're not that's a bug we don't want to ship to users.

Also don't leave commented-out code in. If it's not used, delete it. If this PR is still a draft, make sure to mark it as a draft.

hipcheck/src/cache/plugin.rs
```diff
@@ -140,6 +146,17 @@ impl HcPluginCache {
...
 	}
+	///This Constructor is only meant to be used when testing the code
+	#[cfg(test)]
+	pub fn new_for_test() -> Self {
+		let plugins_path = Path::new("C:/Users/ninaagrawal/AppData/Local/hipcheck/plugins");
+		let entries: Vec<PluginCacheEntry> = HcPluginCacheIterator::new(plugins_path).collect();
+		Self {
+			path: pathbuf![plugins_path],
+			entries,
+		}
```
j-lanson, line 157: This function seems unused and should be deleted. In any case, we shouldn't include your development workstation path in the source code.

hipcheck/src/cache/plugin.rs
```diff
@@ -191,6 +208,98 @@ impl HcPluginCache {
...
+	/// TODO: needs to be tested safely
+	fn internal_delete(&mut self, entry: &PluginCacheEntry) -> Result<()> {
+		let plugin_id = PluginId::new(
+			PluginPublisher(entry.publisher.to_string()),
+			PluginName(entry.name.to_string()),
+			PluginVersion(entry.version.to_string()),
+		);
+		let path_to_plugin = self.plugin_download_dir(&plugin_id);
+		let full_path = pathbuf![&self.path, &path_to_plugin];
+		//std::fs::remove_dir_all(full_path)?; //commenting out in order to not accidentally delete anything
```
j-lanson, line 261: This will need to be uncommented before the PR is merged

j-lanson: You can test locally by first just printing out `vull_path`, and then once you feel convinced the path is right, then delete the print and add the `remove_dir_all` line

hipcheck/src/cache/plugin.rs
```diff
@@ -260,10 +369,11 @@ impl HcPluginCache {
 		println!("{}", Table::new(to_show));
 	}
 }
-
 #[cfg(test)]
 mod tests {
 	use base64::display;
+	use clap::builder::FalseyValueParser;
+	use itertools::Itertools;
```
j-lanson, line 376: I don't see either of these imported symbols being used below. Are they necessary?

j-lanson (changes requested):
If you could, please also take care of #819 in this PR

hipcheck/src/cache/plugin.rs
ninaagrawal: I'm not sure why those were in there now that you mention it. It must be leftover code

hipcheck/src/cache/plugin.rs
```diff
@@ -88,18 +94,22 @@ impl HcPluginCacheIterator {
...
-					name.to_str().map(|s| s.to_string()) //converts the string slices into Strings
+					// extracts components separated by "/" from the path
+					name.to_str().map(|s| s.to_string()) // converts the string slices into Strings
 				} else {
 					None
 				}
 			})
-			.collect(); //collects filepath into vector
-			   //Todo: add error handling for when the path length is less than 3
+			.collect(); // collects filepath into a vector
```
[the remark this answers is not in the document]
ninaagrawal: I can fix it to be more clear, not a problem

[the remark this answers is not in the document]
j-lanson: Sorry for being unclear, I was suggesting to delete the comment because the function name basically explains what the comment explains.

hipcheck/src/cache/plugin.rs
```diff
@@ -88,18 +94,23 @@ impl HcPluginCacheIterator {
...
-					name.to_str().map(|s| s.to_string()) //converts the string slices into Strings
+					// extracts components separated by "/" from the path
+					name.to_str().map(|s| s.to_string()) // converts the string slices into Strings
 				} else {
 					None
 				}
 			})
-			.collect(); //collects filepath into vector
-			   //Todo: add error handling for when the path length is less than 3
+			.collect(); // collects filepath into a vector to make it easier to get the relevant components
```
j-lanson, line 103: I meant for this comment to be deleted

hipcheck/src/cache/plugin.rs
```diff
@@ -88,18 +94,23 @@ impl HcPluginCacheIterator {
...
+			return Err(hc_error!(
+				"Error, the path to the plugin in cache has fewer than 3 parent folders"
+			));
+		}
 		let relevant_components: &[String] = &components[components.len() - 3..];
 		let plugin_publisher = relevant_components[0].to_owned();
 		let plugin_name = relevant_components[1].to_owned();
-		let plugin_version = relevant_components[2].to_owned();
+		let plugin_version = Version::parse(relevant_components[2].to_owned().as_str())
+			.expect("Error, invalid plugin version");
```
j-lanson, line 113: `expect()` will cause a panic. This should return an `Err(hc_error!())` like above, and it should include the offending plugin and version string so the user can see what version was incorrect

j-lanson (changes requested):

j-lanson (approved):

Issue #821, opened 2025-01-24
rfd9 backwards compatibility


patrickjcasey:
The recent merge of RFD 9 into `main` has broken backwards compatibility with the `3.9.1` plugins. Investigate the root cause and apply a patch to make the changes compatible with 3.9.1 plugins.

PR #827, opened 2025-01-24
update mitre/git and mitre/affiliation for RFD9

patrickjcasey:
This PR completes the implementation of RFD9 outside of #821.

## What Changed

I experimented with using `batch_query` in `mitre/affiliation`, but on real repos, I found it was typically about 10x slower than our current implementation, due the sheer number of GRPC calls that were needed.

Based on this, I updated `mitre/affiliation` and removed the `TODO` sections around batching and also removed code related to custom chunking, as `hipcheck-common` now fully handles that. I also added a query to `mitre/git` that enables querying for all commits, as well as the author and committer of each commit. I also improved the ergonomics of checking each contributor for affiliation and reduced the amount of comparisons that are needed. I spent a few minutes benchmarking this branch with `hyperfine` and found this branch was about 5% faster when only running the affiliation plugin!

## Future Work
Working on this branch highlighted potential need for a `hipcheck-git-types` or similarly-named crate. This common crate would hold struct definitions and implementations of common traits (Serialize, Deserialize, Debug, From, ...) and would limit the amount of work that would be needed to reproduced in each of our plugins.

### DIFF
--- a/plugins/affiliation/src/main.rs
+++ b/plugins/affiliation/src/main.rs
@@ -20,7 +20,7 @@ use hipcheck_sdk::{
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use std::{
-	collections::{HashMap, HashSet},
+	collections::HashMap,
 	fmt::{self, Display, Formatter},
 	path::PathBuf,
 	result::Result as StdResult,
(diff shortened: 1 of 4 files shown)


j-lanson:
I'd like to wait to merge this until after #833 gets merged

mchernicoff (approved):

mchernicoff:
Needs a rebase, but LGTM otherwise.

PR #833, opened 2025-01-27
fix: backwards compatibility between RFD9 and prior plugins

### LINKED ISSUE #821: rfd9 backwards compatibility

j-lanson:
Resolves #821 .

Fixes some logic in hipcheck-common to support backwards compatibility in the form of communication with plugins using the pre-RF9 hipcheck-common. I have wrapped the behavior in a `rfd9-compat` feature so that it can be turned off / removed in the future if we decide backwards compatibility is no longer important.

I will spend time this week fleshing out ways to test that this is correct more concretely, but this does pass the following:
- all pre-RFD9 plugins
- all post-RFD9 plugins
- pre-RFD9 `git`, post-RFD9 `churn`
- post-RFD9 `git`, pre-RFD9 `churn`

### DIFF
--- a/hipcheck-common/Cargo.toml
+++ b/hipcheck-common/Cargo.toml
@@ -18,3 +18,7 @@ tonic = "0.12.3"
 anyhow = "1.0.95"
 pathbuf = "1.0.0"
 tonic-build = "0.12.3"
+
+[features]
+default = ["rfd9-compat"]
+rfd9-compat = []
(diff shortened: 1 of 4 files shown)


patrickjcasey:
I have been testing this branch across a few different repos with different combinations of pre/post RFD9 plugins and everything seems to be working!

patrickjcasey (approved):
Looks go to me!

PR #836, opened 2025-01-28
feat: Default to policy file at standard location (#582, #682)

### LINKED ISSUE #582: Add search for default policy file
We currently cannot automatically find a `Hipcheck.kdl` file in a default location like we did with `Hipcheck.toml`. Just need to add a way to find it.

### LINKED ISSUE #682: Replace config file with policy file as default searched-for file

cstepanian:
Resolves #582
Resolves #682

Hipcheck now defaults to using a policy file called `Hipcheck.kdl` in the default config dir.
It will fall back to loading the legacy TOML format if that fails.

If the user explicitly calls for `--policy` or `--config`, Hipcheck will only attempt that source of configuration, not falling back to the other.

Add a new `Provenance` enum for tracking whether a setting was set explicitly by a user or is from defaults.

Add a new `ConfigMode` enum to make fewer invalid states representable.

### DIFF
--- a/hipcheck/src/session/mod.rs
+++ b/hipcheck/src/session/mod.rs
@@ -2,7 +2,7 @@
 
 use crate::{
 	cache::plugin::HcPluginCache,
-	cli::Format,
+	cli::{ConfigMode, Format},
 	config::{
 		Config, ConfigSource, ConfigSourceStorage, RiskConfigQuery, RiskConfigQueryStorage,
 		WeightTreeQueryStorage,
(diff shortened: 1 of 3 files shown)


hipcheck/src/session/mod.rs
```diff
@@ -116,27 +115,25 @@ impl Session {
...
+
+					use_config(config, &mut session)?;
+				}
+			}
+			ForcePolicy { policy } => {
+				use_policy(policy, &mut session)?;
+			}
+			ForceConfig { config } => {
+				use_config(config, &mut session)?;
+			}
```
j-lanson, line 136: Could you add a `log::info` (or whatever log level you think is appropriate) statement in each match branch so we can inform the user which `ConfigMode` we derived?

j-lanson (approved):
I have one requested change, but it's minor so I'm approving. This is great work, super elegant solution to our problem.

alilleybrinker (approved):
Nice work! 🎊

PR #841, opened 2025-01-28
feat: adds config files for testing release binaries

patrickjcasey:
- adds `local.release.Hipcheck.kdl` , which is a copy of `local.Hipcheck.kdl` with paths to `local-release-plugin.kdl` instead of `local-plugin.kdl`
- adds `local-release-plugin.kdl` file to each plugin, which specifies release binary paths

The above changes make it more convenient to test the release versions of binaries

I tested these files by doing the following:

```bash
rm -rf ~/.cache/hipcheck/plugins && cargo clean # to ensure there are no cached plugins or build artifacts
cargo build --release --workspace
./target/release/hc check --policy ./config/local.release.Hipcheck.kdl https://github.com/mitre/hipcheck
```

### DIFF
--- a/config/local.release.Hipcheck.kdl
+++ b/config/local.release.Hipcheck.kdl
@@ -1,0 +1,55 @@
+
+plugins {
+    plugin "mitre/activity" version="0.2.0" manifest="plugins/activity/local-release-plugin.kdl"
+    plugin "mitre/affiliation" version="0.2.0" manifest="plugins/affiliation/local-release-plugin.kdl"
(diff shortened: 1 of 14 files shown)


j-lanson:
On the subject of clearing the cache, we've been having issues with local dev dependencies overwriting released plugins that are downloaded into the cache. Assuming the `semver` crate allows, I'm thinking we should set `version` for all `release-local-plugin.kdl`, `local-plugin.kdl`, `local.Hipcheck.kdl`, `release.local.Hipcheck.kdl` to versions `0.0.0`, so they do not overwrite published versions of plugins when in the cache

j-lanson (approved):

PR #845, opened 2025-01-28
fix: local plugins no longer overwrite released plugins

### LINKED ISSUE #844: local plugins overwriting release plugins
see https://github.com/mitre/hipcheck/pull/841 comment from @j-lanson

patrickjcasey:
- updated all local.*.kdl files to use version `0.0.0` to prevent overwriting released plugins

By doing this, I see two main benefits:
- all of the `local.*.kdl` files no longer need to be updated each time a plugin version is bumped
- cached released plugins are not overwritten with development plugins

closes #844


j-lanson (changes requested):
Don't forget the version specifiers in the `dependencies` section of `*local-plugin.kdl`

j-lanson (approved):

PR #852, opened 2025-01-29
Jlanson/config file with rel path

### LINKED ISSUE #781: Add relative file paths to `config_to_policy.rs`
We want the policy files generated by `config_to_policy.rs` to have file paths relative to the `Hipcheck.toml` file that created them, similar to what `#rel()` accomplishes in policy file parsing. This will require recording and passing along to location of `Hipcheck.toml` to the `config_to_policy.rs` entrypoint, which is not currently immediately accessible.

At the end of this, config fields like `binary-path` should take "Binary.kdl" as a path and combine it with the parent dir of the source "Hipcheck.toml" file.

j-lanson:
Resolves #781 .

This PR updates `config_to_policy.rs` to set plugin config paths as relative to the location from which `Config` was loaded. This is closer to how things used to work for config before we added the policy file conversion.

In a separate commit, I also walked back what `Config` and `config_to_policy.rs` expect as the file format for all config files except `Orgs.kdl`, since `Config` targets the remote plugins that don't have the new .kdl format yet. My plan is to walk this back once the plugins are released but before Hipcheck 3.10 is released.

### DIFF
--- a/hipcheck/src/config.rs
+++ b/hipcheck/src/config.rs
@@ -214,7 +214,7 @@ pub struct AffiliationConfig {
 #[serde(default)]
 pub struct BinaryConfig {
 	/// Binary file extension configuration file.
-	#[default = "Binary.kdl"]
+	#[default = "Binary.toml"]
 	pub binary_config_file: String,
 
 	/// Whether the analysis is active.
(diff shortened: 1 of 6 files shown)


mchernicoff (approved):

mchernicoff:
Needs a rebase, but works for me when I tested it.

PR #872, opened 2025-02-03
feat: Adds English language parsing for successful analyses

### LINKED ISSUE #837: Add English language parsing of *successful* plugins' policy expressions
Following up on Issue #383 , we should add clear English explanations of plugins that have succeeded. This will work by parsing those plugins' policy expressions in a similar manner to the failed plugins and make use of the existing `English` struct and its `impl` of `ExprVisitor`.

mchernicoff:
Resolves #837 We now attempt to parse policy expressions into English-language explanations whether the plugin's analysis succeeded or failed (but not if it errored). If a policy expression cannot be parsed, an error message is logged and the default plugin explanation text is displayed instead.

### DIFF
--- a/hipcheck/src/policy_exprs/mod.rs
+++ b/hipcheck/src/policy_exprs/mod.rs
@@ -17,6 +17,7 @@ pub use crate::policy_exprs::{
 	token::LexingError,
 };
 use env::Binding;
+use expr::FunctionDef;
 pub use expr::{parse, Primitive};
 use json_pointer::LookupJsonPointers;
 use serde_json::Value;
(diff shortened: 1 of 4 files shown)


hipcheck/src/policy_exprs/mod.rs
```diff
@@ -207,6 +210,10 @@ pub fn parse_failing_expr_to_english(
 			None => "no value was returned by the query".to_string(),
 		};
 
+		// Format the returned String depending on whether the plugin's analysis succeded or not
+		if passed {
+			return Ok(format!("Required {english_expr} and {inner_value}"));
```
j-lanson, line 215: Would it be possible to write as `"<VALUE> was <EXPECTATION>"`

mchernicoff: Made that change. Let me know what you think of the new version.

j-lanson (approved):

Issue #838, opened 2025-01-28
Improve English language parsing of known complex policy expression patterns.


mchernicoff:
We should improve the parsing of English language policy expressions that are complex but follow a known pattern.

For example, the policy expression

`(lte (count (filter (eq #t) $)) 0)`

for a failed `affiliation` plugin currently parses to

"Expected the number of elements in "equal to true" filtered on the repository's contributors flagged as affiliated to be less than or equal to 0"

but could say something like

"Expected the number of the repository's contributors flagged as affiliated to be less than or equal to 0"

And the policy expression

`(lte (divz (count (filter (eq #f) $)) (count $)) 0.05)`

for a failed `review` plugin currently parses to

"Expected the number of elements in "equal to false" filtered on commits to the repo indicating review or not divided by the number of elements in commits to the repo indicating review or not to be less than or equal to 0.05"

but could say something like

"Expected the percent of commits to the repo not indicating review to be less than or equal to 0.05"

These are suggestions, and even something in between the current and suggested expressions could work.

mchernicoff:
We could also improve the display of `Span` duration primitives. Currently they display as an internal `jiff` format e.g. 

"Expected span of time that has elapsed since last activity in repo to be less than or equal to P364D but it was PT44315963.530639277S"

mchernicoff:
> FWIW, that format is ISO 8601. Jiff does also support a ["friendly" format for durations](https://docs.rs/jiff/latest/jiff/fmt/friendly/index.html).

Thanks for the tip. I used it in MR #878 to resolve this component of the issue.
It was very simple and easy to use.

PR #885, opened 2025-02-05
feat: Improve English phrasing of count and percent policy expressions.

mchernicoff:
Finishing resolving #838 Adds simpler English phrases for the commonly occurring "count" and "percent" patterns. Changes default explanations for `churn` and `entropy` to better fit the new phrasing.

### DIFF
--- a/hipcheck/src/policy_exprs/mod.rs
+++ b/hipcheck/src/policy_exprs/mod.rs
@@ -238,11 +238,144 @@ pub struct English<'a> {
 	message: String,
 }
 
+impl English<'_> {
+	/// Returns compact English phrases for common policy expressions indicating a filtered count
+	/// These should look like `(count (filter (<operator> <primitive>) $))`
+	fn parse_count_function(&self, func: &Function) -> Option<String> {
+		// The function must have a count operator
+		if func.ident.to_string() != "count" {
+			return None;
+		}
+
+		// The first argument must be another function
+		let Expr::Function(inner_func) = &func.args[0] else {
+			return None;
+		};
+
+		// That inner function must have a filter operator
+		// Include an extra safety check that the function has a second argument
+		if inner_func.ident.to_string() != "filter" || inner_func.args.len() != 2 {
+			return None;
+		}
+
+		// The inner filter function must have a lambda and a JSON pointer as its arguments
+		let (Expr::Lambda(lambda), Expr::JsonPointer(_)) =
+			(&inner_func.args[0], &inner_func.args[1])
+		else {
+			return None;
+		};
+
+		let lambda_func = &lambda.body;
+		let lambda_arg = &lambda_func.args[1];
+
+		// Special case of `(count (filter (eq #t) $))`
+		if lambda_func.ident.to_string() == "eq" {
+			if let Expr::Primitive(Primitive::Bool(bool)) = lambda_arg {
+				if *bool {
+					return Some(format!("the number of {}", self.message.clone()));
+				}
+			}
+		}
+
+		// More general lambda expressions
+		let english_operator = get_function_def(lambda_func, &self.env).ok()?.english;
+		let english_expression = self.visit_expr(lambda_arg).ok()?;
+
+		Some(format!(
+			"the number of {} {} {}",
+			self.message.clone(),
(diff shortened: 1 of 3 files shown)


mchernicoff:
This is an example of what the newer phrasing would look like. Open to suggestions for changes.

I also realize some of the code is not the easiest to read, and I have no strong attachment to the way I did it here.

This currently only handles the special functions when they are used against a JSON pointer. I could add another case where there is something else inside the count and percent expressions (e.g. a primitive, a nested function), and parse that differently, but that seemed too complicated for the expected use cases.

hipcheck/src/policy_exprs/mod.rs
```diff
@@ -238,12 +238,132 @@ pub struct English<'a> {
...
 
+		// Check if the function is a filtered count function, and return a simple English string if it is
+		if let Some(count_string) = self.parse_count_function(func) {
+			return Ok(count_string);
+		}
+
+		// Check if the function is a percent function, and return a simple English string if it is
+		if let Some(percent_string) = self.parse_percent_function(func) {
+			return Ok(percent_string);
+		}
```
j-lanson, line 365: I would do a `match` on `func.ident` here to decide whether to call these functions. Saves setting up and tearing down a stack frame 95% of the time when we do not have a function of interest.

plugins/entropy/src/main.rs
```diff
@@ -203,7 +203,7 @@ impl Plugin for EntropyPlugin {
 
 	fn explain_default_query(&self) -> Result<Option<String>> {
 		Ok(Some(
-			"the entropy calculation of each commit in the repository".to_owned(),
+			"entropy calculations of each commit in the repository".to_owned(),
```
j-lanson, line 206: feels like both `churn` and `entropy` should start with "the" or both should not, rather than each doing it a different way.

hipcheck/src/policy_exprs/mod.rs
```diff
@@ -238,12 +238,132 @@ pub struct English<'a> {
...
+	/// These should look like `(divz (count (filter (<operator> <primitive>) $)) (count $))`
+	fn parse_percent_function(&self, func: &Function) -> Option<String> {
+		// Extra safety check that the function has a second argument
+		if func.ident.to_string() == "divz" && func.args.len() == 2 {
+			match (&func.args[0], &func.args[1]) {
+				(Expr::Function(num_func), Expr::Function(denom_func)) => {
+					if num_func.ident.to_string() == "count"
+						&& denom_func.ident.to_string() == "count"
+					{
+						match (&num_func.args[0], &denom_func.args[0]) {
```
j-lanson, line 296: This and the above function get indented far too deeply for readability. Can invert almost all the `if` and `match`es to do an early return if a condition is *not* met, thus keeping the indentation at the same level.
```
if !(func.ident.to_string() == "divz" && func.args.len() == 2) {
  return None;
}
// more logic here...
```

j-lanson (changes requested):

hipcheck/src/policy_exprs/mod.rs
```diff
@@ -238,12 +238,145 @@ pub struct English<'a> {
...
+		let lambda_arg = &lambda_func.args[1];
+
+		// Special case of `(divz (count (filter (eq #t) $)) (count $))`
+		if lambda_func.ident.to_string() == "eq" {
+			if let Expr::Primitive(Primitive::Bool(bool)) = lambda_arg {
+				if *bool {
+					return Some(format!("the percent of {}", self.message.clone()));
+				}
+			}
+		}
```
j-lanson, line 342: Is there a way the patter can work on lambdas other than `(eq #t)`. The patter is still a percentage, even if the lambda is `(gt 8)`

mchernicoff: Isn't that what the code below it does, for the other cases.

e.g. "0.005319148936170213 was the percent of churn frequencies of each commit in the repository greater than 3, which was required to be less than or equal to 0.02"

mchernicoff: Unrelated, maybe I should make a quick change to round the number of decimal points we use.

PR #887, opened 2025-02-05
chore: CI improvements

### LINKED ISSUE #688: Speed up release builds by excluding non-releasing artifacts
Right now when we run a release, it looks like `dist` is actually building the whole workspace, not just the thing being released.

I haven't yet explored how to fix it, but it's causing long release build times and lots of useless CI/CD work.

See: https://github.com/mitre/hipcheck/actions/runs/12184894132

### LINKED ISSUE #789: catch Container failures in CI
We have a [docker action that fails](https://github.com/mitre/hipcheck/actions/workflows/docker.yml), we weren't noticing it. 

Create a GitHub badge for the repo that turns red when this test fails.

### LINKED ISSUE #812: Run plugin crates tests on CI when branch changes Rust SDK
Currently, PRs with only changes to the Rust SDK do not trigger the ubuntu/mac/windows tests. While we shouldn't need to run the main Hipcheck tests, we should be testing everything in `plugins`.

### LINKED ISSUE #823: Add execution of Hipcheck using policy file and config file to CI
We keep getting bitten by `--config` based executions not working, and discovering backwards incompatibility after changes are made. Add a stage to the PR CI that runs a built Hipcheck against itself using both a policy file and a config file.

j-lanson:
Resolves #823 , Resolves #812 , Resolves #789 , Resolves #688 .

While in draft, I added a dummy commit to trigger the HC builds and tests that would not be triggered if I just updated the hipcheck.yml, so I have verified that our code running these updated actions will pass.

### DIFF
--- a/.github/workflows/hipcheck.yml
+++ b/.github/workflows/hipcheck.yml
@@ -45,7 +45,13 @@ jobs:
   test:
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        # NOTE: this list of runners is manually synced with the set used by
+        # `dist` and should be kept up to date.  we've purposefully chosen to
+        # use two different versions of macOS here to achieve ARM and x86
+        # support (macos-14 is ARM, macos-13 is x86), since alternative runners
+        # with matching versions aren't all on the free plan for GitHub
+        # Actions.
+        os: [ubuntu-20.04, windows-2019, macos-14, macos-13]
     name: "Test (${{ matrix.os }})"
     runs-on: ${{ matrix.os }}
     timeout-minutes: 15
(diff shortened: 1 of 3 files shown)


alilleybrinker (changes requested):
One nit, otherwise LGTM! 👍🏻

alilleybrinker (approved):

j-lanson:
Added the `precise-builds` line

j-lanson:
Did some local `dist host` and `dist build` commands to verify that `precise-builds = true` gives the desired behavior. Plugin tags now will only build that specific plugin instead of the entire workspace. Same for the hc core.

alilleybrinker (approved):

PR #888, opened 2025-02-05
chore(docs): update docs around building and using containerized hipcheck

### LINKED ISSUE #773: Improve documentation for using the `Containerfile` on the install page
We should improve this section: https://hipcheck.mitre.org/docs/getting-started/install/#using-the-containerfile

Specifically to reflect the need to pass through the `HC_GITHUB_TOKEN` environment variable and to specify a policy file.

An invocation with `podman` would look like:

```sh
$ podman run -e HC_GITHUB_TOKEN=$HC_GITHUB_TOKEN localhost/hc-image:latest --policy /home/hc_user/.config/hipcheck/Hipcheck.kdl check https://github.com/mitre/hipcheck
```

Assuming you've built the `Containerfile` and have it in your local image store.

j-lanson:
Resolves #773 .

The issue mentioned needing a policy file flag, but thanks to the relative file macro improvements, we can use the default config so `hc check <TGT>` works as-is. I added some more language explaining how people can move to using custom policy files in a containerized environment if they wish.

I do not have docker so I was only able to test the podman instructions, but I verified the docker instructions against the docs.


alilleybrinker (approved):

alilleybrinker:
Ugh sorry, thought GitHub's UI would do a rebase and instead it did a terrible merge commit from `main`.

alilleybrinker (changes requested):
Fixing commits.

alilleybrinker (approved):

PR #889, opened 2025-02-06
feat: consolidated kdl parsing logic into new 'hipcheck-kdl' crate

patrickjcasey:
When upgrading the `kdl` crate from 4.x to the latest 6.x version, I noticed there was considerable amounts of `kdl.rs` files copied into various plugins. The `kdl` crate versions were subtly different between them and any changes to KDL parsing code would currently need to be updated in 5 different crates. 

To ease future maintenance burden, I consolidated all of the KDL logic into  a new library crate `hipcheck-kdl`, which contains:
- common traits and functions typically used in `hipcheck` code (`ParseKdlNode`, `ToKdlNode`, `extract_data`, `string_newtype_parse_kdl_node`)
-  re-export of the `kdl` crate under the `kdl` module of `hipcheck-kdl`


alilleybrinker:
Rather than doing this, I recommend checking out `cargo-hakari`. The pattern you're doing is called the "workspace hack" in the Rust community, and is a way to enforce unification of versions of a common dependency.

https://crates.io/crates/cargo-hakari

alilleybrinker (changes requested):
See prior comment.

alilleybrinker (approved):
Discussed. We can do a workspace hack with `cargo hakari` in a follow up PR

PR #894, opened 2025-02-07
feat: Add more information to JSON output

### LINKED ISSUE #886: Augment JSON output to match expressiveness of current shell output

mchernicoff:
Resolves #886 Improves JSON output to better match current shell output and provide more information by doing the following:

-  Adding English language explanations of passing and failing analyses to the JSON
- Adds the remote repository owner, if it can be determined, to the JSON
- Adding a "final_value" field to the JSON that prints the value computed by the policy expression from the raw plugin output that is used in the top-level evaluation of the policy expression.
- Adds a new `full` option to the `-f/--format` flag. This option is identical `-f json` except that that the latter will not include the raw plugin output values in the printed JSON.


alilleybrinker (commented):
LGTM, although I think `--format full` should maybe be `--format debug` or `json-debug` or something similar. I think we ought to only have one standard stable-committed JSON format; and something that includes more information can be useful but should probably be marked as unstable and usable only for debugging.

alilleybrinker (approved):

Issue #880, opened 2025-02-04
plugin initialization error unclear


patrickjcasey:
![Image](https://github.com/user-attachments/assets/6ed9b4b2-ecb0-4b28-bf26-a3761d504ab5)

This output does not make it clear which plugin had an error being initialized. Also as a small nitpick, there should be a space or some delimiter between string and configuration

Issue #674, opened 2024-12-02
🛤️ Tracking: `hc ready` Plugin Integration


j-lanson:
Currently, `hc ready` does a set of static checks against the environment to report if `hc` is ready to run. Some of these are outdated (like checking for a `HC_GITHUB_TOKEN` env var), and a crystallized `hc ready` can't capture the amorphous requirements that may depend on what plugins we are using. For instance, `HC_GITHUB_TOKEN` is not required if we are not using the `github` plugin, but another plugin written by a 3rd party might rely on the existence of a software that we don't check for in `hc ready`.

Thus we must do the following:

- [x] #682
- [x] #582 
- [x] #675
- [x] #931
- [x] Add `--arch` flag to `hc ready` options, same as `hc check`
- [x] Update documentation to reflect new `hc ready` and emphasize config step

Once #675 is done, `hc ready` will need to retrieve/start plugins from the config file and report any configuration errors, then close the plugins.

alilleybrinker:
Updated to the 3.11.0 milestone; not planning to push further. Should be land-able by then.

cstepanian:
@j-lanson @alilleybrinker 
All the (current) TODO items are complete. Is there more left to do before `hc ready` is considered to be integrated with plugins?

j-lanson:
From my perspective, this issue can now be closed.

alilleybrinker:
Agreed! Thanks @cstepanian!

PR #895, opened 2025-02-08
feat: Improve error message context from plugins

cstepanian:
Relates to #880, in that this improves the error message, but doesn't fix the problem that plugin errors aren't attributed to the failing plugin.

- Print plugin errors with full context, like which file could not be opened.
- Clarify wording of some error messages.
- Add `miette` dependency to print helpful KDL parsing errors.

### DIFF
--- a/plugins/affiliation/src/main.rs
+++ b/plugins/affiliation/src/main.rs
@@ -50,17 +50,15 @@ impl TryFrom<RawConfig> for Config {
 		if let Some(ofv) = value.orgs_file_path {
 			// Get the Orgs file path and confirm it exists
 			let orgs_file = PathBuf::from(&ofv);
-			file::exists(&orgs_file).map_err(|_e| ConfigError::InvalidConfigValue {
-				field_name: "orgs-file".to_owned(),
-				value: ofv.clone(),
-				reason: "could not find an orgs file with that name".to_owned(),
+			file::exists(&orgs_file).map_err(|_e| ConfigError::Unspecified {
+				// Print error with Debug for full context
+				message: format!("could not find an orgs file at path {:?}", ofv),
 			})?;
 			// Parse the Orgs file and construct an OrgSpec.
 			let orgs_spec =
-				OrgSpec::load_from(&orgs_file).map_err(|e| ConfigError::InvalidConfigValue {
-					field_name: "orgs-file".to_owned(),
-					value: ofv.clone(),
-					reason: format!("Failed to load org spec: {}", e),
+				OrgSpec::load_from(&orgs_file).map_err(|e| ConfigError::Unspecified {
+					// Print error with Debug for full context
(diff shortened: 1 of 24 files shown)


plugins/affiliation/src/main.rs
```diff
@@ -50,17 +50,15 @@ impl TryFrom<RawConfig> for Config {
 		if let Some(ofv) = value.orgs_file_path {
 			// Get the Orgs file path and confirm it exists
 			let orgs_file = PathBuf::from(&ofv);
-			file::exists(&orgs_file).map_err(|_e| ConfigError::InvalidConfigValue {
-				field_name: "orgs-file".to_owned(),
-				value: ofv.clone(),
-				reason: "could not find an orgs file with that name".to_owned(),
+			file::exists(&orgs_file).map_err(|_e| ConfigError::Unspecified {
+				// Print error with Debug for full context
+				message: format!("could not find an orgs file at path {:?}", ofv),
```
j-lanson, line 55: It doesn't seem obvious to me that removing information like `field_name` would result in better clarity?

j-lanson: Or should I understand it as `InvalidConfigValue` is being replaced with a miette-formatted error specifically when a formatted file is parsed and found invalid?

j-lanson (approved):
Overall it looks good to me. It seems unintuitive to no longer include the field name when the path in a field is parsed as a file and fails, but I suppose it would be obvious from the printed path which field it was. We'll need to keep an eye on build times with miette added as a dep.

plugins/affiliation/src/main.rs
cstepanian: Ah, I forgot to explain this change in the description.
I'm changing the categorization of this error from `InvalidConfigValue`, because I don't think it makes sense to call the path "invalid" just because a file at that path did not exist. To me, a path being invalid would imply it contained slashes, null bytes, or similar (and it would be impossible for such a path to actually be used).

I made this same change in the `typo` plugin as well.

But also, the plugins `binary` and `linguist` already use `Unspecified` when they fail to read their config files.

This probably relates to the To-Do item in #674 called "Ensure that `ConfigError` enum is sufficiently expressive". We should probably have a `ConfigError` variant for `FileNotFound` or similar.

Admittedly, changing it to just `Unspecified` here is arguably making it slightly worse.

I could add the `FileNotFound` variant to actually give this a categorization.

j-lanson: Yes, I was thinking of that issue. I'm fine if you want to merge this PR and address #674 separately and add `FileNotFound` when you do

cstepanian: Sounds good. I was planning to revisit all the `Unspecified` uses later, and try to give them better categorization.

PR #896, opened 2025-02-10
feat: Track plugin name and version when reporting startup errors

### LINKED ISSUE #880: plugin initialization error unclear

cstepanian:
Resolves #880

Add plugin version to the `Plugin` struct so that it's available when starting plugins.

Also add logging when Hipcheck retrieves and starts plugins.


j-lanson (approved):
LGTM!

PR #902, opened 2025-02-10
feat: first version of listing and deleting plugins with CLI 

ninaagrawal:
CLI interface should just be CLI.

### DIFF
--- a/hipcheck/src/cache/plugin.rs
+++ b/hipcheck/src/cache/plugin.rs
@@ -26,7 +26,7 @@ use walkdir::{DirEntry, WalkDir};
 pub enum PluginCacheSort {
 	OldestDate, // same as "Oldest" in repo.rs
 	Alpha,
-	NewestVersion, // sorts the versions of the plugins from newest to oldest. Ex: 0.2.1 would come before 0.2.0
+	LatestVersion, // sorts the versions of the plugins from latest to oldest. Ex: 0.2.1 would come before 0.2.0
 }
 
 #[derive(Debug, Clone)]
@@ -109,7 +109,7 @@ impl HcPluginCacheIterator {
 		let relevant_components: &[String] = &components[components.len() - 3..];
 		let plugin_publisher = relevant_components[0].to_owned();
 		let plugin_name = relevant_components[1].to_owned();
-		let plugin_version_string = relevant_components[2].as_str(); // variable for created for convenience.
+		let plugin_version_string = &relevant_components[2]; // variable for created for convenience.
 		let plugin_version_result = Version::parse(plugin_version_string);
 		let plugin_version = match plugin_version_result {
 			Ok(version) => version,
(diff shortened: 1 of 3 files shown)


hipcheck/src/cache/plugin.rs
```diff
@@ -213,15 +213,16 @@ impl HcPluginCache {
 
 		entries.sort_by(|a1: &A, a2: &A| sort_func(a1.borrow(), a2.borrow()));
 	}
-	/// All functions up to list involve deleting entries
+	/// All functions up from delete up to list involve deleting entries
```
j-lanson, line 216: `All functions up from delete` --> `All functions from delete`

hipcheck/src/cache/plugin.rs
```diff
@@ -255,46 +256,58 @@ impl HcPluginCache {
 	}
 	/// modified internal_delete function from repo.rs
 	fn internal_delete(&mut self, entry: &PluginCacheEntry) -> Result<()> {
-		std::fs::remove_dir_all(self.get_path_to_plugin(entry))?;
+		let path_to_version = self.get_path_to_plugin(entry).0;
+		let path_to_name = self.get_path_to_plugin(entry).1; // deletes plugin folder if all versions are empty
+		std::fs::remove_dir_all(path_to_version)?;
+		let mut dir_entries = std::fs::read_dir(&path_to_name)?;
+		if dir_entries.next().is_none() {
+			std::fs::remove_dir_all(&path_to_name)?;
```
j-lanson, line 264: You can revert the changes you made to `get_path_to_plugin` below, so it goes back to returning 1 PathBuf.

Then do `std::fs::remove_dir_all(path_to_version.clone())?;` and `path_to_version.pop()` to then get the parent dir which would be deleted `if dir_entries.next().is_none() {`

hipcheck/src/main.rs
```diff
@@ -674,6 +675,42 @@ fn cmd_cache(args: CacheArgs, config: &CliConfig) -> ExitCode {
 	}
 }
 
+fn cmd_plugin_target(args: CachePluginArgs, config: &CliConfig) -> ExitCode {
```
j-lanson, line 678: call this `cmd_cache_plugin()`

hipcheck/src/cli.rs
```diff
@@ -1024,12 +1029,204 @@ impl TryFrom<Vec<String>> for RepoCacheDeleteScope {
...
+pub enum PluginOp {
+	List {
+		scope: PluginCacheListScope,
+		name: Option<String>,
+		publisher: Option<String>,
+		version: Option<VersionReq>,
+	},
+	Delete {
+		scope: PluginCacheDeleteScope,
+		name: Option<String>,
```
j-lanson, line 1060: Delete should probably include `publisher` as an optional field. You may want to restructure the logic of the filter function so you don't have to handle 8 different combinations

j-lanson (changes requested):
I like the refactor into `filter_function`

j-lanson (changes requested):

j-lanson (changes requested):
Requesting some simple fixes on doc comments.

hipcheck/src/cli.rs
```diff
@@ -869,16 +893,16 @@ pub enum CacheOp {
 
 #[derive(Debug, Clone, clap::Subcommand)]
 #[command(arg_required_else_help = true)]
-pub enum CacheSubcmds {
+pub enum CacheTargetSubcmds {
 	/// List existing caches.
 	List(CliCacheListArgs),
 	/// Delete existing caches.
```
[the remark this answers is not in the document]
ninaagrawal: Should cache plugins be list cached plugins?

j-lanson (approved):
Looks good, thanks Nino! Please do not merge this yet, I want to wait for the 3.11 blog post to be added to main.

PR #908, opened 2025-02-10
feat: Update `hc ready` to start plugins and confirm no errors

### LINKED ISSUE #675: Update `hc ready` to take an optional `--policy` flag
**note**: this is a sub-issue of #674 . there are other tasks on that to take care of first

Once #582 is implemented, we can add tiered `PolicyFile` resolution for `hc ready`. 

1. If provided with `--policy`, use that
2. If `Hipcheck.kdl` is in current working dir, use it
3. If `Hipcheck.kdl` in resolved HcConfig dir, use it

cstepanian:
Resolves #675

Add new helper `print_result` to allow `cmd_ready` to return `Result` and use `?` operator rather than explicit pattern matching on all errors.

Refactor `Session::new` to not set the target; instead that is done in a new method `Session::new_with_target`.
This has the consequence that plugins are now started before resolving the target, rather than after.

Also delete obsolete comments on `Session::new` about passing `Shell`.


mchernicoff (approved):

hipcheck/src/main.rs
```diff
@@ -559,7 +559,30 @@ fn cmd_plugin(args: PluginArgs, config: &CliConfig) -> ExitCode {
...
+	log::info!("Using configuration source: {}", config_mode);
+
+	let _session = Session::new(
+		config_mode,
+		config.cache().map(ToOwned::to_owned),
+		config.exec().map(ToOwned::to_owned),
+		config.format(),
+	)?;
+
+	log::info!("All plugins are running!");
```
j-lanson, line 584: Rather than "all plugins are running", it should probably say something like "configuration succeeded for all plugins"

j-lanson (commented):
To adjoin this PR, I'd like you to create or update the docs for `hc ready` on our site in the pre-3.11 PR

patrickjcasey:
`hc check` works as expected for me. `hc ready` also works great for me. this looks good to me!

Issue #890, opened 2025-02-06
Remove SourceQuery, add target to Session


patrickjcasey:

patrickjcasey:
closed by #910

PR #910, opened 2025-02-11
refactor: removed non-HcEngine salsa usage

patrickjcasey:
- removed all salsa usage that was not related to query caching
- added SessionBuilder struct to make it more ergonomic to create a Session

After this PR, attempting to migrate to the latest version of `salsa` should be much more manageable


patrickjcasey:
closes #893
closes #890

alilleybrinker (approved):
Exciting stuff! Thanks Patrick 🚀

PR #914, opened 2025-02-12
fix: forward plugin logging output to Hipcheck core stderr/stdout

### LINKED ISSUE #381: Proper logging for plugins
Currently in `plugin/manager.rs::start_plugin()`, we directly pipe plugin stdout/stderr to Hipcheck stdout/stderr. We should have a more structured logging that integrates with the existing logging in Hipcheck, and also specifies which plugin the message came from.

KirilldogU:
Resolves #381.
This PR forwards the logging output of plugins to the stderr/stdout output of Hipcheck core. 
It does this through:
- Creating a feature in hipcheck_sdk to initialize an env_logger
- Calling the init_logger() feature from each plugin
- When plugins are launched as a child process, having them inherit the Stderr/Stdout of the parent Hipcheck core process.

It also adds docs in the sdk on how to use the init_logger() feature.

### DIFF
--- a/sdk/rust/src/server.rs
+++ b/sdk/rust/src/server.rs
@@ -1,9 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
-	engine::HcSessionSocket,
-	error::{Error, Result},
-	Plugin, QuerySchema,
+	engine::HcSessionSocket, error::{Error, Result}, Plugin, QuerySchema
 };
 use hipcheck_common::proto::{
 	plugin_service_server::{PluginService, PluginServiceServer},
@@ -32,6 +30,10 @@ pub struct PluginServer<P> {
 impl<P: Plugin> PluginServer<P> {
 	/// Create a new plugin server for the provided plugin.
 	pub fn register(plugin: P) -> PluginServer<P> {
+		#[cfg(feature = "init_logger")]
+		{
+			crate::init_logger();
+		}
 		PluginServer {
 			plugin: Arc::new(plugin),
 		}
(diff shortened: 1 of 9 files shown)


KirilldogU:
Would appreciate input on:
- Is there a better place than main() for the init_logger() feature to be called in the plugins?
- Is any of the added documentation too wordy or out of place? Specifically, the edit to the rust-sdk.md I am uncertain about as it is not essential to getting started with creating a plugin but it does relate to the features of the SDK.

j-lanson:
My intuition would be to move the logger initialization into `PluginServer::register()` if the feature is enabled

j-lanson:
Can you state the steps you use to get plugin logs to display? I don't see anything related to the plugins when I used `HC_LOG=debug`

j-lanson:
Also, please change `fix:` to `feat:`

j-lanson:
Currently in the debug output the plugins appear as `[2025-02-12T20:43:51Z INFO  affiliation]`. Would it be possible to have them appear as `[2025-02-12T20:43:51Z INFO  plugin::affiliation]` so that we can control all the plugins' logging at once if we want? (e.g. `HC_LOG=plugin=info`)

sdk/rust/src/server.rs
```diff
@@ -32,6 +30,7 @@ pub struct PluginServer<P> {
 impl<P: Plugin> PluginServer<P> {
 	/// Create a new plugin server for the provided plugin.
 	pub fn register(plugin: P) -> PluginServer<P> {
+		init_logger();
```
j-lanson, line 33: Unless `init_logger` already has idempotency, I might suggest a mechanism like a `OnceLock` to ensure we only initialize the logger once even if `register()` happens to be called multiple times (it shouldn't but we don't have a mechanism to prevent someone doing it)

KirilldogU:
This PR is made unnecessary by commit `4503bef`.

PR #918, opened 2025-02-13
feat: Enable merge queues for CI actions.

alilleybrinker:
Make all current PR-enabled CI actions also run within merge groups. This is part of introducing merge queues to the Hipcheck CI process, which should help ensure we avoid breakage where changes pass on a branch in CI, but then fail post-merge because of how the change interacts with changes on `main` since the feature branch for the PR diverged.

### DIFF
--- a/.github/workflows/commits.yml
+++ b/.github/workflows/commits.yml
@@ -3,6 +3,8 @@ name: Commit Checks
 on:
   pull_request:
     branches: [main]
+  merge_group:
+    types: [checks_requested]
 
 permissions:
   contents: read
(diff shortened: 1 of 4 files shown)


alilleybrinker:
Okay, weirdly this shows as passing CI overall despite a bunch of failures with `merge_group` not being a valid trigger. I'll have to explore.

alilleybrinker:
Okay, it was a typo in my trigger name 🤦🏻

alilleybrinker:
Also amended to not edit the `dist init` generated action, as `dist` flags that edit in CI.

PR #919, opened 2025-02-14
feat: migrated to latest salsa v3 code

patrickjcasey:
## What Changed
- migrated from last `salsa` code on crates.io (0.16.1) to latest `salsa v3` code in https://github.com/salsa-rs/salsa
- moved engine.rs code into new engine module
- split out salsa caching logic into new file in `engine` module (`salsa_cache.rs`)
- all `salsa` events are logged at the `debug` level
- using `hipcheck-v<CURRENT_VERSION>` convention for tagging `mitre/salsa` to keep track of what commit of `salsa` is being used by `hipcheck`

## Notes about migrating to latest version of `salsa`

- `#[salsa::db]` on a struct requires the struct to also be `Clone`
- `salsa::ParallelDatabase` no longer exists in `salsa`, as it is no longer needed
- using `#[salsa::tracked]` to mark a function for tracking via `salsa` is the new way to track functions. 
- `#[salsa::input]` is used to track input to `#[salsa::tracked]` functions and requires types to be wrapped in a newtype pattern where the newtype has `#[salsa::input]` applied to the struct
- `#[salsa::query_group(HcEngineStorage)]` -like patterns are no longer necessary
- `salsa::Database` now requires the implementation of `fn salsa_event(&self, event: &dyn Fn() -> salsa::Event)`. This is where any logging of `salsa` events occurs
- the documentation for the repo is still for `salsa 2022` and has not been updated for the latest code

Here are some links that I found helpful when trying to integrate with the latest `salsa` code
- https://hackmd.io/@salsa/ry5ZGyBiA
- https://hackmd.io/7f8isuLzQ7CA1GiPqNDYfA
- https://hackmd.io/z5M6WYHFQ6qKxYF8Wwjv_w
- https://salsa.zulipchat.com/#narrow/search/zalsa_db

### DIFF
--- a/hipcheck/src/session/mod.rs
+++ b/hipcheck/src/session/mod.rs
@@ -4,7 +4,7 @@ use crate::{
 	cache::plugin::HcPluginCache,
 	cli::{ConfigMode, Format},
 	config::Config,
-	engine::{start_plugins, HcEngine, HcEngineStorage, HcPluginCore},
+	engine::{start_plugins, HcPluginCore, PluginCore},
 	error::{Context as _, Error, Result},
 	exec::ExecConfig,
 	hc_error,
@@ -119,16 +119,15 @@ impl SessionBuilder {
 			None => return Err(hc_error!("Missing HcPluginCore")),
 		};
 
-		let mut session = Session {
+		let session = Session {
 			storage: Default::default(),
 			target,
 			policy_file,
 			exec_config,
 			started_at: self.started_at,
 			format: self.format,
+			core,
 		};
-		// ensure core is set, needed by salsa
-		session.set_core(core);
 		Ok(session)
 	}
 
@@ -159,7 +158,8 @@ impl SessionBuilder {
 }
 
 /// Immutable configuration and base data for a run of Hipcheck.
-#[salsa::database(HcEngineStorage)]
+#[salsa::db]
+#[derive(Clone)]
 pub struct Session {
 	/// Query storage (used by salsa)
 	storage: salsa::Storage<Self>,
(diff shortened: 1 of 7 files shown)


j-lanson:
Looks good Patrick, but you'll need to fix a merge conflict introduced by Cal's branch being merged into main.
(only fair because your refactor of `Session` with a builder caused Cal to have to refactor as well)

alilleybrinker:
Very excited to be getting onto the latest Salsa version! :tada:

patrickjcasey:
resolved the merge conflict, just running through CI

patrickjcasey:
CI is all green :partying_face:

hipcheck/src/session/mod.rs
```diff
@@ -119,16 +119,15 @@ impl SessionBuilder {
...
 			storage: Default::default(),
 			target,
 			policy_file,
 			exec_config,
 			started_at: self.started_at,
 			format: self.format,
+			core,
 		};
-		// ensure core is set, needed by salsa
-		session.set_core(core);
```
j-lanson, line 131: I can't imagine that `ReadySession` doesn't also need the core

j-lanson (changes requested):

hipcheck/src/session/mod.rs
patrickjcasey: core is added to Session on line 129. The `set_core` function was a leftover from salsa v1 that required use of salsa v1 setters when core was being tracked by salsa

j-lanson: I was asking about `ReadySession` in particular. But now I see the plugins get started in `setup_base_session()`, which is all `ReadySession` cares about so it does not keep arount the `core` used to start the plugins.

j-lanson (approved):

PR #921, opened 2025-02-14
fix: Stop site deploy on merge queue.

alilleybrinker:
A prior fix mistakenly added site deploys to run on merge queues. This fixes that by removing the workflow config item for it.


j-lanson (dismissed):

patrickjcasey (dismissed):
LGTM

PR #930, opened 2025-02-17
Responsive layout

Mariazeigler:
Fixed padding issue when window is resized

### DIFF
--- a/site/templates/bases/base.tera.html
+++ b/site/templates/bases/base.tera.html
@@ -8,7 +8,7 @@
         <title>{% block title %}Hipcheck{% endblock %}</title>
         {% include "partials/head.tera.html" %}
     </head>
-    <body class="bg-white dark:bg-neutral-900 {% block body_classes %}{% endblock %}">
+    <body class="bg-white dark:bg-neutral-900 {% block body_classes %}{% endblock %} px-4 md:px-8 lg:px-16">
         <div class="">
             <div class="">
                 {% include "partials/nav.tera.html" %}


alilleybrinker:
Hi Maria, I'm not sure what `my-project` is, but it should be removed from the commit history here. You should also squash the changes down to one commit. As-is, this commit would be one of the largest in the history of the Hipcheck repo, with more than 105,000 new lines of code and 1,394 files impacted, mostly new files.

site/templates/bases/base.tera.html
```diff
@@ -8,7 +8,7 @@
         <title>{% block title %}Hipcheck{% endblock %}</title>
         {% include "partials/head.tera.html" %}
     </head>
-    <body class="bg-white dark:bg-neutral-900 {% block body_classes %}{% endblock %}">
+    <body class="bg-white dark:bg-neutral-900 {% block body_classes %}{% endblock %} px-4 md:px-8 lg:px-16">
```
alilleybrinker, line 11: Hm, I don't think this is quite the right fix. If you look at the homepage, you'll see the navigation bar at the top is squeezed inward horizontally, and the blue "values" section, which previously spanned the width of the page, is now boxed in. I think these padding styles should only be applied to the section with the intro text and diagram.

alilleybrinker (changes requested):
One comment, plus a few on polishing the PR. To merge you'll need to:

- Squash your commits down to one commit.
- Modify the commit message to follow the Conventional Commits spec, probably a message like "fix: Correct padding at width changes on index"
- Include a `Signed-off-by` entry in your commit by passing the `--signoff` or `-s` flag to `git commit` (can also modify the _last_ commit with `git commit --amend --signoff`).

PR #937, opened 2025-02-19
feat: Prints human output directly from Report

### LINKED ISSUE #909: Add mini-jinja template that takes the JSON and produces the shell output

mchernicoff:
Resolves #909  

Replaces the old method of printing human readable output to the shell with a new one that parses the existing `Report` struct (which is what we serialize and print as the JSON output) using MiniJinja and prints a single report String.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2186,6 +2186,8 @@ dependencies = [
  "logos",
  "maplit",
  "miette 7.4.0",
+ "minijinja",
+ "minijinja-contrib",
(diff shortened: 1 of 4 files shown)


j-lanson (approved):
any aspects of this you know of that diverge from the existing output format? Overall looks great, love seeing the LOC decrease.

mchernicoff:
I think it matches the old output exactly. I did not add the owner information (when we have it) that is now in the JSON.

PR #942, opened 2025-02-20
feat: Added Semver Version and Versionreq support for plugin versions

devin-b-lake:


devin-b-lake:
The `pre_process_plugin_version_req()` function to prepend a `=` to a version if necessary is written in two places: one in policy_file.rs and one in plugin_manifest.rs. This may be worth putting in the `util` directory. However, I didn't see a file in there where this function would be a good fit and could be added. Should I create a new file for it in `util`?

j-lanson:
> The policy-file.md documentation file where I updated the plugin version information includes the line 'For an example of the manifest field, see [@todo - link to For-Developers section].'
> 
> I looked for the 'For-Developers' section and references to `manifest` field in the docs and did not see an example/explanation for the `manifest` field within the policy file to link to. I did see a reference to the `manifest` field within the plugin manifest in the [Packing and Release a Plugin](https://hipcheck.mitre.org/docs/guide/making-plugins/release/) section. Should I link this TODO to that section?

Sure Devin that would be helpful.

hipcheck/src/plugin/plugin_manifest.rs
```diff
@@ -165,6 +251,18 @@ impl ParseKdlNode for PluginDependency {
 	}
 }
 
+// Pre-process the plugin version requirement string, then instantiate and return a Result containing a PluginVersionReq.
+// If the version_req string is not in the expected Semver Version Requirement format, a "=" will be prepended to it to
+// make the Semver VersionReq object treat the version as an exact version requirement
+fn pre_process_plugin_version_req(version_req: &str) -> Result<PluginVersionReq, Error> {
+	let version = Version::parse(version_req);
+	if version.is_ok() {
```
j-lanson, line 259: since you don't use `version` again you could just do `if Version::parse(version_req).is_ok()`

hipcheck/src/policy/policy_file.rs
```diff
@@ -641,3 +661,70 @@ impl ParseKdlNode for PolicyPatchList {
...
+			VersionReq::parse("2.*").unwrap()
+		);
+
+		let version_req = "3.1.*";
+		let processed_version_req = pre_process_plugin_version_req(version_req).unwrap();
+		assert_eq!(
+			processed_version_req.version_req,
+			VersionReq::parse("3.1.*").unwrap()
+		);
+	}
```
j-lanson, line 729: You don't need most of these tests, just one for a concrete version and one for a range specifier. I would also add one where the range specifier is more complex like `">=1.2.3, <1.8.0"`

site/content/docs/guide/config/policy-file.md
```diff
@@ -75,6 +75,13 @@ to the plugin's download manifest. For an example of the manifest field, see
 registry is established, the manifest field will become optional. In the
 immediate term it will be practically required.
 
+The `version` field adheres to [SemVer](https://semver.org/) version requirement syntax.
+In the policy file example above, the version requirement `^0.1` specifies that any plugin
+version between `0.1.0` (inclusive) and `0.2.0` (exclusive) is acceptable. *NOTE:* If an exact
+version is specified in the `version` field, e.g., `0.2.1`, then a plugin with that specific
+version is required. This exact version handling deviates from how some package managers such
+as [Cargo](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) specify dependencies.
```
j-lanson, line 83: Excellent writing.

hipcheck/src/plugin/retrieval.rs
```diff
@@ -93,41 +100,88 @@ fn retrieve_plugin(
...
+		.is_some_and(|comp| comp.op == semver::Op::Exact)
+	{
+		let version_req_syntax_string = plugin_id.version().version_req.to_string();
+		let version_syntax_string = version_req_syntax_string
+			.chars()
+			.skip(1)
+			.collect::<String>();
+		let plugin_version = PluginVersion::new(version_syntax_string.as_str())?;
+		let plugin_id_for_cache = PluginId::new(
+			plugin_id.publisher().clone(),
```
j-lanson, line 123: Really clean, I like this. Might want to check though if this code stands up to a version req like `=0.3.1,>4.0.1`. If `VersionReq::parse` doesn't catch a string like that as being nonsense, the code here might not get a proper concrete version number

j-lanson (changes requested):
A few changes to make, but overall looking very clean.

hipcheck/src/policy/policy_file.rs
```diff
@@ -112,6 +120,18 @@ impl ParseKdlNode for PolicyPlugin {
...
+// make the Semver VersionReq object treat the version as an exact version requirement
+fn pre_process_plugin_version_req(version_req: &str) -> Result<PluginVersionReq> {
+	let version = Version::parse(version_req);
+	if version.is_ok() {
+		PluginVersionReq::new(&format!("={}", version_req))
+	} else {
+		PluginVersionReq::new(version_req)
+	}
+}
+
```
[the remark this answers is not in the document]
devin-b-lake: Copy that. Should I put it in the `util` dir? If so, I didn't see a file in there where this function would be a good fit and could be added. Should I create a new file for it in `util`?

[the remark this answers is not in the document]
alilleybrinker: Calling it `util/version.rs` should be fine.

site/content/docs/guide/config/policy-file.md
alilleybrinker: It's clear, but I think it's slightly wrong. SemVer only specifies a scheme for writing version strings, not a scheme for specifying version requirements. So it's not correct to say we follow SemVer version requirement syntax, there is no such thing.

Instead it's that we follow SemVer syntax for version strings, and add on support for requirement syntax, with version prefaced with requirement operators and multiple requirements permitted, separated by commas.

alilleybrinker (changes requested):
Thanks for all the work on this @devin-b-lake! I added my comments under points Julian had made already. I'd say the biggest thing is to be clear in our documentation and naming in the code itself where we're dealing with versions (as specified by SemVer) vs. version requirements.

hipcheck/src/plugin/retrieval.rs
devin-b-lake: If a nonsense version req is passed in, e.g., `=0.3.1, >4.0.1`, how do we want to handle it? Should it error or should we assign a version? If we assign a version, should we make it the latest available plugin version?

It looks like the SemVer VersionReq doesn't detect `=0.3.1,>4.0.1` is nonsense; it creates a VersionReq object with that version range. I'll have to rethink this logic.

alilleybrinker: Nonsense reqs should be rejected.

devin-b-lake: Copy that thanks. Currently, the program errors when a nonsense version req is passed into the `version` field because it cannot find a version in the download manifest within the nonsense range.

If we want the nonsense version req to be rejected when the `PluginVersionReq` object is instantiated (`PluginVersionReq` is a wrapper for `VersionReq`), I can try to write a validation function into the `PluginVersionReq` struct.

Which approach should we go with?

alilleybrinker: I think erroring as early as possible is better, and means we can give clear error messages when a constraint is not satisfiable inherently because it's expressing incompatible bounds

j-lanson: Sorry, the point of my original comment was not to get into discussion about how to detect nonsense versions. The point is that a version req could have multiple parts, in which case if the first comparator is `=`, we can't necessarily assume that the rest of the string can be parsed as a `Version`.

j-lanson: For example, if I do this:
```
plugins {
    plugin "mitre/activity" version="=0.4.0,>0.2.0" manifest="https://hipcheck.mitre.org/dl/plugin/mitre/activity.kdl"
}
```

Hipcheck errors with: `unexpected character ',' after patch version number`

devin-b-lake: Thanks Julian, that makes sense. I created a function in the `PluginVersionReq` struct called `validate_version_req()` that handles invalid version req ranges. There is a test set at the bottom of the `plugin_manifest.rs` file.

I also moved the version_req preprocessing functionality that prepends a `=` to a version when no comparator operators (`>`, `<=`, etc.) are provided, to the `validate_version_req()` function.

These updates are up and ready for review.

j-lanson: @alilleybrinker it really doesn't feel like we should be responsible for validating user-provided version requirements. They will be caught when no matching version is found. Cargo doesn't tell you before it checks versions that your version requirements don't make sense.

site/content/docs/guide/config/policy-file.md
```diff
@@ -74,10 +74,19 @@ This section defines the plugins that will be used to run the analyses
...
 
+The `version` field adheres to [SemVer](https://semver.org/) syntax for version strings. We've
+provided additional support for requirement syntax, with version prefaced with requirement operators
+(e.g., `^0.2`) and multiple requirements permitted, separated by commas (e.g., `>=1.2.5, <1.5.0`).
+In the policy file example above, the version requirement `^0.1` specifies that any plugin
+version between `0.1.0` (inclusive) and `0.2.0` (exclusive) is acceptable. *NOTE:* If an exact
+version is specified in the `version` field, e.g., `0.2.1`, then a plugin with that specific
+version is required. This exact version handling deviates from how some package managers such
+as [Cargo](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) specify dependencies.
+
```
j-lanson, line 89: I would say that the operators are the same as those offered by Cargo, and just that the raw exact version handling is different. Also, this paragraph doesn't match the 80-char max width of the other lines in this file.

j-lanson (changes requested):

hipcheck/src/plugin/retrieval.rs
devin-b-lake: I removed the validation logic for nonsense version requirement inputs. I put it in a file on my machine in case we want to add it back in later.

j-lanson (dismissed):

hipcheck/src/policy/tests.rs
```diff
@@ -272,8 +272,8 @@ mod test {
 	#[test]
 	fn test_parse_policy_file() {
 		let data = r#"plugins {
-            plugin "mitre/activity" version="0.1.0" manifest="https://github.com/mitre/hipcheck/blob/main/plugin/dist/mitre-activity.kdl"
-            plugin "mitre/binary" version="0.1.1" manifest="https://github.com/mitre/hipcheck/blob/main/plugin/dist/mitre-binary.kdl"
+            plugin "mitre/activity" version="^0.1" manifest="https://github.com/mitre/hipcheck/blob/main/plugin/dist/mitre-activity.kdl"
+            plugin "mitre/binary" version="^0.1.1" manifest="https://github.com/mitre/hipcheck/blob/main/plugin/dist/mitre-binary.kdl"
```
alilleybrinker, line 276: I think this is supposed to drop the patch version like you have elsewhere.

hipcheck/src/policy/tests.rs
```diff
@@ -299,7 +299,7 @@ mod test {
 		));
 		plugins.push(PolicyPlugin::new(
 			PolicyPluginName::new("mitre/binary").unwrap(),
-			PluginVersion::new("0.1.1".to_string()),
+			PluginVersionReq::new("^0.1.1").unwrap(),
```
alilleybrinker, line 302: Same, drop the patch version.

plugins/review/local-plugin.kdl
```diff
@@ -11,5 +11,5 @@ entrypoint {
 }
 
 dependencies {
-  plugin "mitre/github" version="0.0.0" manifest="./plugins/github/local-plugin.kdl"
+  plugin "mitre/github" version="^0.0.0" manifest="./plugins/github/local-plugin.kdl"
```
alilleybrinker, line 14: Drop the patch version.

hipcheck/src/policy/tests.rs
```diff
@@ -55,7 +55,7 @@ mod test {
 		));
 		expected.push(PolicyPlugin::new(
 			PolicyPluginName::new("mitre/binary").unwrap(),
-			PluginVersion::new("0.1.1".to_string()),
+			PluginVersionReq::new("^0.1.1").unwrap(),
```
alilleybrinker, line 58: Drop the patch version.

hipcheck/src/policy/tests.rs
```diff
@@ -37,15 +37,15 @@ mod test {
 	#[test]
 	fn test_parsing_plugin_list() {
 		let data = r#"plugins {
-        plugin "mitre/activity" version="0.1.0" manifest="https://github.com/mitre/hipcheck/blob/main/plugin/dist/mitre-activity.kdl"
-        plugin "mitre/binary" version="0.1.1" manifest="https://github.com/mitre/hipcheck/blob/main/plugin/dist/mitre-binary.kdl"
+        plugin "mitre/activity" version="^0.1" manifest="https://github.com/mitre/hipcheck/blob/main/plugin/dist/mitre-activity.kdl"
+        plugin "mitre/binary" version="^0.1.1" manifest="https://github.com/mitre/hipcheck/blob/main/plugin/dist/mitre-binary.kdl"
```
alilleybrinker, line 41: Drop the patch.

alilleybrinker (changes requested):

j-lanson (approved):

alilleybrinker (approved):

PR #945, opened 2025-02-20
feat: update npm plugin error types and add plugin startup spinner phase

j-lanson:


alilleybrinker (approved):
Yeah, this seems reasonable to surface to users.

PR #959, opened 2025-02-24
Support for containerized plugins

aamohd:


sdk/rust/src/server.rs
```diff
@@ -168,16 +168,25 @@ impl<P: Plugin> PluginService for PluginServer<P> {
 	) -> QueryResult<Resp<Self::InitiateQueryProtocolStream>> {
 		let rx = req.into_inner();
 		// TODO: - make channel size configurable
-		let (tx, out_rx) = mpsc::channel::<QueryResult<InitiateQueryProtocolResp>>(10);
+		let (tx, out_rx) = mpsc::channel::<QueryResult<InitiateQueryProtocolResp>>(100);
```
alilleybrinker, line 171: Why the increase to the channel size?

sdk/rust/src/server.rs
```diff
@@ -40,7 +40,7 @@ impl<P: Plugin> PluginServer<P> {
 	/// Run the plugin server on the provided port.
 	pub async fn listen(self, port: u16) -> Result<()> {
 		let service = PluginServiceServer::new(self);
-		let host = format!("127.0.0.1:{}", port).parse().unwrap();
+		let host = format!("0.0.0.0:{}", port).parse().unwrap();
```
alilleybrinker, line 43: I'm not sure this change is valid when running _outside_ of a container; did you try it? If it's not, we may need to expose an SDK setting for what host to listen on, perhaps something like:

```rust
enum Host {
    // 127.0.0.1
    Loopback,
    // 0.0.0.0.0
    Any,
    // Any other IP address.
    Other(Ipv4Addr),
}
```

test-plugins/activity-container/local-plugin.kdl
```diff
@@ -0,0 +1,15 @@
+publisher "mitre"
+name "activity-container"
+version "0.0.0"
+license "Apache-2.0"
+
+entrypoint {
+  on arch="aarch64-apple-darwin" "activity-container-deploy.sh"
+  on arch="x86_64-apple-darwin" "activity-container-deploy.sh"
+  on arch="x86_64-unknown-linux-gnu" "activity-container-deploy.sh"
+  on arch="x86_64-pc-windows-msvc" "activity-container-deploy.sh"
```
alilleybrinker, line 10: Is Windows able to run this shell script? You may need to also have a PowerShell script for Windows specifically.

test-plugins/activity-container/local-plugin.kdl
```diff
@@ -0,0 +1,15 @@
...
+}
+
+dependencies {
+  plugin "mitre/git" version="0.0.0" manifest="./plugins/git/local-plugin.kdl"
```
alilleybrinker, line 14: The version here should match the current latest `mitre/git` plugin version.

alilleybrinker (changes requested):
Overall looks good, and very exciting to have this close to landing! Some comments, but I think this is on the right track.

sdk/rust/src/server.rs
aamohd: When I ran it with a channel size of 10, I would sometimes get an error that the channel closed abruptly. I increased it by a factor of 10 but if there's another recommended size I should change it to, let me know

sdk/rust/src/server.rs
aamohd: I just tried running normally with the 0.0.0.0 address and it was successful, but I can try to rework this so it isn't at 0.0.0.0 by default

test-plugins/activity-container/local-plugin.kdl
cstepanian: @alilleybrinker I'm confused -- It looks like the other plugins' `local-plugin.kdl` and `local-release-plugin.kdl` files reference version "0.0.0" like this as well. I only see real version numbers in the `plugin.kdl` files.

j-lanson: Cal is right, we've established the prcedent that the `local-X.kdl` plugin files declare themselves as version `0.0.0`. We found that smooths out plugin development a lot. The remote plugin resolution for published plugins will not install a plugin if an entry already exists in the cache, so you can get annoying and subtle situations where your `Hipcheck.kdl` that is supposed to only target published plugins tries to use a local dev version of unpublished changes that breaks things or causes unexpected behavior.

sdk/rust/src/server.rs
j-lanson: The question is whether we should be concerned about listening on all network interfaces and potentially enabling a remote process to connect to the plugin first and surreptitiously pose as hipcheck core.

alilleybrinker: Right, 0.0.0.0 inside a container is less concerning because what really matters on the host is the port bindings that the container runtime does. But for plugins running outside of the container we really ought to use the loopback address.

sdk/rust/src/server.rs
```diff
@@ -38,13 +59,16 @@ impl<P: Plugin> PluginServer<P> {
...
 
 	/// Run the plugin server on the provided port.
-	pub async fn listen(self, port: u16) -> Result<()> {
+	pub async fn listen(self, port: u16, host: Option<Host>) -> Result<()> {
 		let service = PluginServiceServer::new(self);
-		let host = format!("127.0.0.1:{}", port).parse().unwrap();
+		let host_addr = match host {
+			Some(c) => c.to_socket_addr(port),
+			None => Host::Loopback.to_socket_addr(port)
+		};
```
j-lanson, line 67: Lets have two functions:
`fn listen_local(self, port: 16)` and `fn listen(self, host: Host, port: port)`, where `listen_local` just calls `listen(Host::Loopback, ...)`

j-lanson (changes requested):

j-lanson (changes requested):
A few final changes

j-lanson (approved):

sdk/rust/src/server.rs
```diff
@@ -168,16 +200,28 @@ impl<P: Plugin> PluginService for PluginServer<P> {
...
 
 		let cloned_plugin = self.plugin.clone();
-
+		let tx_clone = tx.clone();
 		tokio::spawn(async move {
 			let mut channel = HcSessionSocket::new(tx, rx);
 			if let Err(e) = channel.run(cloned_plugin).await {
-				panic!("Error: {e}");
+				eprintln!("Channel error: {e}");
```
alilleybrinker, line 213: This should be a logging call (`error!`) rather than an `eprintln`.

sdk/rust/src/server.rs
```diff
@@ -168,16 +200,28 @@ impl<P: Plugin> PluginService for PluginServer<P> {
...
+				if !tx_clone.is_closed() {
+					if let Err(send_err) = tx_clone
+						.send(Err(tonic::Status::internal(format!("Session error: {e}"))))
+						.await
+					{
+						eprintln!("Failed to send error through channel: {send_err}");
```
alilleybrinker, line 219: This should also be an `error!` call.

tests/test-plugins/activity-container/activity-container-deploy.sh
```diff
@@ -0,0 +1,33 @@
...
+if [[ ! -f "$IMAGE_TAR" ]]; then
+    echo "Error: Image tar file '$IMAGE_TAR' not found!"
+    exit 1
+fi
+
+
+# Check if the image is already loaded
+if ! docker images | grep -q "$IMAGE_NAME"; then
+    echo "Image '$IMAGE_NAME' not found. Loading the image..."
+    docker load -i "$IMAGE_TAR" > /dev/null 2>&1
```
alilleybrinker, line 28: We probably shouldn't throw out all errors here; would be better to let them be printed so if errors arise we can debug them without editing the script.

tests/test-plugins/activity-container/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+	Ok(time_since_last_commit.to_string())
+}
+
+#[derive(Clone, Debug)]
+struct ActivityPlugin;
+
+impl Plugin for ActivityPlugin {
+	const PUBLISHER: &'static str = "mitre";
+
+	const NAME: &'static str = "activity";
```
alilleybrinker, line 61: Rename to `activity-container` just to differentiate from the non-containerized one if, for some reason, both were running and we needed to debug issues.

tests/test-plugins/activity-container/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+mod test {
+	use super::*;
+
+	use hipcheck_sdk::types::LocalGitRepo;
+	use jiff::{Span, SpanRound, Unit};
+	use std::result::Result as StdResult;
+
+	fn repo() -> LocalGitRepo {
+		LocalGitRepo {
+			path: "/home/users/me/.cache/hipcheck/clones/github/expressjs/express/".to_string(),
```
alilleybrinker, line 119: Add a comment that this is a dummy path that doesn't need to be real for these tests.

alilleybrinker (changes requested):
Very close! Only minor nits. Ping for re-review after fixes and I'll approve.

alilleybrinker (approved):
LGTM! Nice work @aamohd! 🎊

j-lanson (approved):
Thanks Aisha!

alilleybrinker:
Ah! We're stuck in that state again where there are outdated comments on a commit which was force-pushed over, which I need to mark as resolved in order to permit this to proceed.

PR #960, opened 2025-02-24
chore: CI workflows always run, jobs do filtering

### LINKED ISSUE #957: Fix merge queue with required actions
We realized our merge queue hasn't been running tests for the past 10 days. It turns out merge queues only run if triggered by required status checks. We have some checks we want it to run (conventional commits, website-test, Hipcheck tests), but they currently do path filtering, which isn't compatible with being set as required status checks.

We need to move the filtering down from the `on` trigger level to its own step in the workflow, so that they run unconditionally but then report success if the path requirements aren't met.

j-lanson:
Resolves #957 .

Merge queues trigger jobs based on the "required jobs" section of a branch's merge protections. Unfortunately, merge queues cannot detect when a workflow is "skipped", and so expects all of the required jobs to return success. Our current workflows decide at the top-level whether to run based on the modified files in the PR, so sometimes these jobs considered "required" by the merge queue do not fire, causing a deadlock. This PR moves the filtering into a job of the workflow, so now the workflow always "runs", but the relevant job will not unless any files in the given paths are modified.

### DIFF
--- a/.github/workflows/website-test.yml
+++ b/.github/workflows/website-test.yml
@@ -13,11 +13,30 @@ permissions:
   contents: read
 
 jobs:
+  # job to run change detection
+  changes:
+    runs-on: ubuntu-latest
+    # Required permissions
+    permissions:
+      pull-requests: read
+    # Set job outputs to values from filter step
+    outputs:
+      site: ${{ steps.filter.outputs.site }}
+    steps:
+    # For pull requests it's not necessary to checkout the code
+    - uses: dorny/paths-filter@v3
+      id: filter
+      with:
+        filters: |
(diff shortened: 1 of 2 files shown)


.github/workflows/website-test.yml
alilleybrinker, line 1: Still has the `paths` filter on the `pull_request` trigger.

alilleybrinker (changes requested):
See comment.

alilleybrinker (approved):

alilleybrinker:
LGTM. I think we can merge as-is, and once merged update the repo config to set these checks to required.

PR #961, opened 2025-02-25
feat: update and sync runners (ubuntu-22.04, macos-14)

### LINKED ISSUE #924: Move Ubuntu CI from 20.04 to 22.04
GitHub is in the process of phasing out support for Ubuntu 20.04 in GitHub Actions. Their policy is that the two most recent versions of stable Ubuntu releases are available, and since Ubuntu 24.04 just became Generally Available, GitHub has now initiated their wind-down of support for it.

https://github.com/actions/runner-images/issues/11101

For us, this means two things:

- We'll need to move our testing actions to use Ubuntu 22.04.
- We'll need to explore options for getting our release action, generated by `dist`, to use 22.04.

The first one is easy; we can introduce an experimental new action that runs on Ubuntu 22.04, run that in parallel as a non-PR-blocking step, and when we're confident, remove 20.04 and just use 22.04.

The second one is harder. The release action is generated by `dist`, and during CI runs `dist` validates that we haven't edited it and errors out if we have. This means we need `dist` to generate an action with 22.04 instead of 20.04. There may be an option to configure that, or they may already be working on it. I haven't checked yet.

We do want both the release and testing actions to be in sync, so we can feel confident we won't hit surprise issues during release CI because of a platform inconsistency.

alilleybrinker:
This commit does two things:

- Updates our Ubuntu runner (used for x86 Linux builds) to `ubuntu-22.04` instead of `ubuntu-20.04`. GitHub is in the process of sunsetting their `ubuntu-20.04` image. For now, dist hasn't changed the default runner for `x86_64-unknown-linux-gnu`, so we're overriding it here.
- Syncs the runner used for Apple Silicon macOS builds across dist and regular CI testing to always use `macos-14`. Previously, dist would use `macos-13`, which is x86 macOS, and cross compile to ARM. This didn't match what we did in our own regular CI testing. Now they match, which should hopefully reduce the chance of surprise errors at release time in the future.

Closes #924

### DIFF
--- a/.github/workflows/hipcheck.yml
+++ b/.github/workflows/hipcheck.yml
@@ -29,13 +29,13 @@ jobs:
   test:
     strategy:
       matrix:
-        # NOTE: this list of runners is manually synced with the set used by
-        # `dist` and should be kept up to date.  we've purposefully chosen to
-        # use two different versions of macOS here to achieve ARM and x86
-        # support (macos-14 is ARM, macos-13 is x86), since alternative runners
-        # with matching versions aren't all on the free plan for GitHub
-        # Actions.
-        os: [ubuntu-20.04, windows-2019, macos-14, macos-13]
+        # NOTE: MUST be synced manually with runners in 'dist-workspace.toml'
+        #
+        # We've purposefully chosen to use two different versions of macOS here
+        # to achieve ARM and x86 support (macos-14 is ARM, macos-13 is x86),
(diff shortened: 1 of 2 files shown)


alilleybrinker:
@j-lanson I'm perfectly happy to wait on #960 and rebase this once that's merged.

j-lanson (approved):

Issue #12, opened 2024-04-01
Write the "Incremental Computation" RFD


alilleybrinker:
Currently, Hipcheck uses the `salsa` crate to provide "incremental computation," meaning the ability to write code which recognizes when some value has already been computed, and to reuse that computation rather than calculating it again. This kind of property is _extremely_ useful for something like Hipcheck where we want to support fairly arbitrary calculation of measurements based on data, some of which may be _expensive_ to calculate (a classic example in Hipcheck's past which motivated this was the calculation of diffs for _all_ commits in the history of a project; diffs are slow to compute and we really only want to do it once).

Salsa was chosen more than a year ago, at a time when it was still under active development. It has always been experimental, but it was in active use in the `rust-analyzer` project and we believed it would continue to improve and become more stable over time. That bet has turned out to be wrong, Salsa has made _some_ progress since we initially adopted it, but is largely defunct at this point. The `rust-analyzer` project maintains their own fork of it which is specialized to their own use cases and not reusable.

Given this, we ought to consider other options for incremental computation. These options could include:

* Use of an alternative system for incremental computation in Rust, if one exists.
* Vendoring of `salsa` to maintain and further develop it ourselves.
* Exploration of options exploiting outside database / data management solutions.

The goal of this effort would be to produce an RFD which outlines what the new system would be, and how we'd move to it.

There's a ton of good prior art in the references shared by the `salsa` project itself. There are also good and relevant talks related to the [Have You Tried Rubbing a Database On It? (HYTRADBOI)](https://www.hytradboi.com/) online conference held in 2022.

alilleybrinker:
I should add that it would also be a valid conclusion to still use upstream `salsa` if we investigate it and determine that it makes sense to keep using.

alilleybrinker:
We've ended up discussing this amongst the Hipcheck dev team and have determined to stick with Salsa and upgrade to the latest version (not published to Crates.io, so there will be some complexity with this).

PR #963, opened 2025-02-26
feat: add RFD describing support for multi-repo target

### LINKED ISSUE #962: Write RFD describing proposed changes to HC core to support multi-Target

j-lanson:
Resolves #962 .

(EDIT from Andrew: [Rendered](https://github.com/mitre/hipcheck/blob/5c1b5071741b538b19a3d2030abc01d1fe19eca7/site/content/docs/rfds/0011-lock-file-target.md))

### DIFF
--- a/site/content/docs/rfds/0011-lock-file-target.md
+++ b/site/content/docs/rfds/0011-lock-file-target.md
@@ -1,0 +1,223 @@
+---
+title: Dependency and Lock Files as Targets
+weight: 11
+slug: 0011
+extra:
+  rfd: 11
+  primary_author: Julian Lanson
+  primary_author_link: https://github.com/j-lanson
+  status: Proposed
+---
+
+# Dependency and Lock Files as Targets
+
+Currently, Hipcheck operates on a single-target-per-execution basis. However,
+this model has some friction with one aspect of Hipcheck's elevator pitch,
+namely that you could run it against all the dependencies of your codebase.
+Scripting or manually entering separate Hipcheck invocations for each of your
+project's dependencies puts extra burden on would-be Hipcheck adopters.
+Furthermore, it is computationally wasteful since each invocation would spin
+up, configure, and tear down the same plugin processes that the next invocation
+would also need. To address this, this RFD proposes to make project package and
+lock files that specify dependencies into first-class analysis targets. To do
+so, this document describes proposed changes to the internal API for `Session`
+objects and the `TargetSeed` to `Target` resolution step so that a spun-up set
+of plugins can be re-used for multiple `Target`s in series.
+
+Since projects can have many dependencies, we also propose some parallelization
+strategies. These include potential tweaks to enable multiple Hipcheck instances
+to run simultaneously, or for a single instance to analyze multiple `Targets`
+in parallel with `async` code, despite current restrictions caused by `salsa`.
+
+## TargetSeed to Target Conversion Changes
+
+One of the main API changes introduced by adding package and lock files as
+target types is that we no longer can assume each target supplied to Hipcheck
+resolves to only one repository (a single-repo target). A package file
+containing a list of dependencies would instead correspond to zero or more
+separate repositories (a multi-repo target).
+
+### Target Resolution Refactoring
+
+The proposed refactor to support this change is to rename the existing `struct
+TargetSeed` object to `SingleTargetSeed`, and create a new `enum TargetSeed`
+with variants for `Single`, and `Multi` targets. This creates space for the
+creation of `MultiTargetSeedKind` and `MultiTargetSeed` objects as shown below:
+
+```rust
+pub enum SingleTargetSeedKind {
+    LocalRepo(LocalGitRepo),
+    RemoteRepo(RemoteGitRepo),
+    Package(Package),
+    MavenPackage(MavenPackage),
+    Sbom(Sbom),
+}
+
+pub struct SingleTargetSeed {
+    pub kind: SingleTargetSeedKind,
+    pub refspec: Option<String>,
+    pub specifier: String,
+}
+
+pub enum MultiTargetSeedKind {
+    CargoToml(PathBuf),
+    GoMod(PathBuf),
+    PackageJson(PathBuf),
+}
+
+pub struct MultiTargetSeed {
+    pub kind: MultiTargetSeedKind,
+    pub specifier: String,
+}
+
+pub enum TargetSeed {
+    Single(SingleTargetSeed),
+    Multi(MultiTargetSeed),
+}
+```
+
+We will update `TargetResolver::resolve()` from taking a `TargetSeed` to a
+`SingleTargetSeed`, to preserve the API that it can only return one `Target`
+(whereas taking a `TargetSeed` in the above design would allow a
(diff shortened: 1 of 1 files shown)


site/content/docs/rfds/0011-lock-file-target.md
```diff
@@ -0,0 +1,128 @@
...
+The above proposed design makes the assumption that users will want to apply the
+same policy to all dependencies in a lock file. We can imagine a scenario where
+in the first execution, Hipcheck finds a handful of dependencies that should be
+investigated, but the user wants to ignore the warning or use a different policy
+file or risk threshold for those dependencies. The ability to specify multiple
+policies or slightly tweak the existing one from the CLI is something that the
+`hc check` subcommand is not well-equipped for. We probably want to avoid having
+to add global flags to the `hc check` subcommand that apply to some but not all
+`TargetSeed` variants.  Therefore, as an backup design we may instead add a new
+subcommand specifically for multi-repo targets.
```
mchernicoff, line 81: We could also have some kind of KDL or TOML file that lists dependencies the user wants to check and optionally has rules for how to check each dependency. So a "dependency policy file" or something like that. This could be used either in lieu of the repo's lock file or as a supplement to the lock file to specify variant policies for a handful of specific dependencies.

alilleybrinker:
Regarding the `ref` flag and multi-target seeds, we could add support for the single-target "VCS URL" spec from PyPI: https://pip.pypa.io/en/stable/topics/vcs-support/

Then, when resolving targets from a multi-target seed, any that point directly to Git URLs could be resolved to `git+` VCS URLs, including with a `ref` specifier if necessary. This would permit `ref` handling in the multi-target case, and is probably what we'd need to do to support things like a `Cargo.toml` with Git-dependencies anyway. Similarly, I know Go dependencies are basically Git repos, so we'd also need to do this for future `go.mod` support.

site/content/docs/rfds/0011-lock-file-target.md
```diff
@@ -0,0 +1,137 @@
...
+objects that each get a separate target-less (see above) `Session` object on
+initialization, then make synchronous calls to their own memoized `query()`
+function, with the different tasks' output finally aggregating into a
+`HashMap<(Target, Policy), Report>`. If we create a pool of tasks smaller than
+the number of targets, the existing session in a task that gets a new target
+must first clear its memoziation cache.
+
+As an alternative to intra-process parallelization with async code, we could
+spawn Hipcheck subprocesses that each communicate with their own copies of the
+plugin processes.
```
alilleybrinker, line 137: We probably would also want to produce reports as we run.

alilleybrinker (changes requested):
Left some thoughts. Happy to discuss further.

site/content/docs/rfds/0011-lock-file-target.md
```diff
@@ -0,0 +1,137 @@
...
+separate repositories (a multi-repo target).
+
+### Command-line Interface Change Considerations
+
+The simple proposed refactor to support this change is to first add new
+`TargetSeedKind` variants to represent package/lock file types. As an example,
+a user may invoke `hc` on the command-line as `hc check Cargo.toml` to
+indicate Hipcheck should run against all dependencies in that file. To make
+logic unique to single- or multi-repo targets easier to write, we probably
+Gshould rename the existing `TargetSeedKind` to `SingleTargetSeed`, add a
```
alilleybrinker, line 47: Typo: Gshould → should

site/content/docs/rfds/0011-lock-file-target.md
```diff
@@ -0,0 +1,137 @@
...
+	Multi(MultiTargetSeed)
+}
+```
+
+There are some types of existing options that do not make sense for a package
+file, such as the `--ref` flag (it is extremely unlikely you would have the same
+target refspec for every repository resolved). Additionally, there are some
+non-existent flags that would make sense for a package target and not for a
+single-repository target; for example, a list of dependencies within the package
+file on which to skip analysis.
```
alilleybrinker, line 62: I think there's more to say on refspec issues and it should probably get its own section. Basically yes, the `--ref` flag doesn't make sense for multi-target specifiers, but also some specifiers can embed version / Git ref information. For example, a `Cargo.lock` can have Git info and/or version info, and similarly if we resolve individual package URLs for deps, those can carry version info. So we should make clear that ref-specification is still possible in a multi-target context, just not with the `--ref` flag.

site/content/docs/rfds/0011-lock-file-target.md
alilleybrinker: I'd like to explicitly foreclose both of the possibilities discussed here, both the idea in the text above that we may have a separate `check` command for multi-target invocations, and the idea that we may have multiple policy files.

Julian and I have sketched out in discussion some other possibilities, and I think in general he is more deeply exploring how to handle caching of prior results (which is the key issue implicit here in the policy discussion) and re-analysis in his draft for RFD #12.

site/content/docs/rfds/0011-lock-file-target.md
```diff
@@ -0,0 +1,137 @@
...
+- An `Arc<HcPluginCore>` handle to communicate with a set of plugins.
+- A `Shell` handle it will use for all output. In parallelized contexts the
+  Shell object can refrain from printing anything to output.
+- A re-usable `fn run(target: Target, policy: Policy, clear_db: bool) ->
+  JsonValue` that returns the raw JSON report.
+
+With the above changes, the decision about output format is moved to the calling
+function, so we can, for example, decide to create a summary (for multi-target
+seeds) or a detailed human-readable report on the shell (for a single-target
+seed).
```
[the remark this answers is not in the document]
alilleybrinker: Marking this as resolved. After more thinking I like the design sketched here.

site/content/docs/rfds/0011-lock-file-target.md
alilleybrinker: Okay, remembering more of what I meant by the comment above. I think this section should have one more paragraph added explaining that when we're executing analyses in parallel we should be printing reports as they're finished. This avoids holding reports in memory and means we provide a more responsive user experience.

alilleybrinker (changes requested):

site/content/docs/rfds/0011-lock-file-target.md
j-lanson: I'd presume for multi-target we'd want to offer some kind of "summary" printing option? and/or print the full report as JSON to a target folder?

mchernicoff: Adding the option to print a JSON report to a target folder, even for a single target run, is something I brought up in one of the discussions last week. At least one Hipcheck user is going that already as part of a pipeline that includes Hipcheck. They do it now by piping the JSON output from stdout to a file, but we could make a flag to handle that ourselves. Happy to make an issue for that if we think it's a good idea.

alilleybrinker (changes requested):
Just minor fixes, ping for re-review once those are fixed and I'll approve.

alilleybrinker (approved):

PR #965, opened 2025-02-27
feat: add new 'xtask build' subcommand

alilleybrinker:
This commands makes it easier to handle whole-workspace rebuilds, doing release rebuilds by default to make testing easier.


.cargo/config.toml
```diff
@@ -1,5 +1,5 @@
 [alias]
-xtask = "run --package xtask --bin xtask --quiet --"
+xtask = "run --package xtask --bin xtask --"
```
alilleybrinker, line 2: Made `cargo xtask` no longer run quietly, as in my experience it made it less obvious whether a slowness issue was waiting for Cargo to rebuild `xtask` itself, or `xtask` being slow.

alilleybrinker:
Here's the help text, for reference:

```
Rebuild parts of the workspace

Usage: xtask build [OPTIONS]

Options:
  -v, --verbose...  Increase logging verbosity
  -q, --quiet...    Decrease logging verbosity
  -h, --help        Print help (see more with '--help')
  -V, --version     Print version

Arguments:
      --profile <PROFILE>  The build profile to use [default: release] [possible values: release, debug, dist]
  -p, --pkg <PKG>          What to build in the workspace [default: all] [possible values: all, core, sdk, plugins]

```

xtask/src/task/build.rs
```diff
@@ -0,0 +1,150 @@
...
+		BuildProfile::Release => vec![String::from("--release")],
+		BuildProfile::Debug => vec![],
+		_ => vec![String::from("--profile"), profile.to_string()],
+	}
+}
+
+// If there are any plugins being rebuilt, clear the plugin cache.
+fn should_clear_plugin_cache(pkgs: &BTreeSet<Pkg>) -> bool {
+	pkgs.intersection(&Pkg::plugins()).count() > 0
+}
```
j-lanson, line 61: I don't know that plugins being rebuilt implies the need to clear the plugin cache. Dev / local plugins always copy overtop of the existing plugin cache entry, so we always use 0.0.0 version for them.

alilleybrinker: Okay, cool. If this isn't necessary then that's great. I'd added this as a maximally conservative recommendation.

.cargo/config.toml
cstepanian: Happy to see this tweak -- I also was frustrated that there was no feedback when `xtask` needed to be recompiled before it could run.

xtask/src/task/build.rs
```diff
@@ -0,0 +1,144 @@
...
+		}
+		_ => vec.extend(pkgs_for(what)),
+	}
+}
+
+/// Get the packages for a given selection.
+fn pkgs_for(what: BuildPkg) -> Vec<Pkg> {
+	match what {
+		BuildPkg::All => {
+			unimplemented!("should never be called with 'BuildWhat::Workspace'")
```
cstepanian, line 76: Looks like this error message may have been missed during a rename of the enum variant to `BuildPkg::All`.

cstepanian:
What specifically about release builds makes testing easier? Is it just that the built executables run faster? I normally build in debug mode, because the build time frustrates me more than the run time. Though that could be just because most of my work lately has been in code that occurs before analysis, or early in the analysis process.

I'm wondering if defaulting to release builds would be surprising to folks used to `cargo build` defaulting to debug instead. It would trigger a rebuild of all dependencies and approximately double your `target` directory size, if you haven't been building with `--release` already.

alilleybrinker:
I opted to default to release builds here since I figured the goal when building to validate functionality for runtime testing includes performance characteristics that would be unreliable to assess under the `debug` build. I am planning to separately add `cargo xtask check` for just validating that it can build.

xtask/src/task/build.rs
alilleybrinker: Good catch! Pushed a fix.

alilleybrinker:
So basically my guidance would be:

- Validating that everything compiles: `cargo xtask check`
  - Runs `cargo check` under the hood, fast.
- Building so you can test runtime behavior: `cargo xtask build`
  - Runs `cargo build --release` under the hood, slower compilation, but good performance for testing the CLI.

alilleybrinker:
Eh, on thinking about it further, I don't feel strongly about this, and am fine moving the default to `debug` to match regular `cargo build`.

alilleybrinker:
Update: the `xtask check` command turned out to be easy. I renamed the existing `check` command to `validate` and updated `xtask ci` to use the new name for it.

j-lanson (approved):

Issue #933, opened 2025-02-17
🛤️ Tracking: Plugin Logging


j-lanson:
The goal is to provide some structure to plugin logging so that we can control it from the HC core. We will update plugin requirements to take a `--log-level=<DEBUG/INFO/ETC>` flag if they want to participate, and they will have to print out the logs in a JSON string format that we define, which includes the log level and parent/module. We then need to spawn a handler in HC core to listen for stderr, parse the JSON as it finds it, and forwards to the HC core logging subscriber. We first need to ensure this last part is possible, so it is the first subtask. Within HC core when we receive these messages, we can use `tracing` again to force the parent module to be `plugin::<PLUGIN_NAME>` so we can control whether the message gets printed with `HC_LOG`.
 
- [x] Investigate how we can collect stderr handles for each plugin subprocess in HC core
- [x] Replace `log` crate with `tracing` crate and subscriber in plugins through SDK
- [x] SDK tracing subscriber adds `plugin` parent, formats logs in JSON, prints to stderr, document format requirements
- [x] Introduce `--log-level=` cmdline flag for plugins and docment
- [x] Update `Command::spawn` in main to handle stderr output from plugins

PR #968, opened 2025-02-28
Switch Plugins to use Tracing crate Logging and Forward Plugin Logging to Hipcheck core stderr/stdout 

### LINKED ISSUE #381: Proper logging for plugins

KirilldogU:
Resolves #381 and #933.
This PR switches the Hipcheck plugins to use the tracing crate for logging instead of the logging crate (as will be done on Hipcheck core).
It also forwards the logging output of plugins to the stderr/stdout output of Hipcheck core.
It does this through:
- Creating a feature in hipcheck_sdk to initialize a tracing_subscriber
- Calling the init_logger() feature from each plugin
- When plugins are launched as a child process: pipe their output to a tokio task in manager.rs that deserializes the json logs, parses them, and re-emits them to Hipcheck core output.

It also adds docs in the sdk on how to use the init_logger() feature.

### DIFF
--- a/hipcheck/src/plugin/manager.rs
+++ b/hipcheck/src/plugin/manager.rs
@@ -3,13 +3,116 @@
 use crate::{
 	hc_error,
 	plugin::{try_get_bin_for_entrypoint, HcPluginClient, Plugin, PluginContext},
+	policy::policy_file::PolicyPluginName,
 	Result,
 };
 use futures::future::join_all;
-use hipcheck_common::proto::plugin_service_client::PluginServiceClient;
+use hipcheck_common::{proto::plugin_service_client::PluginServiceClient, types::LogLevel};
+use log::{debug, error, info, trace, warn};
 use rand::Rng;
-use std::{ffi::OsString, ops::Range, path::Path, process::Command};
-use tokio::time::{sleep_until, Duration, Instant};
+use serde::Deserialize;
+use std::{collections::HashMap, env, ffi::OsString, ops::Range, path::Path, sync::LazyLock};
+use tokio::process::Command;
+use tokio::{
+	io::{AsyncBufReadExt, BufReader},
+	time::{sleep_until, Duration, Instant},
+};
+
+static PLUGIN_LOG_SETTINGS: LazyLock<PluginLogLevels> =
+	LazyLock::new(PluginLogLevels::new_from_env);
+
+/// structs for deserializing plugin logs
+#[derive(Debug, Deserialize)]
+struct LogEvent {
+	level: LogLevel,
+	fields: LogFields,
+	target: String,
+}
+
+#[derive(Debug, Deserialize)]
+struct LogFields {
+	message: String,
+}
+
+pub struct PluginLogLevels {
+	plugin_wide_level: LogLevel,
+	plugin_level_map: HashMap<String, LogLevel>,
+}
+impl PluginLogLevels {
+	/// Parses HC_LOG env var to fetch log level for each plugin as Struct
+	pub fn new_from_env() -> Self {
+		// default level to error when var unset
+		let mut plugin_wide_level = LogLevel::Error;
+		let mut plugin_level_map: HashMap<String, LogLevel> = HashMap::new();
+
+		let Ok(rust_log) = env::var("HC_LOG") else {
+			return PluginLogLevels {
+				plugin_wide_level,
+				plugin_level_map,
+			};
+		};
+
+		let settings: Vec<&str> = rust_log.split(',').collect();
+		let mut plugin_wide_set = false;
+
+		for setting in settings {
+			let parts: Vec<&str> = setting.split('=').collect();
+
+			match (parts.first(), parts.get(1)) {
+				(Some(module), Some(level)) => {
+					let log_level = match LogLevel::level_from_str(level) {
+						Ok(level) => level,
+						Err(e) => {
+							log::error!("{}", e);
+							continue;
+						}
+					};
+
+					if *module == "plugin" {
+						// e.g. HC_LOG="plugin=info"
+						// plugin wide log level is set, overriding global level
+						plugin_wide_level = log_level;
+						plugin_wide_set = true;
+					} else {
+						// e.g. HC_LOG="plugin::typo=info"
+						// plugin specific log level is found
+						plugin_level_map.insert(module.to_string(), log_level);
+					}
+				}
+				(Some(level), _) => {
+					// e.g. HC_LOG="info"
+					let log_level = match LogLevel::level_from_str(level) {
+						Ok(level) => level,
+						Err(e) => {
+							log::error!("{}", e);
+							continue;
+						}
+					};
+					// Global log level used for plugins, if no plugin wide found
+					if !plugin_wide_set {
+						plugin_wide_level = log_level;
+					}
+				}
+				(None, _) => log::error!("Incomplete HC_LOG var set"),
+			}
+		}
+		PluginLogLevels {
+			plugin_wide_level,
+			plugin_level_map,
+		}
+	}
+
+	/// Retrieves the log level for a given plugin, using most specific level available
+	fn get_log_level_for_module(&self, plugin_name: &str) -> &LogLevel {
+		// use plugin specific
+		if let Some(level) = self.plugin_level_map.get(plugin_name) {
+			level
+		} else {
+			// use plugin wide
+			&self.plugin_wide_level
+		}
+	}
+}
 
 #[derive(Clone, Debug)]
 pub struct PluginExecutor {
@@ -63,6 +166,52 @@ impl PluginExecutor {
 		}
 
 		Err(hc_error!("Failed to find available port"))
+	}
+
+	/// Launches async forwarding of plugin logs to program output.
+	fn forward_plugin_logs(
+		&self,
+		reader: impl tokio::io::AsyncRead + Unpin + Send + 'static,
+		plugin_title: String,
+	) {
+		tokio::spawn(async move {
+			let reader = BufReader::new(reader);
+			let mut lines = reader.lines();
+			// parse plugin log json lines to LogEvent objects and output the messages
+			loop {
+				match lines.next_line().await {
+					Ok(Some(line)) => {
+						if let Ok(event) = serde_json::from_str::<LogEvent>(&line) {
+							let message = event.fields.message;
+							let event_target = match event.target.contains("hipcheck_sdk") {
+								true => {
+									format!("plugin::{}::{}", plugin_title, event.target)
+								}
+								false => format!("plugin::{}", event.target),
+							};
+							match event.level {
+								LogLevel::Error => error!(target: &event_target, "{}", message),
+								LogLevel::Warn => warn!(target: &event_target, "{}", message),
+								LogLevel::Info => info!(target: &event_target, "{}", message),
+								LogLevel::Debug => debug!(target: &event_target, "{}", message),
+								LogLevel::Trace => trace!(target: &event_target, "{}", message),
+								LogLevel::Off => {}
+							};
+						}
+						// drop plugin logs without message
+					}
+					Ok(None) => {
+						break;
+					}
+					Err(e) => {
+						log::error!(
+							"I/O erorr occured while listening for logs from {plugin_title}: {e}"
+						);
+						break;
+					}
+				};
+			}
+		});
 	}
 
 	pub async fn start_plugins(&self, plugins: Vec<Plugin>) -> Result<Vec<PluginContext>> {
@@ -146,14 +295,26 @@ impl PluginExecutor {
 			spawn_args.push("--port");
 			spawn_args.push(port_str.as_str());
 
+			// pass logging level for plugin
+			let plugin_title = PolicyPluginName::new(&plugin.name)
+				.expect("plugin name")
+				.name
+				.0;
+			let plugin_path = format!("plugin::{}", plugin_title);
+			let plugin_log_level = PLUGIN_LOG_SETTINGS.get_log_level_for_module(&plugin_path);
+			let plugin_log_level_str = plugin_log_level.to_string();
+			spawn_args.push("--log-level");
+			spawn_args.push(&plugin_log_level_str);
+
 			// Spawn plugin process
 			log::debug!("Spawning '{}' on port {}", &plugin.entrypoint, port_str);
 			let Ok(mut proc) = Command::new(&canon_bin_path)
 				.env("PATH", &cmd_path)
 				.args(spawn_args)
-				// @Temporary - directly forward stdout/stderr from plugin to shell
-				.stdout(std::io::stdout())
-				.stderr(std::io::stderr())
+				// pipe output to tokio task for plugin log forwarding
+				.stdout(std::process::Stdio::piped())
+				.stderr(std::process::Stdio::piped())
+				.kill_on_drop(true)
 				.spawn()
 			else {
 				spawn_attempts += 1;
@@ -190,12 +351,19 @@ impl PluginExecutor {
 			// If opt_grpc is None, we did not manage to connect to the plugin. Kill it
 			// and try again
 			let Some(grpc) = opt_grpc else {
-				if let Err(e) = proc.kill() {
-					println!("Failed to kill child process for plugin: {e}");
+				if let Err(e) = proc.kill().await {
+					log::error!("Failed to kill child process for plugin: {e}");
 				}
 				spawn_attempts += 1;
 				continue;
 			};
+
+			let plugin_stdout = proc.stdout.take().expect("Failed to capture plugin stdout");
+			let plugin_stderr = proc.stderr.take().expect("Failed to capture plugin stderr");
+
+			self.forward_plugin_logs(plugin_stdout, plugin_title.clone());
+			self.forward_plugin_logs(plugin_stderr, plugin_title.clone());
+
 			// We now have an open gRPC connection to our plugin process
 			return Ok(PluginContext {
 				plugin: plugin.clone(),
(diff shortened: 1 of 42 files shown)


hipcheck/src/plugin/manager.rs
```diff
@@ -64,6 +83,90 @@ impl PluginExecutor {
...
+					if module == plugin_name {
+						return Some(level.to_string());
+					}
+
+					// Store plugin wide level as fallback
+					if module == "plugin" {
+						plugin_wide_level = Some(level.to_string());
+					}
+				} else if parts.len() == 1 {
+					// Global log level
```
j-lanson, line 111: Calculating this again for every plugin is inefficient, we could inspect `HC_LOG` once when the `PluginExecutor` is instantiated and "save" info on the global log level / per-plugin log level.

hipcheck/src/plugin/manager.rs
```diff
@@ -7,9 +7,28 @@ use crate::{
...
+use tokio::process::Command;
+use tokio::{
+	io::{AsyncBufReadExt, BufReader},
+	time::{sleep_until, Duration, Instant},
+};
+
+/// structs for deserializing plugin logs
+#[derive(Debug, Deserialize)]
+struct LogEvent {
+	level: String,
```
j-lanson, line 23: `level` should be an enum so we don't have to do String-based matching below, and we're more protected against small differences in capitalization. We also would want to make sure our serde deserialization can handle differently-cased strings.

hipcheck/src/plugin/manager.rs
```diff
@@ -146,14 +249,29 @@ impl PluginExecutor {
 			spawn_args.push("--port");
 			spawn_args.push(port_str.as_str());
 
+			// pass logging level for plugin
+			let plugin_title = plugin
+				.name
+				.split_once('/')
+				.map(|(_, part)| part)
+				.unwrap_or(&plugin.name);
```
j-lanson, line 257: use `PolicyPluginName::new(plugin.name).name` instead of splitting out the non-publisher part here yourself.

hipcheck-sdk-macros/src/lib.rs
```diff
@@ -268,7 +268,7 @@ pub fn queries(_item: TokenStream) -> TokenStream {
 		};
 		agg.extend(out);
 	}
-	log::info!(
+	tracing::info!(
```
j-lanson, line 271: Can you reduce this to debug level?

j-lanson (changes requested):

sdk/rust/Cargo.toml
```diff
@@ -21,13 +21,17 @@ schemars = { version = "0.8.21", features = ["url"] }
...
+tracing-subscriber = { version="0.3", features=["env-filter", "json"], optional = true }
 hipcheck-common = { version = "0.3.0", path = "../../hipcheck-common" }
+console = "0.15.10"
 
 [features]
+default = ["init_logger"]
 macros = ["dep:hipcheck-sdk-macros"]
 mock_engine = []
 print-timings = []
+init_logger = ["tracing", "tracing-subscriber"]
```
j-lanson, line 34: Rename feature to `log_forwarding`

j-lanson (changes requested):

hipcheck/src/plugin/manager.rs
```diff
@@ -146,14 +249,29 @@ impl PluginExecutor {
...
+			let plugin_path = format!("plugin::{}", plugin_title);
+			let plugin_log_level = match self.get_log_level_for_module(&plugin_path) {
+				Some(level) => level,
+				None => String::from("OFF"), // Default to "OFF"
```
cstepanian, line 261: What's the motivation for defaulting to `OFF`? If I understand correctly, this suppresses all logging, even errors. (I have yet to try these changes in my environment, so I may be missing some subtlety)

If this does suppress all errors from plugins by default, I think that it would be surprising to users to not be informed when plugins fail and why, since it effects how they interpret the results, and can point to problems with their environment or source data.

This is especially unfortunate for debugging, given that currently, the only way for plugins to report errors other than by logging (now tracing) is the `UnspecifiedQueryState` error that gives no information. (Though I'm working on a proposal to improve that situation)

hipcheck/src/plugin/manager.rs
```diff
@@ -64,6 +94,90 @@ impl PluginExecutor {
 
 		Err(hc_error!("Failed to find available port"))
 	}
+	/// Retrieves the log level for a given plugin from HC_LOG env_var.
+	/// Falls back to overall plugin level or global level if not specifically set.
+	fn get_log_level_for_module(&self, plugin_name: &str) -> Option<String> {
```
cstepanian, line 99: Since this function does non-trivial parsing, it should have unit tests.

For testability, perhaps `rust_log` should be passed in as a parameter instead of accessing the environment from the function.

Also, more of a minor nitpick: does this need to be a method on `PluginExecutor`?, It does not appear to use the reference to `self` to do its work, and the results don't (and shouldn't) depend on the state of the `PluginExecutor`.

hipcheck/src/plugin/manager.rs
KirilldogU: The parsing logic is moved to a the new() method of the PluginLogLevels struct which is run once as part of a Lazylock static variable called PLUGIN_LOG_SETTINGS.

hipcheck/src/plugin/manager.rs
KirilldogU: manager.rs now utilizes an Enum called "LogLevel" which handles differently cased strings by converting them to uppercase.

sdk/rust/Cargo.toml
j-lanson: no, they are fine with unrelated names

j-lanson: my opinion is that the feature name should be designed to be clear to potential user. `log_forwarding` is more clear to the user what the feature does

KirilldogU: Sounds good, the feature is renamed.

hipcheck/src/plugin/manager.rs
KirilldogU: Unit tests implemented!
We have a function called `with_env_vars` that works for setting environmental variables that are local to a test and unaffected by running tests in parallel.
The method is also moved out PluginExecutor to be a method of its own struct called PluginLogLevels. This was partially done based on Julian's input above that it would be most efficient to run it once in a static global variable using a LazyLock.

hipcheck/src/plugin/manager.rs
KirilldogU: Based on our discussion it appears we are defaulting plugins to OFF, but looking to find ways to help plugins communicate when targets are invalid.

j-lanson:
@KirilldogU please resolve the merge conflicts. probably a rebase on `origin/main`, and unless you have a good reason, defer to the existing change on `origin/main` over your own since its mostly `Cargo.toml` files

hipcheck/src/plugin/manager.rs
```diff
@@ -3,13 +3,106 @@
...
+						// plugin wide log level
+						plugin_wide_level = level;
+						plugin_wide_set = true;
+					} else {
+						// plugin specifc log level
+						plugin_level_map.insert(module, level);
+					}
+				}
+			}
+		}
```
j-lanson, line 85: Replace `if let` with `let Ok(rust_log) = env::var("HC_LOG") else {}` to reduce indentation

hipcheck/src/plugin/manager.rs
```diff
@@ -3,13 +3,106 @@
...
+	fn get_log_level_for_module(&self, plugin_name: &str) -> &String {
+		// use plugin specific
+		if self.plugin_level_map.contains_key(plugin_name) {
+			return self
+				.plugin_level_map
+				.get(plugin_name)
+				.expect("Failed to fetch plugin log level");
+		}
+		// use plugin wide
+		&self.plugin_wide_level
```
j-lanson, line 103: why not
```
if let Some(level) = self.plugin_level_map.get(plugin_name) {
   level
} else {
  &self.plugin_wide_level
}

File: hipcheck/src/plugin/manager.rs (hunk 3 above)
Comment (j-lanson) [reply hipcheck/src/plugin/manager.rs:287]: This looks the same to me.

File: sdk/rust/src/server.rs [hunk 10]
```diff
@@ -31,7 +31,12 @@ pub struct PluginServer<P> {
 
 impl<P: Plugin> PluginServer<P> {
 	/// Create a new plugin server for the provided plugin.
-	pub fn register(plugin: P) -> PluginServer<P> {
+	pub fn register(plugin: P, log_level: &str) -> PluginServer<P> {
```
j-lanson, line 34: `log_level` should be an Enum, and probably should be `Option<LogLevel>` which is equal to `LogLevel::Error` when set to `None`. Similarly, should be this way in the `Args {}` struct

j-lanson (changes requested):

hipcheck/src/plugin/manager.rs
```diff
@@ -3,13 +3,106 @@
...
+	Trace,
+}
+
+pub struct PluginLogLevels {
+	plugin_wide_level: String,
+	plugin_level_map: HashMap<String, String>,
+}
+impl PluginLogLevels {
+	/// Parses HC_LOG env var to fetch log level for each plugin as Struct
+	pub fn new() -> Result<Self> {
```
cstepanian, line 53: I suggest calling this `new_from_env()`, `from_env()` or similar.

`new()` might generally be expected to be a pure function that just constructs a minimal valid instance of a struct. Here it's reading from the environment, which is technically IO.

This change would hopefully make it clearer to readers of any code that calls this function, that it's reading the environment at that point in the code, which might matter for debugging.

sdk/rust/src/server.rs
j-lanson: You can do `log_level: impl Into<Option<LogLevel>>` so that people can pass the raw LogLevel or None depending.

hipcheck/src/plugin/manager.rs
```diff
@@ -3,13 +3,106 @@
...
+		let mut plugin_level_map: HashMap<String, String> = HashMap::new();
+
+		if let Ok(rust_log) = env::var("HC_LOG") {
+			let settings: Vec<&str> = rust_log.split(',').collect();
+			let mut plugin_wide_set = false;
+
+			for setting in settings {
+				let parts: Vec<&str> = setting.split('=').collect();
+
+				if parts.len() == 1 {
```
cstepanian, line 64: Would you please comment this block and the `parts.len() == 2` block with an explanation of what these cases mean, with at least one example for each? I think that would help a lot, since a reader of this code isn't necessarily familiar with the syntax this code is trying to parse.

hipcheck/src/plugin/manager.rs
KirilldogU: Good catch, fixed!

hipcheck/src/plugin/manager.rs
KirilldogU: Great distinction, thanks.

hipcheck/src/plugin/manager.rs
KirilldogU: Updated to default plugins to error if no relevant HC_LOG var is set.

sdk/rust/src/server.rs
j-lanson: You may want to move `LogLevel` declaration into the `hipcheck-common` crate so both the SDK and the core can use it

hipcheck/src/plugin/manager.rs
alilleybrinker: I think a match construction here would also be clearer: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=501b8472e9c6899e95ad773bca5dd687

hipcheck/src/plugin/manager.rs
```diff
@@ -3,13 +3,106 @@
...
+	}
+
+	/// Retrieves the log level for a given plugin
+	fn get_log_level_for_module(&self, plugin_name: &str) -> &String {
+		// use plugin specific
+		if self.plugin_level_map.contains_key(plugin_name) {
+			return self
+				.plugin_level_map
+				.get(plugin_name)
+				.expect("Failed to fetch plugin log level");
```
alilleybrinker, line 100: Why `expect` rather than reporting an error out with `Result`?

hipcheck/src/plugin/manager.rs
```diff
@@ -65,6 +158,46 @@ impl PluginExecutor {
...
+	fn forward_plugin_logs(
+		&self,
+		reader: impl tokio::io::AsyncRead + Unpin + Send + 'static,
+		plugin_name: String,
+	) {
+		tokio::spawn(async move {
+			let reader = BufReader::new(reader);
+			let mut lines = reader.lines();
+			// parse plugin log json lines to LogEvent objects and output the messages
+			while let Some(line) = lines.next_line().await.unwrap() {
```
alilleybrinker, line 171: Add a safety comment for why this `unwrap` would never be hit, or replace it with error reporting.

plugins/binary/src/main.rs
```diff
@@ -140,14 +140,17 @@ struct Args {
 	#[arg(long)]
 	port: u16,
 
+	#[arg(long, default_value = "OFF")]
```
alilleybrinker, line 143: Default should be `ERROR` not `OFF`, for all plugins.

alilleybrinker (changes requested):

hipcheck/src/plugin/manager.rs
KirilldogU: This expect is removed as the function no longer has a failure condition.

hipcheck/src/plugin/manager.rs
```diff
@@ -3,13 +3,107 @@
...
+		let settings: Vec<&str> = rust_log.split(',').collect();
+		let mut plugin_wide_set = false;
+
+		for setting in settings {
+			let parts: Vec<&str> = setting.split('=').collect();
+
+			match (parts.first(), parts.get(1)) {
+				(Some(module), Some(level)) => {
+					let log_level: LogLevel =
+						LogLevel::level_from_str(level).expect("Invalid log level var");
```
cstepanian, line 64: (1/2) Since this is parsing user-typed data, there's a good chance this can fail in normal use.
So instead of panicking with `expect()`, this should return `Err()`.

hipcheck/src/plugin/manager.rs
```diff
@@ -3,13 +3,107 @@
...
+					} else {
+						// e.g. HC_LOG="plugin::typo=info"
+						// plugin specific log level is found
+						plugin_level_map.insert(module.to_string(), log_level);
+					}
+				}
+				(Some(level), _) => {
+					// e.g. HC_LOG="info"
+					let log_level: LogLevel =
+						LogLevel::level_from_str(level).expect("Invalid log level var");
```
cstepanian, line 80: (2/2) Since this is parsing user-typed data, there's a good chance this can fail in normal use.
So instead of panicking with `expect()`, this should return `Err()`.

hipcheck/src/plugin/manager.rs
```diff
@@ -3,13 +3,107 @@
...
+use serde::Deserialize;
+use std::{collections::HashMap, env, ffi::OsString, ops::Range, path::Path, sync::LazyLock};
+use tokio::process::Command;
+use tokio::{
+	io::{AsyncBufReadExt, BufReader},
+	time::{sleep_until, Duration, Instant},
+};
+
+static PLUGIN_LOG_SETTINGS: LazyLock<PluginLogLevels> =
+	LazyLock::new(|| PluginLogLevels::new_from_env().expect("Invalid log level var set"));
```
cstepanian, line 22: I'm not sure what limitations there may be with `LazyLock`, so this might be the best we can do here.

But is there a non-panicking way to handle errors here? (by passing the `Err()` variant through instead of using `expect()`)

hipcheck/src/plugin/manager.rs
cstepanian: Thanks for adding comments and refactoring to use `match`!

hipcheck/src/plugin/manager.rs
KirilldogU: @alilleybrinker I edited this line to only take and act on the successful result with Ok(), on the basis that error logs about I/O or encoding errors from plugin logs seem not that useful to the user. Would you agree or is it better to log possible errors here?

sdk/rust/src/server.rs
KirilldogU: Implemented everything as described above, with a refactor of LogLevel to hipcheck-common and editing all existing functions to use the enum instead of strings. This required implementing several methods, so that LogLevel could be serialized/deserialized to from serde_json, clap, and strings.

hipcheck/src/plugin/manager.rs
KirilldogU: @cstepanian Regarding the comments on this function, I just did an overhaul that made the function return only a Result with no error which means there are no more Expect()s and the LazyLock cannot panic now. This was mainly done because according to the existing logging/tracing crates, and invalid HC_LOG var will not cause failure of the program but just default logging. I edited the `new_from_env()` function to create a similar experience, where failure to parse a log var is logged as an error but the program continues to use default plugin log levels and does not crash.

KirilldogU: Please let me know what you think about the new implementation!

hipcheck/src/plugin/manager.rs
```diff
@@ -65,6 +168,39 @@ impl PluginExecutor {
...
+	fn forward_plugin_logs(
+		&self,
+		reader: impl tokio::io::AsyncRead + Unpin + Send + 'static,
+		plugin_title: String,
+	) {
+		tokio::spawn(async move {
+			let reader = BufReader::new(reader);
+			let mut lines = reader.lines();
+			// parse plugin log json lines to LogEvent objects and output the messages
+			while let Ok(Some(line)) = lines.next_line().await {
```
cstepanian, line 181: As I understand it, this will silently ignore the `Err()` case, which would contain a `std::io::Error` if there is an I/O error when reading from the plugin's log stream.

While we probably can't do much to respond to such an error, I wonder if it's valuable to report it by logging? It could be a useful hint that something is going wrong, though I've seen quite a few cases where it's hard to avoid I/O errors during shutdown of IPC connections. Issues would likely be platform-dependent too.

j-lanson (approved):
Looks good to me, good work Kirill!

hipcheck/src/plugin/manager.rs
j-lanson: I don't think it hurts to follow Cal's suggestion of notifying the user when I/O with the child process has failed before this async task dies. e.g.

```rust
loop {
    let line = match lines.next_line().await {
       Ok(Some(l)) => l,
       Ok(None) => {break; },
       Err(e) => {
             log::error!("I/O erorr occured while listening for logs from {plugin_title}: {e}");
             break;       
       }
    }
    ... 
}
```

j-lanson: however, if in your testing you find this error getting printed for normal shutdown, as Cal suggested might be possible, let's leave it as it was with the `while let`

KirilldogU: Implemented the case/match statement, and currently not seeing any I/O error logs.

j-lanson (approved):

PR #976, opened 2025-03-03
fix: plugins accept and parse unknown trailing args

KirilldogU:
Plugins can now accept unknown trailing arguments without failure. 
Allows for greater robustness when arguments get added.


j-lanson:
@aamohd you'll want to make this same change to your Containerized plugin

PR #977, opened 2025-03-03
fix: add missing mold dependency to dist/Containerfile

### LINKED ISSUE #975: Add `mold` as a dependency for `dist`
Modify the `dist-workspace.toml` setting for x86 to install `mold`, so we have it during release automation.

patrickjcasey:
closes #975

### DIFF
--- a/dist/Containerfile
+++ b/dist/Containerfile
@@ -8,7 +8,7 @@ WORKDIR /app
 
 RUN set -eux \
     && apt-get update \
-    && apt-get install --no-install-recommends -y ca-certificates git curl xz-utils \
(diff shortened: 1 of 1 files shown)


alilleybrinker:
Sorry, this is a good catch, but it's actually not what I mean. I don't think the containerfile needs it, it's needed in `dist-workspace.toml`.

alilleybrinker:
Update: I was wrong, it would go in the `dist.toml` for Hipcheck core and for each plugin.

j-lanson:
I'm taking care of the dist.toml

PR #978, opened 2025-03-03
feat: Add `--arch` flag to `hc ready`, as used by `hc check`

cstepanian:
Relates to #674

This allows users to override the detected architecture when `hc ready` starts plugins.

Also, `hc ready` will now return `ExitCode::FAILURE` when any of its checks fail.

Feedback requested:
- [x] Which Clap options should be set on `ReadyArgs`? I tentatively copied `subcommand_negates_reqs = true` and `arg_required_else_help = true` from `CheckArgs`, but even after reading the documentation, I'm not sure why `CheckArgs` needs these or how they interact with other parts of the CLI.

### DIFF
--- a/hipcheck/src/cli.rs
+++ b/hipcheck/src/cli.rs
@@ -460,7 +460,7 @@ pub enum FullCommands {
 	Check(CheckArgs),
 	Schema(SchemaArgs),
 	Setup,
-	Ready,
+	Ready(ReadyArgs),
 	Update(UpdateArgs),
 	CacheTarget(CacheTargetArgs),
 	CachePlugin(CachePluginArgs),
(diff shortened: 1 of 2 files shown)


hipcheck/src/cli.rs
```diff
@@ -824,6 +824,15 @@ pub enum SchemaCommand {
 	Repo,
 }
 
+#[derive(Debug, Clone, clap::Args)]
+// Cal TODO does ReadyArgs need these? (copied from CheckArgs)
+#[command(subcommand_negates_reqs = true)]
+#[command(arg_required_else_help = true)]
```
alilleybrinker, line 830: No, `hc ready` doesn't need these.

alilleybrinker (changes requested):

alilleybrinker:
Gave a review. @cstepanian does this still need to be marked draft?

cstepanian:
> Gave a review. @cstepanian does this still need to be marked draft?

I only marked it Draft to make sure nobody merged while my TODO comment was left in.

I'll fix that and mark it ready.

hipcheck/src/cli.rs
cstepanian: Removed.

alilleybrinker (approved):

PR #982, opened 2025-03-03
chore: Removed unused dependencies.

alilleybrinker:
Used 'cargo machete' to identify unused dependencies. There were a few false positives which were added to an ignore list in the appropriate 'Cargo.toml' files.


patrickjcasey (approved):
Looks good to me! I cloned it and everything was working as expected

PR #966, opened 2025-02-28
chore: Folder reorganization.

alilleybrinker:
Moved library crates under a new `library/` directory in the root of the repository, and moves `test-plugins` under a new `tests/` directory, also in the root of the repository. Updated all imports and other references to use the new paths.

PR #992, opened 2025-03-04
Introduce 'workspace-hack' crate.

alilleybrinker:
Merge after #966, keeping as "draft" until then.

NOTE: This would change our flow for publishing. See: [`cargo hakari`'s publishing docs for Crates.io](https://docs.rs/cargo-hakari/0.9.36/cargo_hakari/publishing/index.html#a-temporarily-remove-the-workspace-hack-dependency-before-publishing)

NOTE: While total dependency count goes up (because everything in the workspace depends on the workspace hack), rebuilds should go down because every crate will always be building shared dependencies with a single unified feature set regardless of what packages are currently being built.

CC @patrickjcasey


patrickjcasey:
I can test this out tomorrow, but this looks pretty good to me at first glance

cstepanian:
I'm new to the idea of workspace hack crates, but I saw this note in the `cargo hakari` docs:

> ## 2. Keep the workspace-hack up-to-date in CI
> 
> Run the following commands in CI:
> 
> `cargo hakari generate --diff`  # workspace-hack Cargo.toml is up-to-date
> `cargo hakari manage-deps --dry-run`  # all workspace crates depend on workspace-hack
> 
> If either of these commands exits with a non-zero status, you can choose to fail CI or produce a warning message.
> 
> For an example, see [this GitHub action used by cargo-guppy](https://github.com/guppy-rs/guppy/blob/main/.github/workflows/hakari.yml).
> 
> All `cargo hakari` commands take a `--quiet` option to suppress output, though showing diff output in CI is often useful.

https://docs.rs/cargo-hakari/0.9.36/cargo_hakari/index.html#2-keep-the-workspace-hack-up-to-date-in-ci

Should the CI be updated to add these commands, or are there reasons not to do so?

cstepanian:
I linked this PR to issue #892 on the assumption that this resolves that issue.

alilleybrinker:
Ready for review! This now incorporates the CI integration @cstepanian recommended.

alilleybrinker:
CC @j-lanson, this will change the release flow for crates going on Crates.io. `cargo hakari publish -p <CRATE>` instead of `cargo publish -p <CRATE>`. [Full docs](https://docs.rs/cargo-hakari/0.9.36/cargo_hakari/publishing/index.html#a-temporarily-remove-the-workspace-hack-dependency-before-publishing).

The new command ensures the workspace hack is removed for the version published to Crates.io. The presence of the workspace-hack crate is only to improve compile time when compiling many crates in the workspace, so this doesn't impact any behavioral differences between our local-dev version and what ends up on Crates.io.

patrickjcasey:
Everything looks good to me! Do we have the process for publishing crates documented somewhere?

patrickjcasey (approved):

alilleybrinker:
As I recall so far it's only been documented informally between Julian and I, since it's a flow only he and I have the required permissions on the GitHub repo and on Crates.io to execute.

PR #1001, opened 2025-03-07
Parsing sboms to return a url or package

ninaagrawal:


hipcheck/src/target/cyclone_dx.rs
```diff
@@ -2,23 +2,26 @@
...
 	hc_error,
+	target::types::*,
 };
 use cyclonedx_bom::prelude::*;
 use packageurl::PackageUrl;
 use std::str::FromStr;
 use url::Url;
 
+/// return type options for extract_download_url. The target type can be a Url or Package
+pub enum ReturnType {
```
j-lanson, line 17: Rename `BomTarget`

hipcheck/src/target/cyclone_dx.rs
```diff
@@ -38,51 +41,85 @@ pub fn extract_cyclonedx_download_url(filepath: &str) -> Result<Url> {
...
         .component
         .ok_or(hc_error!("CycloneDX file metadata missing a component field describing its own package. Download location cannot be extracted."))?
         .purl
         .ok_or(hc_error!("CycloneDX file metadata component information does not include a pURL. Download location cannot be extracted."))?
         .as_ref()
     )?;
-
+	Ok(purl)
+}
+/// Returns the target type from the BOM file's purl.
```
j-lanson, line 61: Space above to separate functions

hipcheck/src/target/cyclone_dx.rs
```diff
@@ -38,51 +41,85 @@ pub fn extract_cyclonedx_download_url(filepath: &str) -> Result<Url> {
...
 			let split_package: Vec<&str> = full_package.split('@').collect();
-			let package = split_package[0];
+			let name = split_package[0].to_string();
 			let version = match split_package.len() {
 				1 => "no version",
 				_ => split_package[1],
+			}
+			.to_string();
+			// converts the purl to a url so it can be passed to a package
+			let purl_as_url = Url::parse(&purl.to_string()).map_err(|e| hc_error!("Cannot parse constructed GitHub repository URL constructed from CycloneDX file download location: {}", e))?;
```
mchernicoff, line 86: Probably want to have a different error message here.

hipcheck/src/target/cyclone_dx.rs
```diff
@@ -38,51 +41,85 @@ pub fn extract_cyclonedx_download_url(filepath: &str) -> Result<Url> {
...
-			let package = split_package[0];
 			let version = match split_package.len() {
 				1 => "no version",
 				_ => split_package[1],
+			}
+			.to_string();
+			let first_half_of_package: Vec<&str> = split_package[0].split(":").collect();
+			let name = first_half_of_package[1].to_string();
+			// the purl must be in url format in order to pass to a package
+			let purl_as_url = Url::parse(&purl.to_string()).map_err(|e| hc_error!("Cannot parse constructed GitHub repository URL constructed from CycloneDX file download location: {}", e))?;
```
mchernicoff, line 106: See previous comment about error message.

hipcheck/src/target/resolve.rs
```diff
@@ -280,25 +278,39 @@ impl ResolveRepo for MavenPackage {
...
+				// the download location for a cyclonedx can be a package or a URL
+				match extract_cyclonedx_download_url(source) {
+					Ok(ReturnType::Url(url)) => create_remote_git_repo(t, url),
+					Ok(ReturnType::Package(package)) => {
+						t.package = Some(package.clone());
+						package.resolve(t)
+					}
+					Err(error) => Err(hc_error!("Error, failed to parse package or url {}", error)),
+				}
+			}
```
j-lanson, line 312: I think this would be slightly cleaner and would eliminate the need for the inner function if we restructure it as such:

```
let bom_target: ReturnType = match self.standard { }
match bom_target {
  ReturnType::Url(url) => { \* contents of create_remote_git_repo() *\ }
  ReturnType::Pacakge(pkg) => ...
}
```

j-lanson (changes requested):
Looking good, just some small renames/restructures

j-lanson:
Also, please squash down to 1 commit

ninaagrawal:
@j-lanson apologies, I must have forgotten to squash the commit when I edited the enum

hipcheck/src/target/cyclone_dx.rs
```diff
@@ -38,51 +41,89 @@ pub fn extract_cyclonedx_download_url(filepath: &str) -> Result<Url> {
...
         .as_ref()
     )?;
+	Ok(purl)
+}
 
+/// Returns the target type from the BOM file's purl.
+/// If the purl's type is github or maven, return a repo URL.
+/// If the target type is npm or pypi, return the package
+fn extract_download_url(bom: Bom) -> Result<BomTarget> {
+	let purl = extract_purl(&bom).unwrap();
```
cstepanian, line 67: Please replace `unwrap()` with `?` operator.

cstepanian: Also, please include information about the `Bom` in the error context to help debug the format, in cases when it cannot be parsed correctly.

That could be added here, where `extract_purl()` is called, or inside `extract_purl()` at each error site.

The context probably shouldn't be added to both, as that would result in redundant info in error messages.

ninaagrawal: @cstepanian Can you please explain what you want me to add for the second message? What kind of information do I need to include about the bom and where? 

Also I think extract_purl already includes error messages so what else do I need to add?

cstepanian: Right, there are already error messages, but they don't currently include information from the `Bom` that would show a user or developer what the original data was that couldn't be parsed. All they will see is the static error string.

This can be done by using the pattern `hc_error!("Error message, context = {}", context)`. Here, I'm requesting you include the `Bom` itself as `context`.

If the whole variable can't be printed (doesn't implement `Debug` or `Display`) or is too big to be reasonable to print in full, then some field(s) from it may be enough, or sometimes there needs to be a method called on a variable to make it displayable.

ninaagrawal: Since we're trying to print the purl would it make sense to just print that or something else?

mchernicoff: But you can get this error if you cannot find a pURL in the Bom (honestly, this is the most likely reason we would error). It might be enough to use the filepath to the Bom the user passed, but that might be too simple since the user already provided that. The problem is that a `Bom` struct doesn't `impl Display` and is too big to use `Debug`. Almost all its fields are `Option`s, so it would be hard to pick a single field or subfield to use to display more information.

mchernicoff: I suppose you could print a level higher for some of the errors. We are looking for `bom.metadata.component.purl` (where everything but the passed `bom` is an `Option`), so if something is `None` you can use the `Debug` implementation of the higher level struct and display that. But that would display a lot of information to say "The field you are looking for is not in the file."

mchernicoff: So I'm leaning towards having the errors changed from "CycloneDX file . . . " to "CycloneDX file foo/bar/my_bom.json . . ." to at least give some information. In all cases, this means that the SBOM we are using is not suitable as a Hipcheck target, so there isn't much else the user can do at that point. But having the filename might be useful if there is some future version of Hipcheck that can have multiple BOMs passed to it at once.

ninaagrawal: But how useful would printing this error be if the user already know what they're passing in? Wouldn't it be easier to just return the error the code produces?

mchernicoff: Oh, sorry. I meant include the SBOM path and the error message, as Cal suggested. Since `extract_purl()` doesn't have the path information, you could use `context()` method (`use error::Context as _` to have access to it) when you call `extract_purl()` in `extract_download_url()` to add more error information that includes the file path. Or you can map the error that `extract_purl()` returns if you want to customize the error message.

j-lanson: from meeting: Michael meant `extract_download_cyclonedx_url()`

ninaagrawal:
@j-lanson @cstepanian @mchernicoff I converted this to draft because I think  when you run hc check on an sbom where the pURL **does not** have a version, the check should ask for the reference flag. Otherwise if the pURL does have a reference you do not need the reference flag. Please correct me if I'm wrong.

j-lanson:
> @j-lanson @cstepanian @mchernicoff I converted this to draft because I think when you run hc check on an sbom where the pURL **does not** have a version, the check should ask for the reference flag. Otherwise if the pURL does have a reference you do not need the reference flag. Please correct me if I'm wrong.

I believe the code should continue to do what it used to do when an Sbom parses as a RemoteRepo instead of as Package. If it does parse as a Package and there is no version, yes I think the `--ref` flag is necessary.

ninaagrawal:
In that case, I'll unmerge the draft

hipcheck/src/target/cyclone_dx.rs
```diff
@@ -27,62 +31,102 @@ pub fn extract_cyclonedx_download_url(filepath: &str) -> Result<Url> {
 		})?;
 		if bom.validate().passed() {
 			extract_download_url(bom)
+				.with_context(|| format!("The file where the error occurs is: {}", filepath))
```
j-lanson, line 34: Reword the context message: `"Failed to extract download URL from {}", filepath`

hipcheck/src/target/cyclone_dx.rs
```diff
@@ -27,62 +31,102 @@ pub fn extract_cyclonedx_download_url(filepath: &str) -> Result<Url> {
...
 		} else {
 			Err(hc_error!("CycloneDX file is not a valid SBOM"))
 		}
 	} else if filepath.contains(".xml") {
 		let bom = parse_from_xml(contents)?;
 		if bom.validate().passed() {
 			extract_download_url(bom)
+				.with_context(|| format!("The file where the error occurs is: {}", filepath))
```
j-lanson, line 42: See above.

hipcheck/src/target/cyclone_dx.rs
```diff
@@ -27,62 +31,102 @@ pub fn extract_cyclonedx_download_url(filepath: &str) -> Result<Url> {
...
+				version,
+				host: PackageHost::Npm,
 			};
-			extract_repo_for_npm(package, version).context("Could not get git repo URL for CycloneDX file's corresponding NPM package",)
-		},
-		Some((TargetType::Pypi, full_package)) =>  {
+			Ok(BomTarget::Package(package))
+		}
+		Some((TargetType::Pypi, full_package)) => {
+			println!("full package {}", full_package);
```
j-lanson, line 102: Delete `println`

j-lanson (changes requested):

j-lanson (approved):
Looks good, thanks Nino!

PR #1007, opened 2025-03-10
feat: bidirectional query protocol with chunking in Python SDK

j-lanson:
Continued impl of the Python SDK. Added:
- Better graceful exit code
- Bidirectional query protocol handling
- Chunking and synthesizing

There are also a couple minor fixes to the Rust SDK that I found while porting the code over, mostly to do with returning inaccurate error types.


sdk/python/src/hipcheck_sdk/engine.py
```diff
@@ -1,21 +1,257 @@
...
+    plugin, name = split_once(rest, "/")
+    if name is None:
+        name = ""
+    return (publisher, plugin, name)
+
+
+class QueryBuilder:
+    def __init__(self, engine, target: str):
+        self.engine = engine
+        self.keys = []
```
ninaagrawal, line 43: What are these fields for, it may be worth commenting if it hasn't been explained elsewhere

sdk/python/src/hipcheck_sdk/engine.py
```diff
@@ -1,21 +1,257 @@
...
+                output=[],
+                concerns=[]
+            )
+
+            await self.send(query)
+            resp: Query = await self.recv()
+            return list(resp.output)
+
+    # Query another Hipcheck plugin `target` with key `input`. On success, the deserialized result object
+    # of the query is returned. `target` shoul be a string of the format
```
ninaagrawal, line 124: May be worth putting in line 46 when you initialize the variables

sdk/python/src/hipcheck_sdk/chunk.py
```diff
@@ -0,0 +1,393 @@
...
+    key: List[object]
+    output: List[object]
+    concerns: List[str]
+
+    def try_from(raw: gen.Query):
+        # assert(len(raw.key) == len(raw.output))
+        direction = QueryDirection.try_from(raw.state)
+
+        def get_fields_rfd9(v: gen.Query) -> Tuple[List[object], List[object]]:
+            keys = []
```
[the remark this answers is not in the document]
j-lanson: It's an interesting thought, and would definitely be the way to go if/when we have more fields, but for now I'm going to leave it as-is. Not only is that the way that the Rust SDK does it, but `dict` lookups such as `dict["key"]` are in actuality most likely slower than accessing an element of a tuple (because the key has to be hashed)

ninaagrawal (dismissed):
Looks good, nothing to comment

ninaagrawal (approved):
looks good!

Issue #1008, opened 2025-03-10
Expand `xtask ci` to check for SPDX header on `.py` files


j-lanson:
Currently we have an `xtask` procedure that ensures rust files have the SPDX license header at the top. Expand this check logic to also include .py files.

mchernicoff:
@j-lanson Should I go ahead and add the SPDX licenses to the .py files, now that I can see they are missing?

j-lanson:
I've already added them on the latest Python SDK PR, so no need

PR #1011, opened 2025-03-11
feat(xtask): xtask validates Python projects

mchernicoff:
Resolves issue #1008

`xtask validate` now validates both Rust crates and Python projects. For now, we manually include Hipcheck's only Python project in the list of code packages to examine.

### DIFF
--- a/xtask/src/task/validate.rs
+++ b/xtask/src/task/validate.rs
@@ -6,19 +6,20 @@ use crate::workspace;
 use anyhow::{anyhow, Context as _, Result};
 use glob::glob;
 use pathbuf::pathbuf;
+use pep440_rs::{Version, VersionSpecifiers};
+use pyproject_toml::{Contact, License, Project, PyProjectToml};
 use serde::{de::DeserializeOwned, Deserialize};
 use std::{
 	collections::BTreeSet,
-	fmt,
-	fmt::{Debug, Display, Formatter},
-	fs,
-	fs::File,
+	fmt::{self, Debug, Display, Formatter},
+	fs::{self, File},
 	io::{BufRead, BufReader},
 	ops::Not as _,
 	path::{Path, PathBuf},
+	str::FromStr,
 };
 
-/// Print list of validation failures for crates in the workspace.
+/// Print list of validation failures for packages in the workspace.
 pub fn run() -> Result<()> {
 	log::info!("beginning validation");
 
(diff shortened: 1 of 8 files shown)


xtask/src/task/validate.rs
```diff
@@ -206,69 +221,152 @@ impl Display for ConfigIssues {
...
+		findings.insert(HasAuthors);
+	}
+
+	log::info!(
+		"validating Python project '{}' doesn't specify license_file",
+		pyproject.name
+	);
+	if package_license_file_present(pyproject) {
+		findings.insert(PackageIssues::LicensePresent);
+	}
```
j-lanson, line 320: My lack of understanding about how this xtask works - isn't having authors and a license file a good thing? or is the idea that we only want those things at the "top level"

j-lanson (approved):

xtask/src/task/validate.rs
alilleybrinker: Yeah, we want to just delegate to the workspace and have crates inherit from it.

alilleybrinker (dismissed):

xtask/src/task/validate.rs
j-lanson: Does that still make sense for checking license info / author info for the Python SDK?

alilleybrinker: Hm, fair point. No, I think for Python since it can't inherit from the Cargo workspace, we *do* want license information to be present in the package manifest.

j-lanson (changes requested):
See discussion

mchernicoff:
I will flip the logic the other way around for Python. :)

mchernicoff:
Double checking: Do we want there to be a License file in the Python SDK? Right now we do check for a valid license in the pyproject.toml file, we just don't want there to be two entire License files in the workspace. Should I change that to make sure there is a second license file just for the Python SDK?

j-lanson:
I guess we can still check for a license file, there is in fact a field in the pyproject.toml to specify license files https://packaging.python.org/en/latest/guides/writing-pyproject-toml/

hipcheck/src/target/cyclone_dx.rs
```diff
@@ -2,23 +2,27 @@
...
 use std::str::FromStr;
 use url::Url;
 
+/// return type options for extract_download_url. The target type can be a Url or Package
+#[derive(Debug)]
+pub enum BomTarget {
+	Url(Url),
+	Package(Package),
+}
+
```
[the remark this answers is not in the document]
mchernicoff: I might have messed up something with the rebase. Will investigate.

j-lanson (approved):

Issue #981, opened 2025-03-03
Transition TLS certificate logic to `rustls-platform-verifier`


alilleybrinker:
The `rustls-platform-verifier` crate (https://github.com/rustls/rustls-platform-verifier) provides a better integration with platform certificates, and should be both faster and more correct (it handles revocation correctly) than what we do today. We should swap it out for our current solution using `rustls` native certs.

PR #1028, opened 2025-03-17
feat: Replace rustls-navtive-certs with rustls-platform-verifier

mchernicoff:
Resolves issue #981 . Switches from using `rustls-native-certs` to `rustls-platform-verifier` in the three places in our code where we add TLS certs to an HTTPS Agent.


cstepanian:
I manually linked to issue #981, since it looks like GitHub didn't pick up "Resolves issue #981" as anything other than a mention of that issue.

j-lanson (approved):

PR #1031, opened 2025-03-17
chore: write RFD for report caching

j-lanson:


site/content/docs/rfds/0012-report-caching.md
```diff
@@ -0,0 +1,201 @@
...
+for a simple implementation using a directory hierarchy as keys, and the "value"
+is a file in the lowest subdirectory. The proposed report cache keying strategy
+has more elements and does not have an implicit hierarchy; if we were to try
+using directories as keys, would the policy file hash be parent to the Hipcheck
+binary hash or vice versa? Where would the repository fit in? These are
+important questions because if we adopt a directory-based system, if we wanted
+to invalidate the cache based on a particular sub-key, we would have to search
+the entire `report/` directory recursively for paths containing that sub-key.
+
+Therefore, we propose using a simple database file, such as a SQLite database,
```
mchernicoff, line 95: If I'm understanding this correctly, that will mean anyone using Hipcheck going forward need to have a database set up to run Hipcheck. How much of the installation and configuration overhead for that will Hipcheck be responsible for vs. adding additional system requirements? I admit I'm not super familiar with setting up a SQLite databse alongside an application, so maybe this isn't a big worry.

site/content/docs/rfds/0012-report-caching.md
```diff
@@ -0,0 +1,201 @@
...
+// If no <HC_PATH> supplied, defaults to the current
+// `hc` binary.
+hc cache report delete --hc [<HC_PATH>]
+
+// Invalidate all entries keyed by this policy file
+hc cache report delete --policy <POLICY_PATH>
+
+// Invalidate all entries keyed by this target, use
+// target resolution system to resolve to a repo/ref
+hc cache report delete --target <TARGET> [--ref <REFSPEC>]
```
mchernicoff, line 147: Probably not necessary, but would we want some kind of AND logic on the --hc, --policy, and --target flags?

site/content/docs/rfds/0012-report-caching.md
```diff
@@ -0,0 +1,201 @@
...
+  primary_author_link: https://github.com/j-lanson
+  status: Proposed
+---
+
+# Report Caching
+
+In anticipation of users running Hipcheck as a GitHub action against the
+dependencies of their own project, we propose to add to Hipcheck the ability to
+cache reports. Given the same policy file, Hipcheck binary, target repository
+and ref, Hipcheck is deterministic across runs. Therefore, we can help drive
```
alilleybrinker, line 17: I think this isn't _exactly_ true; for example, I don't think we currently time-limit our analysis of pull requests from the GitHub API to the date/time of the target ref. We could probably do better here.

alilleybrinker (changes requested):
How would marking a report as reviewed work in CI?

j-lanson:
> How would marking a report as reviewed work in CI?

Good point, I'm actually not sure. Will have to think about it

site/content/docs/rfds/0012-report-caching.md
j-lanson: Good point, I've updated the language in the RFD and will make an issue

alilleybrinker:
Since we're in a state where we wouldn't be able to act on implementing this if the RFD were merged, I'm going to close this. I'd like to revive it in the future, but want to be realistic about bandwidth.

PR #1032, opened 2025-03-17
chore: write RFD for expecting set_config hash

j-lanson:


alilleybrinker (changes requested):
On more consideration, I think we *should* specify a hash algorithm, just because with extra degrees of freedom people may try to pick really bad ones like MD5. I recommend requiring SHA-256.

j-lanson:
> On more consideration, I think we _should_ specify a hash algorithm, just because with extra degrees of freedom people may try to pick really bad ones like MD5. I recommend requiring SHA-256.

Post-discussion: we should specify an alg like SHA256

site/content/docs/rfds/0013-plugin-config-hash.md
```diff
@@ -0,0 +1,54 @@
...
+}
+```
+
+The `message` field is currently only used for communicating an error. We
+propose to update the semantics of this structure such that the `message` field
+is expected also for a sucessful configuration attempt. That field's value
+should be a hex string representing the SHA-256 hash of the configuration. For
+plugins that take paths to custom config files as configuration, the hash
+calculation should include the content of these files such that changes to them
+result in a different hash.
```
alilleybrinker, line 41: One more wrinkle, we need to specify that the hexadecimal representation should use only lowercase characters, since we'll need to do lexicographic ordering for concatenation later.

alilleybrinker (changes requested):
Sorry Julian! Thought of one more wrinkle we need to specify.

site/content/docs/rfds/0013-plugin-config-hash.md
j-lanson: we can always `to_lower()` the strings ourselves, but I can add that as a stipulation in the RFD language.

alilleybrinker (approved):

PR #1037, opened 2025-03-20
Standardize the INFO Start/End logs of single-metric plugins

### LINKED ISSUE #915: Standardize Logging Across Plugins
Some plugins like entropy and fuzz produce no logs, while others like typo have both start and end info level logs during runs.

Standardize plugin logging and make them more useful for debugging and hipcheck usage.

KirilldogU:
Resolves #915 

Standardize the plugin logs by writing the same INFO level log "running XXXX query" and "completed XXXX query" in each single-metric plugin. 

This makes the logs more consistent and is good first step in debugging, by noting if any processes failed to complete.


KirilldogU:
This PR adds INFO start/end logs to all the plugins **except**: npm, linguist, git, and github. Those plugins appear to be used by the other plugins during analysis and do not provide their own analysis metric so I chose to leave them without this INFO log.
Is this correct?

alilleybrinker:
> This PR adds INFO start/end logs to all the plugins **except**: npm, linguist, git, and github. Those plugins appear to be used by the other plugins during analysis and do not provide their own analysis metric so I chose to leave them without this INFO log. Is this correct?

I'd still add it. They're not "top-level" plugins (they don't expose a default query), but it can still be useful for debugging to know when they're running.

j-lanson:
> > This PR adds INFO start/end logs to all the plugins **except**: npm, linguist, git, and github. Those plugins appear to be used by the other plugins during analysis and do not provide their own analysis metric so I chose to leave them without this INFO log. Is this correct?
> 
> I'd still add it. They're not "top-level" plugins (they don't expose a default query), but it can still be useful for debugging to know when they're running.

I agree with Andrew, the reason we did the work of adding parsing and per-plugin control through `HC_LOG` was so that people can turn off / reduce the log level of plugins as they wish. If they don't want a bunch of `git` plugin logs they can do `HC_LOG=plugin::git=warn`

plugins/fuzz/src/main.rs
```diff
@@ -8,7 +8,9 @@ use std::result::Result as StdResult;
 /// Returns whether the target's remote repo uses Google's OSS fuzzing
 #[query(default)]
 async fn fuzz(engine: &mut PluginEngine, key: Target) -> Result<Value> {
+	tracing::info!("running fuzz query");
 	if let Some(remote) = &key.remote {
+		tracing::info!("completed fuzz query");
 		engine.query("mitre/github", remote.clone()).await
```
j-lanson, line 14: It would be more correct to receive the result, log, then return.

j-lanson, line 14: We may also want `trace!` level logs before the plugin calls `engine.query` to say we are querying so-and-so plugin for x reason

j-lanson (changes requested):

plugins/fuzz/src/main.rs
KirilldogU: Fixed! Along with a few similar cases

plugins/activity/src/main.rs
```diff
@@ -20,7 +20,7 @@ static CONFIG: OnceLock<Config> = OnceLock::new();
 /// (Which means that anything expecting a `Span` must parse the output of this query appropriately)
 #[query(default)]
 async fn activity(engine: &mut PluginEngine, target: Target) -> Result<String> {
-	tracing::debug!("running activity query");
```
j-lanson, line 23: change to `running default query` for functions marked `#[query(default)]`, else use the name of the function.

plugins/git/src/main.rs
```diff
@@ -32,11 +32,15 @@ pub static CACHE: OnceLock<Mutex<GitRawCommitCache>> = OnceLock::new();
 /// (Which means that anything expecting a `Timestamp` must parse the output of this query appropriately)
 #[query]
 async fn last_commit_date(_engine: &mut PluginEngine, repo: LocalGitRepo) -> Result<String> {
+	tracing::debug!(target: "git::last_commit_date", "running query");
```
j-lanson, line 35: I would make it `debug!("running last_commit_date query")` instead of messing with the target

j-lanson: Also, as Andrew and I mentioned before, these should be `tracing::info` to match what the other plugins have. If people don't like it at info level they can customize the `HC_LOG` value to turn it off for these.

j-lanson (changes requested):

plugins/git/src/main.rs
KirilldogU: Function name has been moved to the message instead of the target in all of these cases. 
These cases are also switched to the INFO level instead of the DEBUG level with the exception of the linguist plugin which produces a spam of output due to frequent use of the is_source_file() query. Linguist remains at debug level.

plugins/fuzz/src/main.rs
KirilldogU: Found 11 instances of plugins querying other plugins and added trace!() level logs above them. Between the log's target of the top-level plugin name and the message with the endpoint it's querying, the trace logs generally appear intuitive, but let me know if these should be extended!

j-lanson (approved):
Looks good Kirill, thanks!

alilleybrinker (approved):

PR #1040, opened 2025-03-20
Infers ref from git VCS URLs

### LINKED ISSUE #1034: Add VCS URL TargetSeed
Add support for git [VCS URLs](https://pip.pypa.io/en/stable/topics/vcs-support/).

Currently in our CLI target kind resolution code we discard the VCS information if we detect it. We shouldn't need to add a new `TargetSeedKind`, since it's technically just a remote repo with refspec information.

### LINKED ISSUE #1057: Add `TargetSeedKind::VcsUrl`

mchernicoff:
Resolves #1034 Now if a remote git VCS URL is provided with ref information, we set the `TargetSeed.refspec` to that ref instead of simply discarding it.

If a ref is also provided by the user with the `--ref` flag that does not match the VCS URL ref, Hipcheck will error out.

Edit: Also resolves #1057

### DIFF
--- a/hipcheck/src/cli.rs
+++ b/hipcheck/src/cli.rs
@@ -15,7 +15,7 @@ use crate::{
 	target::{
 		pm, LocalGitRepo, MavenPackage, Package, PackageHost, Sbom, SbomStandard, SingleTargetSeed,
 		SingleTargetSeedKind, TargetSeed, TargetSeedKind, TargetType, ToTargetSeed,
-		ToTargetSeedKind,
+		ToTargetSeedKind, VcsUrl,
 	},
 };
 use clap::{Parser as _, ValueEnum};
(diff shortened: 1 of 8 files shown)


hipcheck/src/cli.rs
```diff
@@ -586,6 +586,29 @@ impl CheckArgs {
...
+				// Check if the user also provided a ref, and error if it does not agree with the inferred ref
+				if let Some(init_ref) = &self.refspec {
+					if init_ref != ref_spec {
+						return Err(hc_error!(
+							"Provided ref_spec '{}' does not match the ref spec, '{}', inferred from the target '{}'. Check that you have specified the correct ref and provided the intended target.",
+							init_ref, ref_spec, target
+						));
+					}
+				}
+				self.refspec = Some(ref_spec.to_string());
```
j-lanson, line 604: I think we should check here if `self.refspec` is already set and return an error if there's a `--ref` flag and a ref embedded in the URL. I wish this logic could go down where we do the same thing with Packages around line 620, but I don't know that it's possible without adding a new TargetSeedKind.

I'd just want to make sure that if we error here that it looks good on the CLI.

hipcheck/src/cli.rs
```diff
@@ -586,6 +586,29 @@ impl CheckArgs {
 			},
 		}
 
+		// If the target is a remote VCS URL, clean it up now so we can save the ref spec information
+		if subcmd_str == "repo" && target_str.starts_with("git+") {
+			// Remove git prefix
+			let tgt_trimmed = &target_str.replace("git+", "");
+			// Remove any git ref information that trails the end of the URL and set the ref_spec equal to that ref
+			if let Some((url, ref_spec)) = tgt_trimmed.split_once(".git@") {
```
j-lanson, line 594: Are we sure this logic will get triggered if the user does the legacy `hc check repo <VCS_URL>`?

j-lanson (changes requested):

alilleybrinker:
This may conflict with #1047 which is in the process of merging. Should be an easy fix though.

hipcheck/src/cli.rs
mchernicoff: I managed to move it to the same place as the other validation step.

mchernicoff:
Have to handle a couple of failed tests

j-lanson:
It looks like you have accidentally added some of Kirill's changes that are already on main?

mchernicoff:
> It looks like you have accidentally added some of Kirill's changes that are already on main?

I need to figure out why I keep doing that. I'll fix it when I can. Might have to wait.

mchernicoff:
> > It looks like you have accidentally added some of Kirill's changes that are already on main?
> 
> I need to figure out why I keep doing that. I'll fix it when I can. Might have to wait.

Should be fixed now.

j-lanson (approved):
Looks good, thanks Michael!

mchernicoff:
Made changes to make this compatible with the changes to TargetSeed. Note that the current way of doing things only works for cases of a single target VCS URL. If we want to implement this feature for VCS URLs, we will need to make future changes.

hipcheck/src/target/resolve.rs
```diff
@@ -154,6 +154,10 @@ impl TargetResolver {
 				resolver.remote = Some(repo.clone());
 				repo.resolve(&mut resolver)
 			}
+			VcsUrl(vcs) => {
+				resolver.remote = Some(vcs.remote.clone());
+				vcs.remote.resolve(&mut resolver)
```
j-lanson, line 159: How do we ensure that `vcs.git_ref` is reflected in `seed.ref`? Otherwise it feels like there's a latent possibility for this info to get dropped

j-lanson (approved):

hipcheck/src/target/resolve.rs
mchernicoff: So we set `seed.refspec` equal to `vcs.git_ref` when we run `to_target_seed()` on `CheckArgs` as long as `vcs.git_ref` is not `None` and we don't error out from contradictory ref information. So if nothing else, we are losing information if there is no ref in the VCS URL but a `--ref` is provided.

mchernicoff: If we wanted to, it wouldn't be hard to throw in an extra check somewhere that they are the same if `vcs.git_ref.is_some()`. If we wanted to do that later on.

Issue #1009, opened 2025-03-11
Document the `--ref` flag for `hc check` in the site docs


alilleybrinker:
It's not there currently, though it's in the help text for the `hc check` command.

PR #1047, opened 2025-03-21
chore: improve hc check --ref help

patrickjcasey:
When updating the documentation for #1009, I found `--ref <REFSPEC>` to be confusing. We are looking for a [git ref/reference](https://git-scm.com/book/en/v2/Git-Internals-Git-References) NOT a [git refspec](https://git-scm.com/book/en/v2/Git-Internals-The-Refspec)

Here is the new help menu:
image


alilleybrinker (approved):

alilleybrinker:
Good catch!

PR #1048, opened 2025-03-22
refactor: TargetSeedKind refactor to support single and mulit targets

patrickjcasey:
First pass at implementing the minimal amount of changes to support this function signature

```rust
impl TargetSeed {
    pub fn get_targets(config: TargetResolverConfig) -> impl Iterator<Item = Target> {
    ...
```


j-lanson (approved):
Looks good, thanks Patrick!

PR #1061, opened 2025-03-24
Python SDK polishing

### LINKED ISSUE #1060: Add Target type defined by schema to PythonSDK
In the Rust SDK, we use JSON schemas exported from Hipcheck core to define the `Target` type that query endpoints can take as input.

We need to do the same or something similar for the Python SDK

j-lanson:
Resolves #1060 .

- Added formatting with `ruff`
- Added `update_proto.py` script so we can update the contents of `gen` when the Hipcheck proto definition changes. I tried hard to update it using build scripts through `pyproject.toml` but it wasn't possible
- Added docstrings to a variety of types and basic sphinx configuration to convert to HTML


sdk/python/src/hipcheck_sdk/cli.py
```diff
@@ -0,0 +1,29 @@
...
+def get_parser_for(plugin_name: str) -> argparse.ArgumentParser:
+    """
+    Get the default argument parser for a Hipcheck plugin
+
+    :return: An ArgumentParser configured to capture Hipcheck plugin CLI
+        arguments
+    """
+    parser = argparse.ArgumentParser(prog=plugin_name)
+    parser.add_argument("-p", "--port", type=int)
+    parser.add_argument("-l", "--log-level", type=str, default="off")
```
KirilldogU, line 16: We default to Error level in the Rust SDK at the moment

site/content/docs/guide/making-plugins/python-sdk.md
```diff
@@ -0,0 +1,230 @@
...
+
+The Hipcheck team maintains a Python library that provides developers with tools
+to greatly simplify plugin development. This section provides a high-level guide
+on how to implement a Hipcheck plugin using the library.
+
+This guide assumes familiarity with the Python language, async Python using
+[asyncio][asyncio], and Python project management. We leave it to the developer
+to select the right Python project dependency and packaging management tool
+that is right for them, from among the many that exist. For the purpose of
+[packaging and releasing your plugin][plugin_release], we suggest using one
```
KirilldogU, line 16: Unsure if this is local issue, but I think `plugin_release` need a value like:
`[plugin_release]:(link/md).`

sdk/python/scripts/update-gen.py
```diff
@@ -37,8 +38,27 @@
...
+    "--output",
+    f"{schema_gen_path}",
+]
+schema_gen_res = subprocess.run(datamodel_args, capture_output=True)
+if schema_gen_res.returncode != 0:
+    print("types.py generation failed: ", schema_gen_res.stderr)
+
 # Prepend the SPDX License header to the relevant .py files
-py_files = ["hipcheck_pb2.py", "hipcheck_pb2_grpc.py"]
+py_files = ["hipcheck_pb2.py", "hipcheck_pb2_grpc.py", "types.py"]
```
KirilldogU, line 61: Running `uv run scripts/update-gen.py` with no changes causes the timestamp in `types.py` to update while leaving the other files unchanged. This creates a `git diff`, so is there a way to remove the timestamp from the signature?

j-lanson:
> Should the ruff formatter be part of our cargo xtask jobs the way that cargo fmt is?

Yes this would be a nice thing to have.

sdk/python/scripts/update-gen.py
j-lanson: Yes, the script can be updated to do so

site/content/docs/guide/making-plugins/python-sdk.md
j-lanson: Ah the problem is that the second part is `plugin_release` instead what it should be which is  `plugin-release` to match with the list of links at the bottom.

KirilldogU: Fixed!

sdk/python/src/hipcheck_sdk/engine.py
```diff
@@ -37,47 +54,73 @@ def parse_target_str(target: str) -> Tuple[str, str, str]:
     return (publisher, plugin, name)
```
ninaagrawal, line 54: Where are you getting name from? You only seem to get publisher and rest from target

ninaagrawal (approved):
Looks good!

PR #1066, opened 2025-03-25
chore: move plugin query trace log to engine.rs from plugins

### LINKED ISSUE #1064: Move query call start tracing into SDK
When we make a query with `PluginEngine::query` in the Rust SDK, we currently do a manual `tracing::trace!` macro call. That should be part of the SDK, but in the plugins themselves, to ensure we always do it.

KirilldogU:
Resolves #1064 
Moves trace level logs occurring in plugins to the shared functions in the query engine that they call.

### DIFF
--- a/sdk/rust/src/lib.rs
+++ b/sdk/rust/src/lib.rs
@@ -124,6 +124,14 @@ impl FromStr for QueryTarget {
 		}
 	}
 }
+impl std::fmt::Display for QueryTarget {
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+		match &self.query {
+			Some(query) => write!(f, "{}/{}/{}", self.publisher, self.plugin, query),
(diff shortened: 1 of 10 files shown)


sdk/rust/src/lib.rs
```diff
@@ -124,6 +124,14 @@ impl FromStr for QueryTarget {
 		}
 	}
 }
+impl ToString for QueryTarget {
```
patrickjcasey, line 127: It would be more idiomatic to impl Display instead of ToString. ToString is automatically derived from a Display implementation

patrickjcasey (commented):
Besides my commit this looks good!

sdk/rust/src/lib.rs
KirilldogU: Ah, great point. Fixed!

patrickjcasey (approved):

PR #1069, opened 2025-03-26
Jlanson/python sdk cleanup

j-lanson:
1. Fixed up some language and links in the documentation
2. Added support for turning `key` JSON dict into type hinted by query endpoint func signature


sdk/python/README.md
```diff
@@ -11,6 +11,18 @@ From `sdk/python`, run
 uv run pytest ./tests
```
KirilldogU, line 11: I just ran `uv run pytest ./tests` from sdk/python and both tests errored with: `E   ModuleNotFoundError: No module named 'python_jsonschema_objects'`. Is the schema not committed to the repo? Does it need to be regenerated?

sdk/python/README.md
```diff
@@ -31,3 +43,16 @@ entry:
...
+Currently, this will get through configuration and receive the query from
+Hipcheck core, but the plugin will appear as an error in the final report
+because it is not currently written to accept `Target` objects.
+
+# Building the HTML
+
+From `docs/`, run the following:
+
+```bash
+uv run make html # Build the html in the build/ directory
```
KirilldogU, line 56: When running this command I also get an error from missing schema as in the previous comment. Error:
`WARNING: autodoc: failed to import module 'cli' from module 'hipcheck_sdk'; the following exception was raised:
No module named 'python_jsonschema_objects' [autodoc.import_object]`

j-lanson:
> Running `ruff format` edits most of the changed files. Unsure if we're going to make that a default, but if so it should be applied.

Good catch.

sdk/python/README.md
j-lanson: Removed dangling import of removed dependency.

sdk/python/README.md
j-lanson: Removed dangling import of removed dependency.

KirilldogU (approved):
Looks good!

PR #1070, opened 2025-03-26
feat: help target-triple added

### LINKED ISSUE #1015: Add `hc help target-triple` to show the current inferred target triple and list others currently known.
This is intended to help with debugging architecture-related issues.

aamohd:
Closes #1015 .

Test run is `hc explain target-triple` and currently prints as:

```
Current target triple architecture (Known):
  > Apple Silicon (ARM64), macOS

Other supported architectures:
  - Intel x86_64, macOS
  - Intel x86_64, Windows (MSVC)
  - Intel x86_64, Linux (GNU)
  - ARM64, Linux
 ```

### CODE REVIEW

**Review by alilleybrinker — DISMISSED** [review]
The goal had been that this would be a subcommand of `help`, though here it looks like it's been added as its own top-level command. In general I am loathe to add more top-level commands if we can avoid it.

File: hipcheck/src/plugin/arch.rs [hunk 1]
```diff
@@ -28,6 +30,12 @@ pub enum Arch {
 	Unknown(String),
 }
 
+impl Arch {
+	pub fn known_iter() -> impl Iterator<Item = Arch> {
+		KnownArch::iter().map(Arch::Known)
```
j-lanson, line 35: Why bother wrapping in `Arch`? Could just return `Item = KnownArch` and I think the logic in `main.rs` would become simpler

j-lanson (changes requested):
See comment

hipcheck/src/plugin/arch.rs
j-lanson: So I might suggest just removing this function and using `KnownArch::iter()` in main.

j-lanson (dismissed):

j-lanson:
Looks like you have some simple merge conflicts to resolve.

j-lanson (approved):

PR #1072, opened 2025-03-26
feat: Allow plugins to return error information as query responses

cstepanian:
- Add `error` field to `Query` schema in protobuf
- Add `error: Option<String>` to `struct Query` in `hipcheck-common/src/types.rs`
- Add error variant to `QueryState` enum in protobuf
- Add `Error` case to `PluginResponse`

Remaining TODOs
- [ ] Create `PluginResponse::Error` when appropriate in its `impl From<Query>` block
- [ ] Make sure that `handle_session` passes error responses to `send_session_err`
- [ ] Update `send_session_err` to send error strings instead of just setting `QueryState` to `Unspecified`
- [ ] Test that the changes are effective
    - [ ] Change a plugin's use of `QueryState::Unspecified` to `QueryState::Error` and pass an error string
    - [ ] Confirm that Hipcheck prints the error
- [ ] Ensure all possible data paths for plugin errors eventually get printed in a useful way
- [ ] Update documentation
- [ ] Write RFD that explains changes

### DIFF
--- a/library/hipcheck-common/proto/hipcheck/v1/hipcheck.proto
+++ b/library/hipcheck-common/proto/hipcheck/v1/hipcheck.proto
@@ -200,6 +200,10 @@ message Query {
     // Used to indicate whether or not a string field present in a `repeated string` field
     // was split between two messages
     bool split = 9;
+
+    // A plugin-generated error string.
+    // Shall be present if and only if state == QUERY_STATE_ERROR
+    optional string error = 10;
 }
 
 enum QueryState {
@@ -220,6 +224,9 @@ enum QueryState {
 
     // We are sending a query to a plugin and we are expecting to need to send more chunks
     QUERY_STATE_SUBMIT_IN_PROGRESS = 4;
+
+    // An error was encountered while processing.
+    QUERY_STATE_ERROR = 5;
 }
 
 /*===========================================================================
--- a/hipcheck/src/engine/mod.rs
+++ b/hipcheck/src/engine/mod.rs
@@ -10,7 +10,7 @@ use crate::{
 	},
 	policy::PolicyFile,
 	policy_exprs::Expr,
-	Result,
+	Error, Result,
 };
 use futures::future::{BoxFuture, FutureExt};
 use serde_json::Value;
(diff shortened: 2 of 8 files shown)


library/hipcheck-common/proto/hipcheck/v1/hipcheck.proto
```diff
@@ -220,6 +224,9 @@ enum QueryState {
 
     // We are sending a query to a plugin and we are expecting to need to send more chunks
     QUERY_STATE_SUBMIT_IN_PROGRESS = 4;
+
+    // An error was encountered while processing.
+    QUERY_STATE_ERROR = 5;
 }
```
j-lanson, line 230: I don't understand why this needs an additional state variant - "Unspecified" would only be used in an error scenario anyway, it's short for "UnspecifiedError". we can also re-name variants without breaking compatibility, so we could just rename "Unspecified" to "Error"

library/hipcheck-common/proto/hipcheck/v1/hipcheck.proto
```diff
@@ -200,6 +200,10 @@ message Query {
     // Used to indicate whether or not a string field present in a `repeated string` field
     // was split between two messages
     bool split = 9;
+
+    // A plugin-generated error string.
+    // Shall be present if and only if state == QUERY_STATE_ERROR
+    optional string error = 10;
```
j-lanson, line 206: Is there a reason we can't re-use an existing field like "key" or "concern" for the error message? Not only does it preserve the existing structure, but we can guarantee "error" can only be treated as an error message when the state matches

library/hipcheck-common/proto/hipcheck/v1/hipcheck.proto
cstepanian: That's correct. I chose to add a new state in the draft version of the PR, mostly because it was easier to distinguish between the new uses (that include an error string) and the old ones (which don't).

Ideally, I'd like to update all existing uses of Unspecified to use error strings. The idea would be to incrementally replace all of them with Error, and finally rename Unspecified to Error when all are converted.

But I wasn't sure if that would end up being too many changes for one PR (both in terms of dev time, and review effort).

alilleybrinker:
Closing given bandwidth limitations. Would love to revive in the future.

PR #1073, opened 2025-03-28
feat: investigate-if-fail in report

### LINKED ISSUE #1058: Make clear in a report when a target is marked "INVESTIGATE" because of `investigate-if-fail` annotations and not scoring.
See title.

Right now, when we mark something for investigation because of a failing analysis marked in `investigate-if-fail`, it's not clear *that's* why we're marking it failing.

j-lanson:
Resolves #1058 .

Adds an additional field `reason` to `Recommendation` to identify why something is marked for investigation. When an "investigate-if-fail" policy is the cause of the overall "investigate" determination on a target, this now will display why in the output report, both in JSON and pretty-print form. Example below:

Screenshot 2025-03-28 092556

As an additional commit, fixed an issue with the SDK logger to make it print to stderr instead of stdout.

### DIFF
--- a/hipcheck/src/policy/policy_file.rs
+++ b/hipcheck/src/policy/policy_file.rs
@@ -592,9 +592,21 @@ impl PolicyPluginName {
 	}
 }
 
+impl From<PolicyPluginName> for String {
+	fn from(value: PolicyPluginName) -> Self {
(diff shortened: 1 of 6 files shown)


KirilldogU (approved):
Good catch with the plugin output destination, and I like the new investigate verbosity!

PR #1075, opened 2025-03-28
feat: Refactor Session for support of multiple target runs

### LINKED ISSUE #1068: Refactor `Session` for parallelization

mchernicoff:
Resolves #1068 . Refactors `Session` and supporting code for future implementation of running Hipcheck against multiple targets.

 `Session` no longer contains a `Target` but instead has a `run()` function that takes a `Target` as an argument. When running against multiple `Target`s, we will initialize a base session with policy and configuration information, then clone that `Session` for each `Target` we want to call `session.run()`.

The current code for looping through `Target`s and `Report`s is only a placeholder that allows for unchanged runs against single targets but will be changed in future PRs.

### DIFF
--- a/hipcheck/src/main.rs
+++ b/hipcheck/src/main.rs
@@ -31,8 +31,7 @@ use crate::{
 	exec::ExecConfig,
 	plugin::{try_set_arch, Plugin, PluginVersion, PluginWithConfig},
 	policy::{config_to_policy, PolicyFile},
-	report::report_builder::{build_report, Report},
-	score::score_results,
+	report::report_builder::Report,
 	session::ReadySession,
 	session::Session,
 	setup::write_config_binaries,
@@ -48,7 +47,7 @@ use indextree::{Arena, NodeId};
 use ordered_float::NotNan;
 use pathbuf::pathbuf;
 use schemars::schema_for;
-use shell::{color_choice::ColorChoice, spinner_phase::SpinnerPhase};
+use shell::{color_choice::ColorChoice, spinner_phase::SpinnerPhase, verbosity::Verbosity};
 use std::{
 	env,
 	fmt::{Display, Formatter},
@@ -60,6 +59,8 @@ use std::{
 	time::Duration,
 };
 use target::{TargetSeed, ToTargetSeed};
+use tokio::runtime::Runtime;
+use tokio_stream::StreamExt;
 use util::command::DependentProgram;
 use util::fs::create_dir_all;
 use which::which;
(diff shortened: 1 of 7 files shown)


hipcheck/src/session/mod.rs
```diff
@@ -286,14 +220,30 @@ impl Session {
 	pub fn format(&self) -> Format {
 		self.format
 	}
+
+	pub fn run(&mut self, phase: &SpinnerPhase, target: &Target, clear_db: bool) -> Result<Report> {
+		// Clear the Salsa db if told to do so
+		if clear_db {
```
mchernicoff, line 226: If we keep the implementation of cloning a new `Session` before each run against a `Target`, this part may not be necessary.

hipcheck/src/target/types.rs
```diff
@@ -162,6 +194,16 @@ pub enum TargetSeed {
 	Multi(MultiTargetSeed),
 }
 
+impl TargetSeed {
+	/// Return whether the target seed is for multiple targets or not
+	pub fn is_multiple_target(&self) -> bool {
```
j-lanson, line 199: --> `is_multi_target`

hipcheck/src/main.rs
```diff
@@ -142,26 +141,39 @@ fn cmd_check(args: &mut CheckArgs, config: &CliConfig) -> ExitCode {
...
 		config.format(),
-	);
+	) {
+		Ok(reports) => {
+			for report in reports {
+				match report {
+					Ok(report) => {
+						if let Err(err) = Shell::print_report(report, config.format()) {
+							Shell::print_error(&err, config.format());
+							return ExitCode::FAILURE;
```
j-lanson, line 158: can we change `run()` to produce `Result<impl tokio_stream::StreamExt<Item = Result<Report>>>` and update the code here accordingly? The outer result being `Err` would indicate something went wrong during setup / plugin initialization. Inner `Result` as the item would indicate something went wrong during the analysis of a given dependency. Inside `run`, we can keep the same behavior of creating a `Vec<Report>`, and then transform it to this output type.

hipcheck/src/target/types.rs
```diff
@@ -26,6 +27,37 @@ pub struct Target {
 	pub package: Option<Package>,
 }
 
+impl Target {
+	/// git ref of the HEAD commit being analyzed
+	pub fn head(&self) -> Arc<String> {
+		Arc::new(self.local.git_ref.clone())
+	}
```
j-lanson, line 34: Why do these need to be Arc? I would make them `&str` and then have the caller wrap it in `Arc` if they so choose. Currently, if would-be callers each call `head()` instead of sharing the output with one another we just create multiple `Arc`s of the same string, so we don't get much benefit by wrapping it in `Arc` here first.

hipcheck/src/main.rs
mchernicoff: I can probably do that, yes.

hipcheck/src/report/report_builder.rs
```diff
@@ -99,7 +107,7 @@ pub struct ReportBuilder<'sess> {
 
 impl<'sess> ReportBuilder<'sess> {
 	/// Initiate building a new `Report`.
-	pub fn for_session(session: &'sess Session) -> ReportBuilder<'sess> {
+	pub fn for_session(session: &'sess Session, target: &'sess Target) -> ReportBuilder<'sess> {
```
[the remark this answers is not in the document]
mchernicoff: Good point. That should be a straightforward switch.

hipcheck/src/target/types.rs
mchernicoff: Probably because I copied them from `session` to here and didn't change the return type. Makes sense why we would want just `&str`.

hipcheck/src/report/report_builder.rs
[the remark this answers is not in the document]
mchernicoff: Per discussion, the plan will be to pass `policy_file` and `analyzed_at` arguments, potentially freeing up the `Session` at that point.

hipcheck/src/session/mod.rs
```diff
@@ -504,7 +455,7 @@ fn load_exec_config(exec_path: Option<&Path>) -> Result<ExecConfig> {
 	Ok(exec_config)
 }
 
-fn load_target(seed: &TargetSeed, home: &Path) -> Result<Target> {
+pub fn load_target(seed: &TargetSeed, home: &Path) -> Result<Vec<Target>> {
```
j-lanson, line 458: I'm ok with merging the PR as-is since the focus was on `Session`, but I don't really want us to be collecting `get_targets()` into a Vec. there should be a tokio task that gets the next element from `get_targets` and adds it to a work queue for tasks running `Session` objects to pull from.

j-lanson (approved):

hipcheck/src/session/mod.rs
mchernicoff: Agree. I didn't want to get too bogged down in implementing that for this PR, but changing it is part of the next issue that I am working on now.

PR #1076, opened 2025-03-28
fix: fixed bug to allow entropy to report commits as concerns

ninaagrawal:
Draft PR request: The concerns are not being printed despite them being added to record_concern.

### DIFF
--- a/plugins/entropy/src/main.rs
+++ b/plugins/entropy/src/main.rs
@@ -145,19 +145,51 @@ async fn entropy(engine: &mut PluginEngine, value: Target) -> Result<Vec<f64>> {
 	let val_commits = engine.query("mitre/git/commit_diffs", local).await?;
 	let commits: Vec<CommitDiff> =
 		serde_json::from_value(val_commits).map_err(Error::InvalidJsonInQueryOutput)?;
-	let commit_entropies = commit_entropies(engine, commits)
+	// clone the commits list to access the commit information since the list will be moved into commit_entropies
+	// Currently we only need hashes but if other information is needed, it is more efficient to have a copy that can be used in the future
+	let commits_clone = commits.clone();
+	let commit_entropies: Vec<f64> = commit_entropies(engine, commits)
 		.await?
 		.iter()
 		.map(|o| o.entropy)
 		.collect();
-
+	// extract the hashes from the commits
+	let commit_hash_ids: Vec<String> = commits_clone.into_iter().map(|c| c.commit.hash).collect();
+	// creates a tuple with the commit hashes and entropies to be passed into the helper function
+	let commits_with_entropies: Vec<(&String, f64)> = commit_hash_ids
+		.iter()
+		.zip(commit_entropies.iter().copied()) // Automatically dereferences &f64 to f64
+		.collect();
+	report_concerns(engine, commits_with_entropies);
 	tracing::info!("completed entropy query");
 	Ok(commit_entropies)
 }
-
+// Helper function to report concerns
+fn report_concerns(engine: &mut PluginEngine, commit_hashes_with_entropies: Vec<(&String, f64)>) {
+	// Get the entropy threshold from the global policy config file
+	let policy_config = EntropyPlugin::global_policy_conf().get().cloned().flatten();
+	let entropy_threshold = match policy_config {
+		None => -100.00,
+		Some(a) => a.entropy_threshold,
+	};
+	// Add hashes to concerns
+	for element in commit_hashes_with_entropies {
+		if element.1 > entropy_threshold {
+			let concern: String = format!("Commit hash: {}", element.0);
+			engine.record_concern(concern);
+		}
+	}
+}
 #[derive(Clone, Debug, Default)]
 struct EntropyPlugin {
 	policy_conf: OnceLock<Option<PolicyExprConf>>,
+}
+impl EntropyPlugin {
+	// Static method to access policy_conf globally
+	fn global_policy_conf() -> &'static OnceLock<Option<PolicyExprConf>> {
+		static GLOBAL_POLICY_CONF: OnceLock<Option<PolicyExprConf>> = OnceLock::new();
+		&GLOBAL_POLICY_CONF
+	}
 }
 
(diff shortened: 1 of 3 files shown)


plugins/entropy/src/main.rs
```diff
@@ -145,20 +145,52 @@ async fn entropy(engine: &mut PluginEngine, value: Target) -> Result<Vec<f64>> {
...
 		.collect();
-
+	// extract the hashes from the commits
+	let commit_hash_ids: Vec<String> = commits_clone.into_iter().map(|c| c.commit.hash).collect();
+	// creates a tuple with the commit hashes and entropies to be passed into the helper function
+	let commits_with_entropies: Vec<(&String, f64)> = commit_hash_ids
+		.iter()
+		.zip(commit_entropies.iter().copied()) // Automatically dereferences &f64 to f64
+		.collect();
+	report_concerns(engine, commits_with_entropies);
```
j-lanson, line 163: `commit_entropies()` already returns a Vec of `CommitEntropy` objects, which contain both the `Commit` object and the `entropy` calculation. So instaed of doing this cloning and zipping, you can just rework the code so we don't use `.map(|o| o.entropy)` to discard the commit information you're looking for.  In fact, you could just update the `map()` function to check if the entropy > threshold and call `record_cocern()` then and there.

j-lanson (changes requested):

plugins/entropy/src/main.rs
ninaagrawal: That makes a lot more sense, for some reason I though I should leave commit_entropies alone.

j-lanson: leave the `commit_entropies` function alone yes, but I'm describing changing how we process the output

ninaagrawal: Thanks I see what you mean. I meant, leave the let commits_with_entropies line alone.

ninaagrawal:
hc check is timing out when the entropy plugin prints out the concerning commit

plugins/entropy/src/main.rs
```diff
@@ -141,25 +141,54 @@ async fn commit_entropies(
...
 	tracing::info!("completed entropy query");
 	Ok(commit_entropies)
 }
+// Helper function for entropy plugin to extract the entropy threshold from the policy config file
+fn extract_entropy_threshold() -> f64 {
+	let policy_config = GLOBAL_POLICY_CONFIG.get().cloned().flatten();
+	match policy_config {
+		None => 10.0, //default value for the entropy threshold
+		Some(a) => a.entropy_threshold,
+	}
```
j-lanson, line 171: We don't set a default if a threshold is not given. The above `record_concern()` calls should happen only if the entropy threshold is set **and** the entropy > threshold. you can use `Option` functions is `is_some_and()` to get the effect we want.

plugins/entropy/src/main.rs
```diff
@@ -141,25 +141,54 @@ async fn commit_entropies(
...
+fn move_policy_config_to_global(plugin: &EntropyPlugin) {
+	if let Some(policy_config) = plugin.policy_conf.get() {
+		// Move the policy_conf into the global OnceLock variable
+		GLOBAL_POLICY_CONFIG
+			.set(policy_config.clone())
+			.expect("Failed to set global policy config");
+	} else {
+		println!("No policy_config found in the plugin");
+	}
+}
```
j-lanson, line 191: Instead of doing this step, just get rid of the field `EntropyPlugin.policy_conf` and set the global directly in `set_config()`

sdk/schema/hipcheck_target_schema.json
j-lanson, line 1: Please unstage this file as a change

j-lanson (changes requested):

plugins/entropy/src/main.rs
```diff
@@ -226,7 +254,9 @@ struct Args {
 #[tokio::main(flavor = "current_thread")]
 async fn main() -> Result<()> {
 	let args = Args::try_parse().unwrap();
-	PluginServer::register(EntropyPlugin::default(), args.log_level)
+	let plugin = EntropyPlugin::default();
+	move_policy_config_to_global(&plugin);
```
j-lanson, line 258: This doesn't work because `set_config()` gets called through `\.listen_local()`, so the `policy_config` is never set here.

j-lanson (changes requested):

plugins/entropy/Cargo.toml
```diff
@@ -22,6 +22,7 @@ tokio = { version = "1.44.1", features = ["rt"] }
 unicode-normalization = "0.1.24"
 hipcheck-workspace-hack = { version = "0.1", path = "../../library/hipcheck-workspace-hack" }
 
+
```
j-lanson, line 25: Please remove this change

plugins/entropy/src/main.rs
```diff
@@ -141,25 +141,54 @@ async fn commit_entropies(
...
 		serde_json::from_value(val_commits).map_err(Error::InvalidJsonInQueryOutput)?;
-	let commit_entropies = commit_entropies(engine, commits)
+	let commit_entropies: Vec<f64> = commit_entropies(engine, commits)
 		.await?
 		.iter()
-		.map(|o| o.entropy)
+		.map(|o| {
+			if o.entropy > entropy_threshold {
+				engine.record_concern(format!(
+					"Concerning commit hash: {}, Entropy: {}",
```
j-lanson, line 155: Remove the word "Concerning", kind of redundant since its being emitted as a concern

plugins/entropy/src/main.rs
ninaagrawal: I thought about doing that earlier but wasn't sure if I should :)

sdk/schema/hipcheck_target_schema.json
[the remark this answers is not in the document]
ninaagrawal: Which commit should I revert the file back to, yours or Andrew's?

j-lanson (approved):

PR #1088, opened 2025-03-31
chore: add docs describing how to unit-test Rust SDK plugins

j-lanson:


ninaagrawal (approved):
It looks good, I think it makes sense!

PR #1089, opened 2025-04-01
feat: python sdk unit testing capabilities

j-lanson:
- Removed the `mock_engine` SDK option, the `PluginEngine` now just checks if `mock_responses is not None` to decide whether to do mocked response.
- Added example query endpoint unit test to `main.py`, renamed `test_main.py` to make it easy for `pytest` to find
- Refactored `mock_responses` from a dict to a `list[tuple]`, because types we envision to be common query keys (object, dict, list) are not hashable by default. Now PluginEngine uses `find_response` function to get the right output object or return None.
- added default-True `init_logger` parameter to `PluginServer.register()` so people can init their own logger if they wish.
- Added section to Python SDK docs on website explaining how to write unit tests for query endpoint funcs

### DIFF
--- a/sdk/python/pyproject.toml
+++ b/sdk/python/pyproject.toml
@@ -23,6 +23,7 @@ dependencies = [
     "logging>=0.4.9.6",
     "pydantic>=2.10.6",
     "pytest>=8.3.5",
+    "pytest-asyncio>=0.26.0",
     "sphinx>=8.1.3",
     "wheel>=0.45.1",
 ]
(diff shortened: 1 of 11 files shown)


KirilldogU:
Code and markdown changes look good to me, I would just suggest rerunning `update-gen.py` and `ruff format` to standardize the formatting!

j-lanson:
> Code and markdown changes look good to me, I would just suggest rerunning `update-gen.py` and `ruff format` to standardize the formatting!

Has been done already :)

KirilldogU:
Oh good point, it seems the order of running `ruff format` and `update-gen.py` causes non-deterministic edits. So running format, and then update-gen.py on those changes will cause codegen-ed code to take a different format, then just running update-gen.py. Either way, insignificant.

KirilldogU (approved):
Good!

PR #1090, opened 2025-04-01
feat: added first iteration of `cargo xtask benchmark`

patrickjcasey:
This PR contains a first pass at implementing `cargo xtask benchmark`

I added a config file to the repository `./config/benchmark-targets.kdl` that contains three different repositories for benchmarking. I am very open to changing these repositories, but they serve as an example of how configuring benchmark gathering works! Each benchmark target currently requires a `ref` in order to make it easier to draw conclusions from benchmark run, as the targets should be constant across benchmark runs!

## Current Process

1. Ensure prerequisites are installed and `/proc/sys/kernel/perf_event_paranoid` (contains 2 or less)
2. Read config file and parse it to determine all benchmark targets
3. Build the release version of all binaries in the repo
4. Run `-r/--runs` number of benchmark runs for each benchmark targets and save off results
5. Serialize results to user-defined CSV file upon completion

## Data Points Captured 

```rust
#[derive(Serialize, Debug)]
pub struct BenchmarkResult {
	/// repo type being analyzed, passed as `-t` to `hc check`
	target_type: BenchmarkTargetType,
	/// target passed to `hc check`
	target: String,
	/// reference identifier for the target, passed as `--ref` to `hc check`
	reference: String,
	/// average wall time across all runs
	avg_wall_time: f32,
	/// average max RSS across all runs
	avg_max_rss_in_kb: u64,
	/// average total instructions across all runs
	avg_total_instructions: u64,
	/// average total number of cpu cycles across all runs
	avg_cpu_cycles: u64,
}
```

## Example Terminal Output

image

## Example CSV

image

This is still accurate, except the target_types are all lowercase now!

## Final Notes

`cargo xtask benchmark` currently only works on Linux

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,7 +9,7 @@ dependencies = [
  "clap",
  "hipcheck-sdk",
  "hipcheck-workspace-hack",
- "jiff",
+ "jiff 0.1.16",
  "serde",
  "serde_json",
  "tokio",
@@ -206,7 +206,7 @@ checksum = "6e0c28dcc82d7c8ead5cb13beb15405b57b8546e93215673ff8ca0349a028107"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.96",
+ "syn 2.0.100",
 ]
 
 [[package]]
@@ -526,7 +526,7 @@ dependencies = [
  "heck 0.5.0",
  "proc-macro2",
  "quote",
- "syn 2.0.96",
+ "syn 2.0.100",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


j-lanson (commented):
I think if we were to organize the output to be conducive for making a website like `https://perf.rust-lang.org/`, we would have a persistent .csv file for each metric we gather, and have rows be targets and columns be data for a given timestamped run.

I'll leave it to you whether you want to tackle that change now or just make an issue for it for the future.

patrickjcasey:
I can fix that up one this branch and get it up on Friday. Do you think it would make sense that `-o` would take a directory and the tool would handle appending to the required files in that directory?

j-lanson:
> I can fix that up one this branch and get it up on Friday. Do you think it would make sense that `-o` would take a directory and the tool would handle appending to the required files in that directory?

Yeah that would make sense to me

patrickjcasey:
Here is the resulting files from running ` cargo xtask benchmark --config config/benchmark-targets.kdl --output bench-results`

image

Here is an example file for `avg_cpu_cycles` with multiple runs of the benchmark:

image

j-lanson (dismissed):

j-lanson (approved):

PR #1099, opened 2025-04-04
feat: Add async target analysis

### LINKED ISSUE #1092: Refactor `run()` to support multiple targets with `tokio::tasks` (or some other threading) for parallel `Sessions`

mchernicoff:
Resolves #1092 

Adds async code to analyze targets and print reports in parallel.

Adds `-j`/`--jobs` flag to `hc check` to allow users to set the number of parallel jobs.

Adds `summary` option to `-f`/`--format` flag that prints a one-line summary for each `Report`

### DIFF
--- a/hipcheck/src/cli.rs
+++ b/hipcheck/src/cli.rs
@@ -159,7 +159,7 @@ struct DeprecatedArgs {
 	quiet: Option<bool>,
 
 	/// Use JSON output format.
-	#[arg(short = 'j', long = "json", hide = true, global = true)]
+	#[arg(long = "json", hide = true, global = true)]
 	json: Option<bool>,
 
 	/// Print the home directory for Hipcheck.
@@ -529,6 +529,10 @@ pub struct CheckArgs {
 	/// The ref (e.g. commit hash, branch, tag) of the target to analyze
 	#[clap(long = "ref", value_name = "REF")]
 	pub git_ref: Option<String>,
+
+	/// The number of parallel jobs to run at once against a multi-target run of Hipcheck (default 5)
+	#[clap(short = 'j', long = "jobs")]
+	pub num_jobs: Option<usize>,
 
 	#[clap(subcommand)]
 	command: Option<CheckCommand>,
@@ -1350,6 +1354,8 @@ pub enum Format {
 	/// Human-readable format.
 	#[default]
 	Human,
+	/// Summary format, suitable for multi-target runs with minimal information printed for each target
+	Summary,
 }
 
 impl Format {
--- a/hipcheck/src/engine/mod.rs
+++ b/hipcheck/src/engine/mod.rs
@@ -194,7 +194,7 @@ impl HcEngineImpl {
 	// analysis plugin to kick off the execution
 }
 
-pub fn start_plugins(
+pub async fn start_plugins(
 	policy_file: &PolicyFile,
 	plugin_cache: &HcPluginCache,
 	executor: PluginExecutor,
@@ -244,7 +244,6 @@ pub fn start_plugins(
 		plugins.push(plugin_with_config);
 	}
 
-	let runtime = RUNTIME.handle();
-	let core = runtime.block_on(HcPluginCore::new(executor, plugins))?;
+	let core = HcPluginCore::new(executor, plugins).await?;
 	Ok(Arc::new(core))
 }
--- a/hipcheck/src/main.rs
+++ b/hipcheck/src/main.rs
@@ -39,12 +39,15 @@ use crate::{
 	setup::write_config_binaries,
 	shell::Shell,
 };
+use async_channel::bounded;
+use async_stream::stream;
 use cli::{
 	CacheOp, CachePluginArgs, CacheTargetArgs, CheckArgs, CliConfig, ConfigMode, FullCommands,
 	PluginArgs, PluginOp, ReadyArgs, SchemaArgs, SchemaCommand, UpdateArgs,
 };
 use config::AnalysisTreeNode;
 use core::fmt;
+use engine::PluginCore;
 use indextree::{Arena, NodeId};
 use ordered_float::NotNan;
 use pathbuf::pathbuf;
@@ -58,11 +61,12 @@ use std::{
 	path::{Path, PathBuf},
 	process::{Command, ExitCode},
 	result::Result as StdResult,
+	sync::Arc,
 	time::Duration,
 };
 use strum::IntoEnumIterator;
 use target::{TargetSeed, ToTargetSeed};
-use tokio::runtime::Runtime;
+use tokio::{runtime::Runtime, sync::Mutex};
 use tokio_stream::StreamExt;
 use util::command::DependentProgram;
 use util::fs::create_dir_all;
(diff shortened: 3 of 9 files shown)


hipcheck/src/main.rs
```diff
@@ -870,28 +882,51 @@ fn run(
...
+		let targets = Arc::clone(&targets);
+		let sender = sender.clone();
+		let phase = phase.clone();
+
+		tokio::spawn(async move {
+			// Get the next available target
+			let mut targets = targets.lock().await;
+			while let Some(target) = targets.next().await {
+				// Run plugins against the target, generate scores, and build a report
+				// Clear the Salsa db and reset the start time, in case we are re-using an existing session
```
j-lanson, line 908: This won't work for parallelization because the first session will lock and hold the `targets` until it finishes, so no other session gets to grab a target to work on.

hipcheck/src/main.rs
```diff
@@ -870,28 +882,51 @@ fn run(
...
+
+				// Clone non-copy parameters again
+				let mut session = session.clone();
+				let phase = phase.clone();
+				// Use spawn_blocking because Salsa db's cannot be used in async code
+				let report =
+					tokio::task::spawn_blocking(move || session.run(&phase, target, true)).await;
+
```
j-lanson, line 916: Hmm, if we spawn blocking then we lose the benefits of the async code. Could we spawn it in a different thread and `await` for it to finish?

hipcheck/src/main.rs
```diff
@@ -870,28 +882,51 @@ fn run(
...
+			while let Some(target) = targets.next().await {
+				// Run plugins against the target, generate scores, and build a report
+				// Clear the Salsa db and reset the start time, in case we are re-using an existing session
+
+				// Clone non-copy parameters again
+				let mut session = session.clone();
+				let phase = phase.clone();
+				// Use spawn_blocking because Salsa db's cannot be used in async code
+				let report =
+					tokio::task::spawn_blocking(move || session.run(&phase, target, true)).await;
```
j-lanson, line 915: We either `session.clone()` or call `session.run()` with `clear_db = true`, but not both. Can we call `run` on just a reference to session?

hipcheck/src/main.rs
j-lanson: You'll have to `loop {}` and each iteration take the lock, check for next, if None exit, else get element, drop lock, run the analysis

hipcheck/src/cli.rs
```diff
@@ -530,6 +530,10 @@ pub struct CheckArgs {
 	#[clap(long = "ref", value_name = "REF")]
 	pub git_ref: Option<String>,
 
+	/// The number of parallel jobs to run at once against a mult-target run of Hipcheck (default 5)
```
j-lanson, line 533: *multi

hipcheck/src/engine/mod.rs
```diff
@@ -244,7 +244,8 @@ pub fn start_plugins(
 		plugins.push(plugin_with_config);
 	}
 
-	let runtime = RUNTIME.handle();
-	let core = runtime.block_on(HcPluginCore::new(executor, plugins))?;
+	// let runtime = RUNTIME.handle();
+	// let core = runtime.block_on(HcPluginCore::new(executor, plugins))?;
```
j-lanson, line 248: Delete comments

hipcheck/src/main.rs
```diff
@@ -885,15 +897,20 @@ impl CheckKind {
...
 /// Now that we're fully-initialized, run Hipcheck's analyses.
 #[allow(clippy::too_many_arguments)]
-fn run(
+async fn run(
 	seed: TargetSeed,
 	config_mode: ConfigMode,
 	home_dir: Option<PathBuf>,
 	exec_path: Option<PathBuf>,
 	format: Format,
+	num_jobs: usize,
```
j-lanson, line 906: can just call it `jobs`

hipcheck/src/main.rs
```diff
@@ -904,28 +921,67 @@ fn run(
...
+						let core = core.clone();
+						let base_session = base_session.clone();
+						let phase = phase.clone();
+
+						// Run plugins against the target, generate scores, and build a report
+						//
+						// Use spawn_blocking because Salsa db's cannot be used in async code
+						let report = tokio::task::spawn_blocking(move || {
+							let mut session = base_session.new_with_core(core);
+							session.run(&phase, stream_target)
```
j-lanson, line 965: I would prefer to handle the `Result<>` part of `stream_target` here instead of passing to `session.run()`, so `run` can just take `&Target`

before calling `new_with_core()`, can just check if `stream_target` is Err and return early.

site/content/docs/getting-started/install.md
```diff
@@ -149,7 +149,7 @@ Hipcheck image, you might run:
 
 ```sh
 $ export HC_GITHUB_TOKEN=<TOKEN_VALUE>
-$ docker run -e HC_GITHUG_TOKEN=$HC_GITHUB_TOKEN mitre/hipcheck:latest check https://github.com/mitre/hipcheck
+$ docker run -e HC_GITHUB_TOKEN=$HC_GITHUB_TOKEN mitre/hipcheck:latest check https://github.com/mitre/hipcheck
```
j-lanson, line 152: Weird change but not a big deal

j-lanson (changes requested):
Overall, this is in a really good spot. Just a few requested touchups

site/content/docs/getting-started/install.md
mchernicoff: Someone I was working with using Hipcheck for the first time spotted it. Figured it was easier to just fix it in place then make a whole Issue of it. Feel free to fix in another PR that will get merged for the next release.

j-lanson: no, my fault. I did not readily see the typo. good catch, thank you to whoever it was

j-lanson (approved):

hipcheck/src/main.rs
```diff
@@ -93,7 +97,11 @@ fn main() -> ExitCode {
 	}
 
 	match config.subcommand() {
-		Some(FullCommands::Check(mut args)) => return cmd_check(&mut args, &config),
+		Some(FullCommands::Check(mut args)) => {
+			// Panic: Safe to unwrap as Runtime::new() always returns as Ok
+			let runtime = Runtime::new().unwrap();
```
cstepanian, line 102: Is it certain that this `Result` will always be `Ok()`?

The `tokio` docs don't appear to make that guarantee, but also don't explicitly say what types of error can occur.

Their example code for the `runtime` module uses `?` instead of `.unwrap()`: https://docs.rs/tokio/1.44.1/tokio/runtime/index.html#usage

But the shorter example on `Runtime::new()` does use `.unwrap()`: https://docs.rs/tokio/1.44.1/tokio/runtime/struct.Runtime.html#method.new

I followed the call chain down to `tokio::runtime::io::driver::Driver::new()`: https://docs.rs/tokio/1.44.1/src/tokio/runtime/io/driver.rs.html#91

This function calls some `mio` functions, whose docs do mention the possibility of errors from OS-specific syscalls.

I haven't fully tracked down what kind of situations might cause those APIs to fail, but it definitely seems possible (though perhaps only in rare situations).

cstepanian: I now see that this code was likely just moved from `hipcheck/src/session/mod.rs`, which had the same potential issue.

Issue #1071, opened 2025-03-26
Add support for package-lock.json multi-target seed


j-lanson:
Want to: get `TargetSeed.get_targets()` for a package-lock.json return an iterator of Targets.

Can temporarily test with test case instead of CLI.

PR #1109, opened 2025-04-08
feat: parse package-lock.json npm dependencies to SingleTargetSeeds

KirilldogU:
Related to #1071 .
Parses package-lock.json to SingleTargetSeeds. Called by `get_targets()` method of `TargetSeed`, currently just tested via a unit-test and a sample `package-lock.json` added in a test data folder. Will be used after CLI support for package-lock.json input is added.

### DIFF
--- a/hipcheck/src/target/multi.rs
+++ b/hipcheck/src/target/multi.rs
@@ -1,17 +1,20 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
+	cli::format_npm_url,
 	error::{Error, Result},
 	hc_error, source,
-	target::{SingleTargetSeed, SingleTargetSeedKind},
+	target::{Package, PackageHost, SingleTargetSeed, SingleTargetSeedKind},
 	util::http::agent,
 };
 
 use gomod_rs::{Context as GmContext, Directive, Identifier};
 use regex::Regex;
+use serde::Deserialize;
+use serde_json::from_slice;
 use url::Url;
 
-use std::{ops::Deref, path::Path, sync::LazyLock};
+use std::{collections::BTreeMap, ops::Deref, path::Path, sync::LazyLock};
 
 /// A struct with owned data to simplify TryFrom conversion impl
 /// vs. gomod_rs::Context which is full of lifetimes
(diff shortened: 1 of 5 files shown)


hipcheck/src/target/multi.rs
```diff
@@ -46,3 +48,129 @@ pub(crate) async fn resolve_go_mod(path: &Path) -> Result<Vec<SingleTargetSeed>>
...
+
+#[derive(Deserialize)]
+pub struct PackageDependency {
+	pub version: String,
+}
+
+pub(crate) async fn resolve_package_lock_json(path: &Path) -> Result<Vec<SingleTargetSeed>> {
+	// Parse package-lock.json
+	let file =
+		File::open(path).map_err(|e| hc_error!("Failed to load package-lock.json: {}", e))?;
```
j-lanson, line 66: might as well use an `async` function API for reading the file

hipcheck/src/target/multi.rs
```diff
@@ -46,3 +48,129 @@ pub(crate) async fn resolve_go_mod(path: &Path) -> Result<Vec<SingleTargetSeed>>
...
+
+				// If the package is scoped, replace the leading '@' in the scope with %40 for proper pURL formatting
+				let purl = Url::parse(&match version.as_str() {
+					"no version" => format!("pkg:npm/{}", str::replace(&name, '@', "%40")),
+					_ => format!(
+						"pkg:npm/{}@{}",
+						str::replace(&name, '@', "%40"),
+						version.clone()
+					),
+				})
```
j-lanson, line 89: Is this %40 business not something that `Url::parse` handles?

hipcheck/src/target/multi.rs
```diff
@@ -46,3 +48,129 @@ pub(crate) async fn resolve_go_mod(path: &Path) -> Result<Vec<SingleTargetSeed>>
...
+				.unwrap();
```
j-lanson, line 90: No unwrap, should do `map_err()`

j-lanson (changes requested):

hipcheck/src/target/multi.rs
KirilldogU: I don't believe so, as I copied this portion of the code from `impl ToTargetSeedKind for CheckNpmArgs to_target_seed_kind()`, which had this patch.

j-lanson: If you copied it, can you instead make a little helper func that they both use. I really want to avoid parsing logic being duplicated throughout the codebase

KirilldogU: Gotcha, make sense. Just added a small helper func below its original use in `cli.rs`.

PR #1119, opened 2025-04-10
fix: Handle possible error in creating new default tokio runtime

### LINKED ISSUE #1120: Handle possible error for new `tokio` Runtime
In line 102 of /hipcheck/src/main.rs we call `let runtime = Runtime::new().unwrap();`

This may not be a safe `unwrap()` call. @cstepanian says:

> Is it certain that this Result will always be Ok()?
> 
> The tokio docs don't appear to make that guarantee, but also don't explicitly say what types of error can occur.
> 
> Their example code for the runtime module uses ? instead of .unwrap(): https://docs.rs/tokio/1.44.1/tokio/runtime/index.html#usage
> 
> But the shorter example on Runtime::new() does use .unwrap(): https://docs.rs/tokio/1.44.1/tokio/runtime/struct.Runtime.html#method.new
> 
> I followed the call chain down to tokio::runtime::io::driver::Driver::new(): https://docs.rs/tokio/1.44.1/src/tokio/runtime/io/driver.rs.html#91
> 
> This function calls some mio functions, whose docs do mention the possibility of errors from OS-specific syscalls.
> 
> I haven't fully tracked down what kind of situations might cause those APIs to fail, but it definitely seems possible (though perhaps only in rare situations).

To be safe, we should handle this error without an `unwrap()`.

j-lanson:
Resolves #1120

### DIFF
--- a/hipcheck/src/main.rs
+++ b/hipcheck/src/main.rs
@@ -98,9 +98,14 @@ fn main() -> ExitCode {
 
 	match config.subcommand() {
 		Some(FullCommands::Check(mut args)) => {
-			// Panic: Safe to unwrap as Runtime::new() always returns as Ok
-			let runtime = Runtime::new().unwrap();
-			return runtime.block_on(cmd_check(&mut args, &config));
+			// Create a new runtime for the check command, which uses async code
+			match Runtime::new() {
+				Ok(runtime) => return runtime.block_on(cmd_check(&mut args, &config)),
+				Err(e) => Shell::print_error(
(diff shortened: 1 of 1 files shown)


j-lanson:
> Looks good to me!
> 
> As an aside, it's unfortunate that it takes so much boilerplate to deal with errors in `fn main()`. I've been thinking for a while about refactoring this so there's only once place where `Err(_)` gets turned into `Shell::print_error()`.

Make a GitHub issue so we can track it :)

PR #1210, opened 2025-08-19
chore: split GitHub Actions between universal and platform-specific parts

alilleybrinker:


alilleybrinker:
Huh, the Windows tests failed because they can't find Git, though Git is listed as one of the installed pieces of software on that image. Will have to investigate.

I've also been seeing very long build times on macos-13, and it seems I'm not the only one:

- https://github.com/actions/runner-images/issues/12778

alilleybrinker:
Figured out what was going on with the path issue. Running under `cargo run` in CI on Windows was causing the `git` Hipcheck plugin to appear earlier in the path and then get picked up by Hipcheck, causing early `git --version` in Hipcheck to fail. Running the binary directly (since it's already built by a prior step) avoids the path modification and the issue.

PR #1219, opened 2025-08-21
chore: cleanup deps

alilleybrinker:


alilleybrinker:
Looks like a fixup I made for code using `jiff::Span` to get `jiff` 0.2.x compiling actually broke some tests. I'll need to fix the fix.

Amazingly, I'm currently getting linking errors with `libstd` when trying to run nextest locally. Will have to sort that out before I can run these tests myself.

PR #1256, opened 2025-11-18
Bump deps to latest

alilleybrinker:


alilleybrinker:
Okay, now the run tests are failing without any of that logic being changed from prior test runs. Will need to test further.

alilleybrinker:
In local runs of the same command these download issues don't appear. Going to try another rerun on GitHub on the off chance this is a weird GitHub-specific issue, given that they had a big outage recently.

alilleybrinker:
Ah, yup, fresh outage: https://www.githubstatus.com/incidents/cg3wwz9dw5dg

PR #1277, opened 2026-04-23
fix: Adds .spdx.json as valid SPDX SBOM target

### LINKED ISSUE #1163: Add `.spdx.json` to the list of filetypes for SPDX SBOM targets
See title. Currently we don't infer anything from this type, so it'll be determined to a repo-type seed. Specifying `-t sbom` doesn't override, just leads to a conflict between the inference and the specified target type.

mchernicoff:
Resolves #1163 by adding `.spdx.json` as a supported file extension for SPDX SBOM targets.

Also makes some linting changes based on updates to the Rust `clippy` linter.

### DIFF
--- a/hipcheck/src/cli.rs
+++ b/hipcheck/src/cli.rs
@@ -888,7 +888,7 @@ impl ToTargetSeedKind for CheckSbomArgs {
 	fn to_target_seed_kind(&self) -> Result<TargetSeedKind> {
 		let path = PathBuf::from(&self.path);
 		if path.exists() {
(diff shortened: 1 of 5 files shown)


alilleybrinker (approved):

alilleybrinker:
We can ignore the CI failures on ARM Linux, since we haven't done a release on that platform yet, there are no plugin manifests published for it. This will be fixed in CI after the next minor version release. If the other checks all pass, I'll force merge it.

PR #1281, opened 2026-05-06
chore: release Hipcheck 3.15.0

alilleybrinker:
This bumps all appropriate versions across Cargo and Hipcheck plugin manifests. It also amends `config_to_policy` and its main test, so users of the legacy configuration file will continue to get an up-to-date policy file in the
transliteration process.

Once this is merged I can tag the releases and kickoff cargo-dist release builds.

---

DO NOT MERGE THIS until the announcement blog post (in a separate PR) has been reviewed and approved, but not merged).

After this, run `cargo xtask manifest` to update the download manifests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,7 +4,7 @@ version = 4
 
 [[package]]
 name = "activity"
-version = "0.5.2"
+version = "0.6.0"
 dependencies = [
  "clap",
  "hipcheck-sdk",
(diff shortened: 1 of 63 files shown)


alilleybrinker:
Note that platform runs are failing because this is a release PR, which means it's trying to run against the _next version_ which we're in the process of releasing, but which isn't out yet. So it's failing when it pulls down the download manifests for the plugins and can't find the version it needs. We can ignore it.

mchernicoff (approved):

